PR #11, opened 2017-02-22
Bundle js

jtormey:
This changes the js build process so we can make any needed changes to suit JsCore in wallet-ios.js (no more `sed`). It also makes developing alongside My-Wallet-V3 easier, just run `npm run build:watch`. Note, the js only exposes MyWalletPhone globally now.


woudini:
Can you change the scripts to include the `build` steps in `package.json`? v3.26.2 still uses grunt.

jtormey:
The scripts are part of the iOS project, not My-Wallet-V3. They're in [this package.json](https://github.com/blockchain/My-Wallet-V3-iOS/blob/4e46fdd4822f23499f6f55805d3ee41544df1ba0/package.json), can run them from the project root.

PR #12, opened 2017-03-01
Fix refactor to be order independent

jtormey:
This should fix the order issue that @woudini noticed the enum was causing.


woudini:
Looks good, thanks.

PR #16, opened 2017-04-02
Fetch wallet options from server

jtormey:


Blockchain/js/wallet-ios.js
```diff
@@ -2018,3 +1993,24 @@ MyWalletPhone.getNetworks = function() {
 MyWalletPhone.getECDSA = function() {
     return ECDSA;
 }
+
+function WalletOptions (api) {
+  let optionsCache = {};
```
woudini: This needs to be `var` to prevent the app from erroring on iOS 8

jtormey: Force of habit... Thanks, nice catch!

PR #18, opened 2017-04-07
Pass isNew to buy webview

jtormey:
@woudini let me know if there is a better way to handle this.

### DIFF
--- a/Blockchain/BuyBitcoinViewController.m
+++ b/Blockchain/BuyBitcoinViewController.m
@@ -85,7 +85,7 @@ NSString* loginWithGuidScript(NSString* guid, NSString* sharedKey, NSString* pas
 
 NSString* loginWithJsonScript(NSString* json, NSString* externalJson, NSString* magicHash, NSString* password, BOOL isNew)
 {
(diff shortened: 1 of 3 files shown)


Blockchain/BuyBitcoinViewController.m
```diff
@@ -70,21 +71,26 @@ - (WKWebView *)webView:(WKWebView *)webView createWebViewWithConfiguration:(WKWe
...
 - (void)loginWithGuid:(NSString *)guid sharedKey:(NSString *)sharedKey password:(NSString *)password
 {
-    NSString *script = funcWithArgs(@"activateMobileBuy", guid, sharedKey, password, nil);
+    NSString *script = loginWithGuidScript(guid, sharedKey, password);
     [self runScriptWhenReady:script];
 }
 
+NSString* loginWithJsonScript(NSString* json, NSString* externalJson, NSString* magicHash, NSString* password, BOOL isNew)
+{
+    return [NSString stringWithFormat:@"activateMobileBuyFromJson('%@','%@','%@','%@',%s)", [json escapeStringForJS], [externalJson escapeStringForJS], [magicHash escapeStringForJS], [password escapeStringForJS], (isNew) ? "true" : "false"];
```
woudini: Using `%d` instead of `%s` here pass `isNew` as a boolean so no need for "true" or "false"

PR #19, opened 2017-04-11
Shell scripts to make working with js easier

jtormey:
Copy-pasting that script every time I wanted to make a My-Wallet-v3 change was driving me crazy. I think these scripts are a good temporary improvement.

@mpfluger I changed the shell script build phase to use these too, it should work the same as before but please let me know if I missed something.


jtormey:
Cool, didn't know about that, thanks @mpfluger!

woudini:
Thanks! This is a big help.

PR #24, opened 2017-05-17
Localizable.strings

fborges42:
Some sentences in PT-BR were translated to PT-PT.
Some wrong translations were also corrected.


woudini:
Thanks for the fixes!

PR #36, opened 2017-11-21
Exchange details

achtenhagen:
First work in progress pull request for the Shapeshift exchange detail screen.

### DIFF
--- a/Blockchain/ExchangeDetailView.h
+++ b/Blockchain/ExchangeDetailView.h
@@ -1,0 +1,13 @@
+//
+//  ExchangeDetailView.h
+//  Blockchain
+//
+//  Created by Maurice A. on 11/20/17.
+//  Copyright © 2017 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@interface ExchangeDetailView : UIView
+- (void)createPseudoTableWith:(NSString *)depositAmount receiveAmount:(NSString *)recAmt exchangeRate:(NSString *)rate transactionFee:(NSString *)fee networkTransactionFee:(NSString *)netwkFee;
+@end
--- a/Blockchain/ExchangeDetailView.h
+++ b/Blockchain/ExchangeDetailView.h
@@ -9,5 +9,5 @@
 #import <UIKit/UIKit.h>
 
 @interface ExchangeDetailView : UIView
-- (void)createPseudoTableWith:(NSString *)depositAmount receiveAmount:(NSString *)recAmt exchangeRate:(NSString *)rate transactionFee:(NSString *)fee networkTransactionFee:(NSString *)netwkFee;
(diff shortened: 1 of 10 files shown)


Blockchain/ExchangeDetailView.h
```diff
@@ -0,0 +1,13 @@
...
+//  Blockchain
+//
+//  Created by Maurice A. on 11/20/17.
+//  Copyright © 2017 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@interface ExchangeDetailView : UIView
+- (void)createPseudoTableWith:(NSString *)depositAmount receiveAmount:(NSString *)recAmt exchangeRate:(NSString *)rate transactionFee:(NSString *)fee networkTransactionFee:(NSString *)netwkFee;
```
woudini: Typically Obj-C methods include the name of the parameter, e.g. `writeToURL:atomically:encoding:error:` so a clearer name would be `createPseudoTableWithDepositAmount`

Also variable names `recAmt` and `netwkFee` should be spelled out IMO. Common abbreviations like `dict`, `obj` are acceptable but it's almost always better to spell them all out.

Blockchain/ExchangeDetailView.m
```diff
@@ -0,0 +1,65 @@
...
+#import "BCLine.h"
+
+#define MARGIN_HORIZONTAL 20
+
+@implementation ExchangeDetailView
+
+- (instancetype)initWithFrame:(CGRect)frame
+{
+    self = [super initWithFrame:frame];
+    return self;
```
woudini: In Obj-C convention is to check if there `self` is not nil. I usually write `if (self == [super init])`. See https://stackoverflow.com/a/3275411

Blockchain/ExchangeProgressViewController.m
```diff
@@ -0,0 +1,30 @@
...
+@interface ExchangeProgressViewController ()
+
+@end
+
+@implementation ExchangeProgressViewController
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    self.view.backgroundColor = COLOR_TABLE_VIEW_BACKGROUND_LIGHT_GRAY;
+    ExchangeDetailView *detailView = [[ExchangeDetailView alloc] initWithFrame:CGRectMake(0, 0, WINDOW_WIDTH, 283)];
```
woudini: Could change `283` to a `#define` at the top

Blockchain/ExchangeDetailView.m
achtenhagen: I only removed the if statement because I had no code in the `initWithFrame` method.

Blockchain/ExchangeProgressViewController.m
achtenhagen: Makes sense and I'll define the height as a constant.

PR #47, opened 2018-02-05
update(Images): update assets

woudini:


.gitignore
```diff
@@ -27,3 +27,4 @@ Artwork/node_modules
 External/*
 
 Images.xcassets/*/*.png
```
achtenhagen: Would it make sense to remove this all together?

woudini: We can remove it in the v4 rewrite.

PR #52, opened 2018-02-15
Update InfoPlist.strings

topolab:
first pull request for translation


bchrisarriola:
@topolab thanks for the pull request!

Looks good according to Google translate, but since there is no one that speaks Greek on the team, we can't verify if this translation is better than before.

Can you give us a quick explanation of why this translation should be corrected?

PR #57, opened 2018-02-23
more translations 

topolab:
fixing greek translations


bchrisarriola:
Same comment here as in #52. Can you provide an explanation for these translations?

PR #83, opened 2018-04-13
fix(Transaction-Detail): ensure that labels converted to numbers by J…

woudini:
…S are strings in model


bchrisarriola:
Adding a note that `Transaction.from` should be refactored into a model/class instead of a `NSDictionary` (could be a part of the refactor). Good for now though 👍

PR #86, opened 2018-04-16
Update Get SSL Certificate Build Phase

achtenhagen:
This PR ensures that the `Cert` directory exists before trying to copy the directory in the `Copy Bundle Resources` build phase.

### DIFF
--- a/Blockchain.xcodeproj/project.pbxproj
+++ b/Blockchain.xcodeproj/project.pbxproj
@@ -2407,7 +2407,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 1;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


achtenhagen:
Yes, I think that would be a better solution.

woudini:
Is this waiting on more changes?

achtenhagen:
Yes, they will be pushed shortly.

mpfluger:
If we only run this once, how do you get updates if the cert changes?

achtenhagen:
We could run it only on install and just create the `Cert` directory if it does not already exists.

bchrisarriola:
> We could run it only on install and just create the Cert directory if it does not already exists.

Just to confirm, will "run on install" also work if we eventually use fastlane to create builds (i.e. will create an archive but not install to the sim/device)?

achtenhagen:
Will further investigate this issue once continuous deployment is setup. For now I will update the script to always fetch the certificate on installation, which at this time means that should the certificate become invalid, reinstalling the application would be the working solution.

achtenhagen:
Updated in #96.

PR #89, opened 2018-04-17
Fix exchange

woudini:
There's still a hard-to-find issue with this - sometimes when changing the amounts in fields you get different amounts (amounts you may not have typed in). This may be related to the async operation `getApproximateQuote` or an incorrect assignment to the pointers `btcField`, `bchField`, and `ethField`.


bchrisarriola:
lgtm 👍

PR #92, opened 2018-04-17
refactor(NSUserDefaults) create BlockchainSettings to wrap NSUserDefaults.

bchrisarriola:
`BlockchainSettings` is a wrapper around user-specific settings that is saved to `UserDefaults`. As part of the refactor, let's move most settings from `NSUserDefaults` into this class so that the application-level doesn't have to reference string keys.

I just moved a few properties for now, but other keys should pretty much have the same pattern.

Note: this PR is against `delegate` for now, but should be against `swift` once the `swift` branch is up-to-date (@achtenhagen is currently updating that branch).


bchrisarriola:
@woudini any other naming recommendations? I don't think we should name this `BlockchainUserDefaults` because although these settings are currently backed in `UserDefaults`, it doesn't have to be.

@achtenhagen cool, will move the keys there.

achtenhagen:
👍

Bethannfrum:
Let’s restore and put back to original form and fix all issues and then release documents

PR #93, opened 2018-04-18
[refactor] Adding coordinators for main and welcome views

bchrisarriola:
Ended up with a bigger PR than I wanted but a lot of the files changed were from a project-level find/replace. I'll summarize the main changes below so that this is easier to review.

Summary of changes:
* Added `AppCoordinator` - this coordinator (you can read more about the coordinator pattern [here](https://will.townsend.io/2016/an-ios-coordinator-pattern)) is in charge of deciding which views/view controller to display when the app comes to foreground/background (i.e. the main wallet/dashboard view, if the user is logged in, or the welcome view, if the user is logged out). This coordinator also contains the key window of the application.
* Refactored all `app.window` references to `[UIApplication sharedApplication].keyWindow`
* Added `OnboardingCoordinator` - this coordinator has the logic around the onboarding flow (i.e. displaying the welcome screen and any other onboarding-related screens)
* Added `ModalPresenter` which is in charge of presenting a `BCModalContentView`.
* Commented out code in `RootService` to indicate that has been refactored (this way we can tell what still needs to be translated and eventually deleted `RootService`)

Note: 
* I'll address the TODOs in a separate PR, didn't want to include it here so that the code reviews are manageable.
* I did this PR against `blockchain-settings`, this should be updated to be against `swift` once #92  is merged and the `swift` branch is updated.


bchrisarriola:
Closing this. I'll create a new PR with just the AppCoordinator changes.

PR #94, opened 2018-04-18
Refactor App Delegate, Root Service, Build Phases, Localizations & Images.xcassets

achtenhagen:
This PR kicks off the great refactor of 2018. It migrates the application delegate to Swift, sets up the Root Service for decoupling, updates the build phases, begins refactoring localization constants and checks in all assets in Images.xcassets into source control.


Blockchain/RootServiceSwift.swift
```diff
@@ -0,0 +1,227 @@
...
+//  Created by Maurice A. on 2/16/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+//  Swift implementation of the Root Service
+
+import Foundation
+
+// TODO: rename RootServiceSwift -> RootService once migration is complete
+
+final class RootServiceSwift {
```
bchrisarriola: I think we should remove this class and put most of this logic into `AppDelegate.swift`.

So that we can incrementally introduce changes, `RootService` can then call into the `AppDelegate.swift` lifecycle methods. 

Thoughts on this?

Blockchain/RootServiceSwift.swift
```diff
@@ -0,0 +1,227 @@
...
+        //: Was initialized with `NSNumber numberWithInt:AssetTypeBitcoin` before, could cause side unwanted effects...
+        // TODO: test for potential side effects
+        let assetTypekey = UserDefaults.Keys.assetType.rawValue
+        UserDefaults.standard.register(defaults: [assetTypekey: [AssetType.bitcoin]])
+
+        let certPinningkey = UserDefaults.DebugKeys.enableCertificatePinning.rawValue
+        UserDefaults.standard.register(defaults: [certPinningkey: true])
+
+        let swipeToReceiveEnabledKey = UserDefaults.Keys.swipeToReceiveEnabled.rawValue
+        UserDefaults.standard.register(defaults: [swipeToReceiveEnabledKey: true])
```
bchrisarriola: I don't think we should be calling `UserDefaults` directly, instead, we can use `BlockchainSettings` (in #92) which is a wrapper around it. Don't need to address that in this PR since #92 needs to be merged first.

Blockchain/RootServiceSwift.swift
bchrisarriola: Update after call—let's keep it and eventually deprecate

Blockchain/RootServiceSwift.swift
bchrisarriola: Update after call—we all agree on this change

mpfluger:
There's kind of a workaround by appending `?w=1` (ignore whitespace changes) to the URL when viewing changes on Github. The correct solution though is to keep pure formatting changes separate either as a separate PR or separate commit.

PR #96, opened 2018-04-18
fix(Build Phases): fetch certificate when installing only

achtenhagen:
Until the continuous deployment pipeline is setup, the certificate will only be fetched when installing the application. Additionally, the build phase will check whether the `Cert` directory exists and create it if necessary.

### DIFF
--- a/Blockchain.xcodeproj/project.pbxproj
+++ b/Blockchain.xcodeproj/project.pbxproj
@@ -2397,7 +2397,7 @@
 		};
 		23FDEFA31D6E22C900FC80D6 /* Get SSL Certificate */ = {
 			isa = PBXShellScriptBuildPhase;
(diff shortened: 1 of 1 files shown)


achtenhagen:
The `runOnlyForDeploymentPostprocessing` just represents the checkbox labeled `Run script only when installing`, so it will fetch the certificate when the user installs the application.

woudini:
I'm saying that when someone first sets up the project, the certificate will not be fetched until they archive the project. If they run it on a simulator or device, the app will crash because there is no certificate stored.

achtenhagen:
I misunderstood the function of this setting. It will indeed only execute the script when archiving the product. I will create a fix.

woudini:
lgtm

PR #102, opened 2018-04-19
refactor(BCModalView): using ModalPresenter to show/close BCModalViews

bchrisarriola:
Summary of changes:

* Using ModalPresenter to show/close BCModalViews.
* Moving animation constants to `Constants.swift`.
* Created `ConstantsObjcBridge` so that values within `Constants.swift` can be accessed.
* Commented out show/close `BCModalView` related code in RootService.
* Added a few TODOs (busyView related), will address those in a separate PR.

Notes:
* Please review/merge #101 before reviewing this

### DIFF
--- a/Blockchain/ModalPresenter.swift
+++ b/Blockchain/ModalPresenter.swift
@@ -32,13 +32,94 @@ typealias OnModalResumed = (() -> Void)
         super.init()
     }
 
+    @objc func closeAllModals() {
+        // TODO: handle busy view
+//        [self hideBusyView];
+
+        // TODO: figure out why this is related to closing modals
+//        secondPasswordSuccess = nil;
+//        secondPasswordTextField.text = nil;
+
+        // TODO: use wallet singleton
+        app.wallet.isSyncing = false
+
+        guard let modalView = modalView else { return }
+
+        modalView.endEditing(true)
+        modalView.removeFromSuperview()
+
+        let animation = CATransition()
+        animation.duration = Constants.Animation.duration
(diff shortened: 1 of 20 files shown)


bchrisarriola:
@woudini @achtenhagen also want to share my thoughts here about why I created the wrapper `ModalPresenter` over using `BCModalViewController` which I think should be the preferred way to display a `BCModalView`. The reason is that the latter would introduce a big functionality change and it can be difficult to catch all edge cases esp. with all the changes we're already making with refactoring (we discussed this yesterday). Once we have the new architecture in place, we can then move over to `BCModalViewController` #incremental

Blockchain/ModalPresenter.swift
```diff
@@ -32,13 +32,94 @@ typealias OnModalResumed = (() -> Void)
         super.init()
     }
 
+    @objc func closeAllModals() {
+        // TODO: handle busy view
+//        [self hideBusyView];
+
+        // TODO: figure out why this is related to closing modals
+//        secondPasswordSuccess = nil;
+//        secondPasswordTextField.text = nil;
```
woudini: `secondPasswordSuccess` is a global variable that is used to store a JS callback in the form of a block that takes an `NSString` argument as the second password:
https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/Wallet.m#L3487-L3489

bchrisarriola: Got it. Thanks for the explanation.

PR #104, opened 2018-04-20
refactor(AlertViewPresenter): Creating AlertViewPresenter.

bchrisarriola:
Removing alert view presentation out of `RootService` and into `AlertViewPresenter`.

Please review/merge #103 before reviewing this.


bchrisarriola:
Planning on making one quick change

bchrisarriola:
This is good to review now.

I added a computed property on `UIViewController` called `topMostViewController`. Right now, it's only used in `AlertViewPresenter` but this property deprecates `topViewControllerDelegate` (will address that in the new pull request).

PR #105, opened 2018-04-20
fix(Dashboard): always show at least one integer digit for fiat balance

woudini:
To display `$0.00` instead of `$.00`


bchrisarriola:
Also, this is a good example of something that can easily be unit tested

PR #108, opened 2018-04-23
Network manager

achtenhagen:
WIP PR that adds support for the Network Manager to the project. The network manager handles network related tasks such as requests and sessions.


Blockchain/Network/API.swift
```diff
@@ -0,0 +1,136 @@
...
+/**
+ Manages URL endpoints and request payloads for the Blockchain API.
+ # Usage
+ TBD
+ - Author: Maurice Achtenhagen
+ - Copyright: Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+ */
+
+@objc
+final class API: NSObject {
```
bchrisarriola: The name `API` is too general. Can we rename this to something like `BlockchainAPI`?

Blockchain/Network/API.swift
```diff
@@ -0,0 +1,136 @@
...
+
+    //: Prevent outside objects from creating their own instances of this class.
+    private override init() {
+        super.init()
+    }
+
+    // MARK: - Temporary Objective-C bridging functions
+
+    // TODO: remove these once migration is complete
+    @objc func blockchainAPI() -> NSString {
```
bchrisarriola: You can just return `String` here. When this method is bridged to Objective-C, the return value will be `NSString`.

Blockchain/Network/API.swift
```diff
@@ -0,0 +1,136 @@
...
+    @objc func shapeshift() -> NSString {
+        return Endpoints.shapeshift.rawValue as NSString
+    }
+}
+
+protocol Enumeratable: Hashable {
+    static var cases: [Self] { get }
+}
+
+extension Enumeratable {
```
bchrisarriola: Doesn't look like this is used anywhere in this file. Did you mean to use this in an upcoming commit?

Blockchain/Network/API.swift
```diff
@@ -0,0 +1,136 @@
...
+            index += 1
+            return item
+        }) {
+            cases.append(element)
+        }
+        return cases
+    }
+}
+
+protocol RawValued: Hashable, RawRepresentable {
```
bchrisarriola: This is nice! I imagine we might use it in another place in the codebase. Maybe move to its own file? (e.g. a file named `Enum` or something )

Blockchain/Network/Extensions/URL.swift
```diff
@@ -1,58 +1,58 @@
...
+//  Created by Maurice A. on 4/23/18.
 //  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
 //
 
 import Foundation
 
 @objc
-extension Bundle {
-    static var apiUrl: String? {
+extension API {
```
bchrisarriola: Not sure I like the naming convention here—the file is named "URL" but the file contains extensions for "API".

I noticed in our codebase we are just naming files that contain extensions the same name as the class, which works if we are creating extensions for code we didn't write, but obviously won't work if we wrote the class because of naming conflicts.

I've seen 2 naming conventions that I think work pretty well:
(1) file containing an extensions to a class is pluralized (e.g. so this would be `APIs.swift`) 
(2) file containing an extension is suffixed by `+Something` where `Something` are the extensions it adds to the class (e.g. so that would be `API+URL.swift`)

Personally, I like (2) as it clearly communicates what kind of extensions the file contains just by reading the file name.

Blockchain/Network/Extensions/UserAgent.swift
```diff
@@ -0,0 +1,24 @@
...
+//  Blockchain
+//
+//  Created by Maurice A. on 4/16/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+@objc
+extension NetworkManager {
```
bchrisarriola: Same comment here about the naming convention.

Blockchain/RootService.m
```diff
@@ -579,10 +577,10 @@ - (void)registerDeviceForPushNotifications
...
+    NSMutableURLRequest *notificationsRequest = [[NSMutableURLRequest alloc] initWithURL:[NSURL URLWithString:[NSString stringWithFormat:URL_PUSH_NOTIFICATIONS_SERVER_ARGUMENT_GUID_ARGUMENT_SHAREDKEY_ARGUMENT_TOKEN_ARGUMENT_LENGTH_ARGUMENT_LANGUAGE_ARGUMENT, [[API sharedInstance] walletUrl], [self.wallet guid], [self.wallet sharedKey], self.deviceToken, (unsigned long)[self.deviceToken length], languageString]]];
     [notificationsRequest setHTTPMethod:@"POST"];
 
-    NSURLSessionDataTask *dataTask = [[SessionManager sharedSession] dataTaskWithRequest:notificationsRequest completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
+    NSURLSessionDataTask *dataTask = [[[NetworkManager sharedInstance] session] dataTaskWithRequest:notificationsRequest completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
```
bchrisarriola: Will these changes also be migrated to `AppDelegate`? If so, we should comment out these lines so that we know it's been migrated? Ignore if you plan on doing that in a different pull request/commit.

Blockchain/Network/API.swift
achtenhagen: 👍

Blockchain/Network/API.swift
achtenhagen: Updated to BlockchainAPI 👍

Blockchain/Network/API.swift
achtenhagen: 👍

Blockchain/Network/Extensions/URL.swift
achtenhagen: I agree with (2), although we can't really name the extensions to my knowledge. I did update the file names however.

Blockchain/Network/Extensions/UserAgent.swift
achtenhagen: 👍

Blockchain/RootService.m
achtenhagen: Correct, they will be migrated elsewhere, but that isn't in the scope of this PR.

Blockchain/Network/API.swift
achtenhagen: Move to its own file.

Blockchain/Network/API.swift
achtenhagen: This just provides additional functionality to the `Enumerable` type by returning all cases of an `Enum` for example. I thought it could be useful for use in the future.

Blockchain/Extensions/Enum+Enumerable.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Enum+Enumerable.swift
+//  Blockchain
+//
+//  Created by Maurice A. on 4/24/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+protocol Enumeratable: Hashable {
```
bchrisarriola: File name should be `Enum+Enumeratable`

Blockchain/Network/Extensions/UserAgent.swift
bchrisarriola: Did you mean to rename this file to `NetworkManager+UserAgent.swift`?

PR #110, opened 2018-04-24
refactor(PushNotificationManager): creating PushNotificationManager.

bchrisarriola:
Created `PushNotificationManager` which manages everything related to push notifications (i.e. encapsulates the method for requesting for push permission and it sets itself as a `UNUserNotificationDelegate`).

Please review #109 first

----
**Feedback need for the following:** This PR also contains a `WalletManager` object which contains the `Wallet` instance (currently in `RootService`). It also sets itself as the `WalletDelegate` to that instance. Having wallet in `RootService` is currently blocking further refactoring of `RootService` as wallet is referenced in many different places—moving it to `WalletManager` will unblock that. 

Once we get started on an asset-based wallet architecture, we can then remove that instance from `WalletManager`. But, I do think it will remain to be a `WalletDelegate`.

----
**Side-note:** I think we should deprecate the concept of a `WalletDelegate`. Instead of having a single delegate to wallet calls, operations performed on a wallet should either: (1) accept callbacks, or (2) return Promises or a similar async construct (Observables, etc.). Deprecating this is a pretty large effort so let's stick to the current design for now until we have more time to address this.

### DIFF
--- a/Blockchain/Notifications/PushNotificationManager.swift
+++ b/Blockchain/Notifications/PushNotificationManager.swift
@@ -1,0 +1,94 @@
+//
+//  NotificationManager.swift
+//  Blockchain
+//
+//  Created by Chris Arriola on 4/23/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+/**
+ Manager object for push notifications for iOS versions < 10.0.
+ */
+class LegacyPushNotificationManager {
+    static let shared = LegacyPushNotificationManager()
+
+    func requestAuthorization() {
+        let notificationSettings = UIUserNotificationSettings(types: [.sound, .alert, .badge], categories: nil)
+        UIApplication.shared.registerUserNotificationSettings(notificationSettings)
+    }
+}
+
+/**
+ Manager object for push notifications
+ */
+@available(iOS 10.0, *)
+@objc
+class PushNotificationManager: NSObject {
+
+    static let shared = PushNotificationManager()
+
+    @objc class func sharedInstace() -> PushNotificationManager {
+        return shared
+    }
+
+    /// UNNotification that is to be presented to the user
+    @objc var presentingPushNotification: UNNotification?
+
+    /// Requests permission from the user to grant access to receive push notifications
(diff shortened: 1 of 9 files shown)


bchrisarriola:
Updated to add `WalletManager` and `PushNotificationManager` to BlockchainTests target

Blockchain/Notifications/PushNotificationManager.swift
```diff
@@ -0,0 +1,94 @@
...
+    ) {
+
+        print("User received remote notification")
+
+        let notificationContent = response.notification.request.content
+        let type = notificationContent.userInfo[Constants.PushNotificationKeys.userInfoType] as? String
+        let invitationSent = notificationContent.userInfo[Constants.PushNotificationKeys.userInfoId] as? String
+
+        if type == Constants.PushNotificationKeys.typeContactRequest {
+            // TODO set show type & initialize ContactsViewController
```
woudini: I should have mentioned this when we started, but Contacts was a feature that was never rolled out, and can be removed for now, or at the very least given very low priority. Side note: address book is not Contacts-related and should be kept. cc @tuzzolo

Blockchain/RootService.m
```diff
@@ -2111,6 +2111,7 @@ - (void)didGetMessagesOnFirstLoad
 
 - (void)didGetNewMessages:(NSArray *)newMessages
```
woudini: Documentation note: this method along with `loadContactsAndGetMessages`, `getMessages`, `reloadMessageViews`, `didGetMessages`, `didGetMessagesOnFirstLoad`, `setTransactionsViewControllerMessageIdentifier` and `messageIdentifier` are Contacts related. `receivedTransactionMessage` is unrelated (websocket-related) and should be kept.

Blockchain/Notifications/PushNotificationManager.swift
bchrisarriola: Good to know. I'll add a chore to remove that

Blockchain/RootService.m
bchrisarriola: Sweet. Less stuff to move over from RootService.

PR #111, opened 2018-04-24
Fix eth balance

woudini:
Currently `dev` returns 0 fiat balance for ETH because of comma grouping separators. This branch fixes that and adds grouping separators (e.g. thousands separator) to other fiat labels in the pie chart and legend keys. Grouping separators for crypto values should wait until a further release.

### DIFF
--- a/Blockchain/BCPriceMarker.swift
+++ b/Blockchain/BCPriceMarker.swift
@@ -108,7 +108,8 @@ open class BCPriceMarker: MarkerImage {
                 fatalError("Chart data entry has bad data")
         }
         guard let number = NumberFormatter.fiatString(from: entry.y) else {
-            fatalError("Could not generate number string from chart data entry")
+            print("Could not generate number string from chart data entry")
+            return
         }
         let labelText = String(format: "%@\n%@%@", currency, symbol, number)
         setLabel(labelText)
--- a/Blockchain/BCPriceMarker.swift
+++ b/Blockchain/BCPriceMarker.swift
@@ -107,7 +107,10 @@ open class BCPriceMarker: MarkerImage {
             let symbol = data["symbol"] as? String else {
                 fatalError("Chart data entry has bad data")
         }
-        let labelText = String(format: "%@\n%@%@", currency, symbol, String(entry.y))
+        guard let number = NumberFormatter.fiatString(from: entry.y) else {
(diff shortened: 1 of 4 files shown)


Blockchain/BCPriceMarker.swift
```diff
@@ -107,7 +107,10 @@ open class BCPriceMarker: MarkerImage {
             let symbol = data["symbol"] as? String else {
                 fatalError("Chart data entry has bad data")
         }
-        let labelText = String(format: "%@\n%@%@", currency, symbol, String(entry.y))
+        guard let number = NumberFormatter.fiatString(from: entry.y) else {
+            fatalError("Could not generate number string from chart data entry")
```
bchrisarriola: Using a `fatalError` seems a bit aggressive since that crashes the app if any of the above `guard` statements are false. Would it be better to fail gracefully here? (i.e. just print the error message and set the label to an empty string?)

Blockchain/DashboardViewController.m
```diff
@@ -355,7 +355,10 @@ - (double)getBtcBalance
 
 - (double)getEthBalance
 {
-    return [self doubleFromString:[NSNumberFormatter formatEthToFiat:[app.wallet getEthBalance] exchangeRate:app.wallet.latestEthExchangeRate]];
+    app.localCurrencyFormatter.usesGroupingSeparator = NO;
+    double result = [self doubleFromString:[NSNumberFormatter formatEthToFiat:[app.wallet getEthBalance] exchangeRate:app.wallet.latestEthExchangeRate]];
+    app.localCurrencyFormatter.usesGroupingSeparator = YES;
```
bchrisarriola: I think we should create a separate formatter, one that uses a grouping separator, and one that doesn't, vs. having to toggle that flag before and after. This is also error prone because it assumes that `app.localCurrencyFormatter.usesGroupingSeparator` is always set to `YES`.

Blockchain/BCPriceMarker.swift
woudini: Changed to print error to the console and show the currency only inside the `guard`

Blockchain/DashboardViewController.m
bchrisarriola: Ok. Let's address this in the refactor

PR #113, opened 2018-04-24
fix(Exchange): minimize number of quote requests and reset lastChange…

woudini:
…dField after getting quote

Lots of issues with this controller's architecture.


Blockchain/ExchangeCreateViewController.m
```diff
@@ -1082,6 +1102,15 @@ - (void)buildTrade
 
 #pragma mark - Helpers
 
+- (void)getQuoteAfterTyping
```
bchrisarriola: This is a little bit confusing. `getQuoteAfterTyping` is invoked as a result of a recurring scheduled timer and not after the user types which is what the name of the method implies. I think this should be changed for clarity so that either:

(1) this method is invoked after the user types
or 
(2) update the name of this method

woudini: changed to `getQuoteAfterTypingHasStopped`, thoughts?

bchrisarriola: @woudini if this method should be invoked after the user types, it's better to actually listen to an event when the user is done typing in a `UITextField` (i.e. using `textFieldDidEndEditting` and/or some other `UITextFieldDelegate` method).

But if this method needs to be invoked periodically, it should be named to something like `updateQuote`.

woudini: It shouldn't go in a text field delegate method because it will build up a queue of network calls to `getApproximateQuote` that can't be cancelled when the user starts typing again. The point of this kind of implementation is to constantly put off a single network call until a certain period of time has elapsed after the last text field change.

There's already a method called `getApproximateQuote` which would make `updateQuote` confusing.

bchrisarriola: > It shouldn't go in a text field delegate method because it will build up a queue of network calls to getApproximateQuote that can't be cancelled when the user starts typing again. 

It doesn't have to queue up network calls. You can do something like:

```
- (BOOL)textField:(UITextField *)textField shouldChangeCharactersInRange:(NSRange)range replacementString:(NSString *)string
{
    // ...

    [quoteTimer invalidate];
    quoteTimer = [NSTimer timerWithTimeInterval:0.5 repeats:NO block:^(NSTimer * _Nonnull timer) {
        // Invoke the method here
        quoteTimer = nil;
    }];

    return YES;
}
```

...this way, a timer task is not being executed when it's no longer needed.

woudini: Thanks, this is much cleaner.

PR #114, opened 2018-04-25
Refactor

achtenhagen:
This PR continues to refactor the root service by migrating `symbolLocal` to BlockchainSettings and porting over `persistServerSessionIDForNewUIWebViews` and `disableUIWebViewCaching` to Swift.

These  two functions should be moved into a future web view wrapper. Also note that `toggleSymbol` and `reloadSymbols` are only partially ported at this time.


bchrisarriola:
Oh just read this:

> These two functions should be moved into a future web view wrapper.

Lmk if you think my suggestion to move to `NetworkManager` is fine + adding a TODO to create a web view wrapper of some sort.

achtenhagen:
@bchrisarriola I think we can move it into the Network Manager (I can do that now as I am currently working on that branch) because we are not planning to use web views in the long run. It probably makes sense to put all URL (NSURL) related code into the Network Manager.

PR #115, opened 2018-04-25
fix(ConfirmPaymentView): reduce row size if using 4S and more than fo…

woudini:
…ur rows


bchrisarriola:
lgtm. will leave up to @achtenhagen to merge since he's testing it

PR #117, opened 2018-04-25
refactor(create wallet): create wallet flow.

bchrisarriola:
Moving over create wallet flow to new arch.

Summary of changes:
* moved `WalletDelegate` methods that are related to create wallet flow to `WalletManager` (the new WalletDelegate)
* created extensions for `Reachability` & `HTTPCookieStorage`
* created `UIApplication+Suspend.h/.m` extension. This needs to be in Obj-c.
* moved various constants to `LocalizationConstants` & `Constants`

...still need to move over pin create/entry to complete the create wallet flow.


Bethannfrum:
Fix and restore to the original format and release old documents

Blockchain/Settings/BlockchainSettings.swift
```diff
@@ -83,11 +83,35 @@ final class BlockchainSettings: NSObject {
...
             // Private initializer so that `shared` and `sharedInstance` are the only ways to
             // access an instance of this class.
             super.init()
         }
+
+        func clearPin() {
+            encryptedPinPassword = nil
+            pinKey = nil
+            passwordPartHash = nil
+        }
```
woudini: `lastEnteredPin = 0000` may not belong here, but as a documentation note: for when a user changes their PIN in order to prevent them from reusing the one they currently have. `0000` is currently reserved (can't remember why, I think for a server message) and is not allowed as a valid PIN. When the PIN is cleared `lastEnteredPIN` is set to 0000 to allow all possible PIN combinations https://github.com/blockchain/My-Wallet-V3-iOS/commit/a31587667cd982f53b4cb74afe65a941e33ad5f1

Blockchain/Wallet/WalletManager.swift
```diff
@@ -15,16 +15,100 @@ import Foundation
...
+        //Becuase we are not storing the password on the device. We record the first few letters of the hashed password.
+        //With the hash prefix we can then figure out if the password changed
+        guard let password = wallet.password,
+            let passwordSha256 = NSString(string: password).sha256(),
+            let passwordPartHash = BlockchainSettings.App.shared.passwordPartHash else {
+            return
+        }
+
+        let endIndex = passwordSha256.index(passwordSha256.startIndex, offsetBy: min(password.count, 5))
+        if passwordSha256[..<endIndex] == passwordPartHash {
```
woudini: I think this should be `!=`

Blockchain/SettingsTableViewController.m
```diff
@@ -702,18 +703,19 @@ - (void)alertUserToChangeTwoStepVerification
 
 - (void)changeTwoStepVerification
 {
-    if ([app checkInternetConnection]) {
+    if (!Reachability.hasInternetConnection) {
+        [AlertViewPresenter.sharedInstance showNoInternetConnectionAlert];
```
woudini: Should be a `return` here

Blockchain/BCCreateAccountView.m
```diff
@@ -58,29 +58,31 @@ -(id)init
...
-            // TODO i18n
-            [[AlertViewPresenter sharedInstance] standardNotifyWithMessage:BC_STRING_LABEL_MUST_HAVE_LESS_THAN_18_CHAR title:BC_STRING_ERROR];
-            return;
-        }
-        
-        if (![app.wallet isAccountNameValid:label]) {
-            return;
-        }
+    if (!Reachability.hasInternetConnection) {
+        [AlertViewPresenter.sharedInstance showNoInternetConnectionAlert];
```
woudini: should be a `return` here

bchrisarriola: done

PR #118, opened 2018-04-25
fix(Dashboard): remove chart marker on reload to prevent persistence …

woudini:
…after returning from background


bchrisarriola:
Since `hideChartMarker` is no longer used by an external class—should it be removed from `BCBalancesChartView.h`?

woudini:
Yes, removed.

PR #119, opened 2018-04-25
refactor(BlockchainSettings): moving guid and sharedKey to BlockchainSettings.Wallet

bchrisarriola:
So that the application-level never has to interface with the keychain.

Also, should we put other properties that are currently in `BlockchainSettings.App` in here (e.g. `passwordPartHash`)?

### DIFF
--- a/Blockchain/Wallet/WalletManager.swift
+++ b/Blockchain/Wallet/WalletManager.swift
@@ -93,9 +93,8 @@ extension WalletManager: WalletDelegate {
             return
         }
 
-        // TODO put this in BlockchainSettings
-        KeychainItemWrapper.setGuidInKeychain(guid)
-        KeychainItemWrapper.setSharedKeyInKeychain(sharedKey)
+        BlockchainSettings.App.shared.guid = guid
+        BlockchainSettings.App.shared.sharedKey = sharedKey
 
         //Becuase we are not storing the password on the device. We record the first few letters of the hashed password.
         //With the hash prefix we can then figure out if the password changed
--- a/Blockchain/AppDelegate.swift
+++ b/Blockchain/AppDelegate.swift
@@ -216,7 +216,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     // TODO: move to BlockchainSettings
     func checkForNewInstall() {
         if !BlockchainSettings.App.shared.firstRun {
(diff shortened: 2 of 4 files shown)


Bethannfrum:
Fix and restore to original format

Blockchain/Wallet/WalletManager.swift
```diff
@@ -93,9 +93,8 @@ extension WalletManager: WalletDelegate {
             return
         }
 
-        // TODO put this in BlockchainSettings
-        KeychainItemWrapper.setGuidInKeychain(guid)
-        KeychainItemWrapper.setSharedKeyInKeychain(sharedKey)
+        BlockchainSettings.WalletInfo.shared.guid = guid
+        BlockchainSettings.WalletInfo.shared.sharedKey = sharedKey
 
         //Becuase we are not storing the password on the device. We record the first few letters of the hashed password.
```
Bethannfrum: Fix and restore to the original format

Bethannfrum:
Store password on the device to

Blockchain/Wallet/WalletManager.swift
Bethannfrum: Fix So everyone can use put back to original document and release old documents

bchrisarriola:
> It might actually make more sense for guid and sharedKey to go in App. When WalletInfo is finished its setters and getters will interact with the wallet in the JS and essentially modify the wallet so that when you log in on web or Android those changes will persist. It still seems ambiguous because the guid and shared key are also parts of the wallet model, but the app needs to have them stored outside of the model. Thoughts?

Got it, yeah in that case it makes more sense to put it in `BlockchainSettings.App`, I'll update this PR and make that change. 

As far as obtaining a reference to `guid`/`sharedKey`, I think the app should have only one source for retrieving those values, and since the app needs those properties outside of the `Wallet` modal, it's best to access those properties through the app's mechanism for storing that info (i.e. through `BlockchainSettings.App`). 

So how that would work is that `Wallet` properties (`guid`, `sharedKey`, etc.) would only be interacted with primarily when logging in/logging out and won't be accessed outside of those scopes (i.e. those should be private). Upon an authentication change, properties such as `guid` and `sharedKey` should be passed in a callback and persisted to the keychain by the callback.

PR #122, opened 2018-04-26
Network manager

achtenhagen:
This PR migrates `checkForUnusedAddress` to Swift using the new `AssetAddress` type.

TODO: move `checkForUnusedAddress` to a more appropriate place than the application delegate.


Blockchain/AppDelegate.swift
```diff
@@ -259,4 +261,66 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            DispatchQueue.main.async {
+                let isUnused = transactions.count == 0
+                successHandler(isUnused)
+            }
+        })
+        task.resume()
+    }
+
+    // TODO: move to appropriate controller
+    func checkForUnusedAddress(_ address: BitcoinCashAddress,
```
bchrisarriola: Duplicate methods in here. In the spirit of DRY, let's change the method signature of this to: 
```
func checkForUnusedAddress(_ address: AssetAddress, successHandler: @escaping UnusedAddressSuccess, errorHandler: @escaping UnusedAddressError) {
```

We'd also need to change to `BlockchainAPI.suffixURL(address: AssetAddress)`.

Blockchain/Network/Extensions/BlockchainAPI+URLSuffix.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Blockchain
+//
+//  Created by Maurice A. on 4/26/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+extension BlockchainAPI {
+    static func suffixURL(btcAddress: BitcoinAddress) -> String? {
```
bchrisarriola: Change from `static` to an instance method and you can access the `walletUrl` directly—much more test friendly.

Blockchain/Assets/AssetAddress.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Blockchain
+//
+//  Created by Maurice A. on 4/26/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+/// Blueprint for creating and validating asset addresses.
+protocol AssetAddress {
```
bchrisarriola: `public`? 😉

Same comment goes for `BitcoinAddress` and `BitcoinCash`

Blockchain/Assets/AssetAddress.swift
```diff
@@ -0,0 +1,17 @@
...
+    /// String representation of the address.
+    var address: String? { get }
+    init(string: String)
+    static func isValid(_ address: String) -> Bool
```
bchrisarriola: Probably best not `static`.

BlockchainTests/BlockchainAPI+URLSuffixTests.swift
```diff
@@ -0,0 +1,55 @@
...
+        super.tearDown()
+    }
+
+    // MARK: - Bitcoin
+
+    func testSuffixURLWithValidBitcoinAddress() {
+        let btcAddress = BitcoinAddress(string: "1W3hBBAnECvpmpFXcBrWoBXXihJAEkTmA")
+        let url = BlockchainAPI.suffixURL(btcAddress: btcAddress)
+        let expected = "https://blockchain.info/address/\(btcAddress.address!)?format=json"
+        XCTAssertNotNil(url, "Expected the url to be \(expected), but got nil.")
```
bchrisarriola: 🔥

Blockchain/Assets/AssetAddress.swift
```diff
@@ -0,0 +1,18 @@
...
+//  Created by Maurice A. on 4/26/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+/// Blueprint for creating and validating asset addresses.
+public protocol AssetAddress {
+    /// String representation of the address.
+    var description: String? { get }
```
bchrisarriola: Don't we always want this to be non-optional?

Blockchain/Assets/AssetAddress.swift
```diff
@@ -0,0 +1,18 @@
...
+    var assetType: AssetType { get }
+    init(string: String)
```
bchrisarriola: instead of `string`, something more descriptive (i.e. `_ address`)?

also, if we switch description to not be optional, this can be an optional init (i.e. will return nil if the provided address is invalid)

achtenhagen: I decided to use `string` as the parameter name to make it clear that an address is instantiated by a string representation of an address, not an address itself (that's also why I changed address to description). This is based on how `URL(string: String)` is instantiated.

I will make the `description` optional to support the optional initialize. Good point!

bchrisarriola: > I decided to use string as the parameter name to make it clear that an address is instantiated by a string representation of an address, not an address itself (that's also why I changed address to description). This is based on how URL(string: String) is instantiated.

👍 

> I will make the description optional to support the optional initialize. Good point!

You mean _non_-optional?

PR #123, opened 2018-04-26
refactor(AuthenticationCoordinator): Move pin entry flow to AuthenticationCoordinator.

bchrisarriola:
Summary:
* Created `AuthenticationCoordinator` and move `PEPinEntryViewController` flow in here
* Migrated `PEPinEntryViewController` pin delegate methods in `AuthenticationCoordinator`
* Created `Pin` class

Note: pin entry flow has not yet been fully migrated (i.e. pin-related wallet delegate calls, touch ID, etc.). Will work on that in a separate PR.

### DIFF
--- a/Blockchain/AuthenticationCoordinator.swift
+++ b/Blockchain/AuthenticationCoordinator.swift
@@ -1,0 +1,307 @@
+//
+//  AuthenticationCoordinator.swift
+//  Blockchain
+//
+//  Created by Chris Arriola on 4/25/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+@objc class AuthenticationCoordinator: NSObject, Coordinator {
+
+    @objc static let shared = AuthenticationCoordinator()
+
+    var lastEnteredPIN: Pin?
+
+    private(set) var pinEntryViewController: PEPinEntryController?
+
+    // TODO: loginTimout is never invalidated after a successful login
+    private var loginTimeout: Timer?
+
+    private var pinViewControllerCallback: ((Bool) -> Void)?
+
+    private var isPinEntryModalPresented: Bool {
+        let rootViewController = UIApplication.shared.keyWindow!.rootViewController!
+        let tabControllerManager = AppCoordinator.shared.tabControllerManager
+        return !(pinEntryViewController == nil ||
+            pinEntryViewController!.isBeingDismissed ||
+            !pinEntryViewController!.view.isDescendant(of: rootViewController.view) ||
+            tabControllerManager.tabViewController.presentedViewController != pinEntryViewController)
+    }
+
+    // MARK: - Public
+
+    @objc func start() {
+        if WalletManager.shared.wallet.isNew {
+            startNewWalletSetUp()
+        } else {
+            showPinEntryView(asModal: false)
+        }
+    }
+
+    @objc func startNewWalletSetUp() {
+        let setUpWalletViewController = WalletSetupViewController(setupDelegate: self)!
+        let topMostViewController = UIApplication.shared.keyWindow?.rootViewController?.topMostViewController
+        topMostViewController?.present(setUpWalletViewController, animated: false) { [weak self] in
+            self?.showPinEntryView(asModal: false)
+        }
+    }
+
+    // Closes the pin entry modal, if presented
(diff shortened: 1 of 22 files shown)


Blockchain/AuthenticationCoordinator.swift
```diff
@@ -0,0 +1,312 @@
...
+                timeInterval: TimeInterval(seconds),
+                target: self,
+                selector: #selector(showLogInError),
+                userInfo: nil,
+                repeats: false
+            )
+        }
+    }
+
+    @objc private func showLogInError() {
```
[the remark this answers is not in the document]
bchrisarriola: Not sure I follow. What do you mean by that?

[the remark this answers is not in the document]
woudini: the letter `i` should be lowercase

bchrisarriola:
Ah, didn't catch the lowercased `i` in your comment. Should be fixed.

Blockchain/AuthenticationCoordinator.swift
```diff
@@ -0,0 +1,312 @@
...
+        // and as a viewController. This checks which one it is and dismisses accordingly
+        let rootViewController = UIApplication.shared.keyWindow!.rootViewController!
+        if pinEntryViewController.view.isDescendant(of: rootViewController.view) {
+            pinEntryViewController.view.removeFromSuperview()
+        } else {
+            let tabControllerManager = AppCoordinator.shared.tabControllerManager
+            if WalletManager.shared.wallet.isNew {
+                tabControllerManager.tabViewController.presentedViewController?.dismiss(animated: animated)
+            } else {
+                tabControllerManager.tabViewController.presentedViewController?.dismiss(animated: animated)
```
woudini: I think this `dismiss` should be called on `tabControllerManager.tabViewController` here and not `tabControllerManager.tabViewController.presentedViewController`

Blockchain/Pin/Pin.swift
```diff
@@ -0,0 +1,83 @@
...
+    var toString: String {
+        return pinCode.pinToString
+    }
+
+    init(code: UInt) {
+        self.pinCode = code
+    }
+
+    // TODO: not the best place for this - move elsewhere
+    func save() throws {
```
achtenhagen: On my local branch for the address crypto functions, I added a file called `Crypto.swift`. Once that file is committed and pushed, this may be a good place to put this function in.

Blockchain/Pin/Pin.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    var hashValue: Int {
+        return Int(pinCode)
+    }
+}
+
+extension UInt {
+    var pinToString: String {
+        return NSString(format: "%lu" as NSString, CUnsignedLongLong(self)) as String
```
achtenhagen: Can this just be `String(format: String, CVarArg...)` instead of `NSString` and then casting it to `String`?

achtenhagen:
Good to merge once my comments have been reviewed.

Blockchain/Pin/Pin.swift
bchrisarriola: K, good to know 👌🏾

Blockchain/Pin/Pin.swift
bchrisarriola: Yep—let me update

Blockchain/AuthenticationCoordinator.swift
bchrisarriola: Good catch

bchrisarriola: Good catch

woudini: I think the reason for the `isNew` if statement branching is because of how the `tabViewController` presents the `WalletSetupViewController` (enabling touch ID, verifying email immediately after creating a wallet). It seems that when creating a new wallet the PIN screen is presented by the `WalletSetupViewController` instead of the `tabViewController`. This can be simplified depending on how the PIN screen is changed but dismissing might fail when the wallet is newly created.

bchrisarriola: My understanding is that in both cases we want to dismiss the `pinEntryViewController`. If that's the case, `pinEntryViewController.dismiss()` should accomplish that regardless of the presenting view controller.

PR #125, opened 2018-04-26
fix(Exchange): use iOS 9 compatible class method for timer initializa…

woudini:
…tion

### DIFF
--- a/Blockchain/ExchangeCreateViewController.m
+++ b/Blockchain/ExchangeCreateViewController.m
@@ -693,12 +693,7 @@
     
     [self disablePaymentButtons];
     
-    __weak ExchangeCreateViewController *weakSelf = self;
-    self.quoteTimer = [NSTimer scheduledTimerWithTimeInterval:1.0 repeats:NO block:^(NSTimer * _Nonnull timer) {
-        [weakSelf disablePaymentButtons];
-        [weakSelf getApproximateQuote];
-        weakSelf.quoteTimer = nil;
-    }];
+    self.quoteTimer = [NSTimer scheduledTimerWithTimeInterval:1.0 target:self selector:@selector(getApproximateQuoteAfterTimer) userInfo:nil repeats:NO];
 }
 
 - (void)doCurrencyConversion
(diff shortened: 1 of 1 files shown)


Blockchain/ExchangeCreateViewController.m
```diff
@@ -1028,13 +1023,17 @@ - (void)getRate
 
 - (void)getApproximateQuote
 {
+    [self disablePaymentButtons];
+    
+    self.quoteTimer = nil;
+    
+    [self cancelCurrentDataTask];
```
bchrisarriola: Intentionally moved up the if statement?

Blockchain/ExchangeCreateViewController.m
```diff
@@ -693,12 +693,7 @@ - (void)doCurrencyConversionAfterTyping
     
     [self disablePaymentButtons];
     
-    __weak ExchangeCreateViewController *weakSelf = self;
-    self.quoteTimer = [NSTimer scheduledTimerWithTimeInterval:1.0 repeats:NO block:^(NSTimer * _Nonnull timer) {
-        [weakSelf disablePaymentButtons];
-        [weakSelf getApproximateQuote];
-        weakSelf.quoteTimer = nil;
-    }];
+    self.quoteTimer = [NSTimer scheduledTimerWithTimeInterval:1.0 target:self selector:@selector(getApproximateQuote) userInfo:nil repeats:NO];
```
bchrisarriola: I think we should be creating a separate method here that gets invoked when the timer is scheduled to trigger so that it's functionally equivalent to the execution order:

```
[self disablePaymentButtons];
[self getApproximateQuote];
self.quoteTimer = nil;
```

By adding the additional method calls in `getApproximateQuote`, it:
1. makes it unclear whether those actions should be performed every time `getApproximateQuote` is called (from the previous change, it seems like that's not always the case)
2. adds several responsibilities to `getApproximateQuote` which creates potentially unwanted side-effects

Blockchain/ExchangeCreateViewController.m
woudini: I had moved it down, but no longer an issue with the changes

Blockchain/ExchangeCreateViewController.m
woudini: Agree that this is cleaner. Committed changes.

PR #127, opened 2018-04-27
feat(Wallet): add JSContextWithDOM class

woudini:
JavaScriptCore is pure JavaScript - it lacks functions that are normally part of the DOM.
`JSContextWithDOM` is intended to define these missing functions in order to make it compatible with JS libraries that use them.

Note: `.pbxproj` file is acting strange when I add a new Swift file.

### DIFF
--- a/Blockchain/Wallet/JSContextWithDOM.swift
+++ b/Blockchain/Wallet/JSContextWithDOM.swift
@@ -9,6 +9,8 @@
 import Foundation
 import JavaScriptCore
 
+/// JavaScriptCore is pure JavaScript - it lacks functions that are normally part of the DOM.
+/// This class is intended to define these missing functions in order to make it compatible with JS libraries that use them.
 class JSContextWithDOM: JSContext {
 
     override init() {
@@ -27,7 +29,12 @@ class JSContextWithDOM: JSContext {
         setObject(ModuleXMLHttpRequest.self, forKeyedSubscript: "XMLHttpRequest" as NSString)
     }
 
-    func setupConsoleFunctions() {
+    required override init!(virtualMachine: JSVirtualMachine!) {
+        super.init(virtualMachine: virtualMachine)
+    }
+
+    private func setupConsoleFunctions() {
+        // TODO: properly define console functions
         let consoleNames = ["log",
                             "debug",
                             "info",
@@ -46,15 +53,15 @@ class JSContextWithDOM: JSContext {
                             "profileEnd"]
 
         consoleNames.forEach { name in
-            let consoleLog: (String) -> Void = { message in
+            let consoleLog: @convention(block) (String) -> Void = { message in
                 print("Javascript \(name): \(message)")
             }
             self.objectForKeyedSubscript("console").setObject(consoleLog, forKeyedSubscript: name as NSString)
         }
     }
 
-    func setupExceptionHandler() {
-        let exceptionHandler: (JSContext?, JSValue?) -> Void = { context, exception in
+    private func setupExceptionHandler() {
+        self.exceptionHandler = { context, exception in
             guard let exception = exception else {
                 print("Exception handler error: could not get exception")
                 return
@@ -65,31 +72,34 @@ class JSContextWithDOM: JSContext {
             }
             guard let stackTrace = exception.objectForKeyedSubscript("stack").toString() else {
                 print("Exception handler error: could not get stack trace")
+                print("Message: \(message)")
                 return
             }
             guard let lineNumber = exception.objectForKeyedSubscript("line").toString() else {
                 print("Exception handler error: could not get line number")
+                print("Message: \(message) \nstack:\(stackTrace)")
                 return
             }
             print("\(message) \nstack: \(stackTrace)\nline number: \(lineNumber)")
         }
-        self.exceptionHandler = exceptionHandler
     }
 
-    var timers = [String: Timer]()
+    /// A dictionary containing the timeout/interval timers and their corresponding identifiers
+    private var timers = [String: Timer]()
 
-    lazy var removeTimer: (String) -> Void = { [unowned self] identifier in
+    private lazy var removeTimer: @convention(block) (String) -> Void = { [unowned self] identifier in
         if let timer = self.timers[identifier] {
             timer.invalidate()
             self.timers.removeValue(forKey: identifier)
         }
     }
 
-    func getAddTimer(repeats: Bool) -> (JSValue?, Double) -> String {
-        return { [unowned self] callback, timeout in
+    private func getAddTimer(repeats: Bool) -> @convention(block) (JSValue?, Double) -> String {
+        let addTimer: @convention(block) (JSValue?, Double) -> String = { [unowned self] callback, timeout in
+            let jsQueue = DispatchQueue(label: "jsqueue")
             let uuid = NSUUID().uuidString
-            let blockOperation = BlockOperation.init(block: {
-                DispatchQueue.main.async {
+            let blockOperation = BlockOperation(block: { [unowned self] in
+                jsQueue.async {
                     if self.timers[uuid] != nil {
                         _ = callback?.call(withArguments: nil)
                     }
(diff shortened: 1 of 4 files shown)


Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+        setupSetTimeout()
+        setupClearTimeout()
+
+        setupSetInterval()
+        setupClearInterval()
+
+        setObject(ModuleXMLHttpRequest.self, forKeyedSubscript: "XMLHttpRequest" as NSString)
+    }
+
+    func setupConsoleFunctions() {
```
bchrisarriola: All of these setup methods can be `private`

Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+                            "time",
+                            "timeEnd",
+                            "count",
+                            "trace",
+                            "profile",
+                            "profileEnd"]
+
+        consoleNames.forEach { name in
+            let consoleLog: (String) -> Void = { message in
+                print("Javascript \(name): \(message)")
```
bchrisarriola: We should actually convert these methods correctly vs. having all of these methods behave as if they were console.logs.

Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+            guard let exception = exception else {
+                print("Exception handler error: could not get exception")
+                return
+            }
+            guard let message = exception.toString() else {
+                print("Exception handler error: could not get message")
+                return
+            }
+            guard let stackTrace = exception.objectForKeyedSubscript("stack").toString() else {
+                print("Exception handler error: could not get stack trace")
```
bchrisarriola: There is a valid `message` at this point, should be printed as well.

Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+                return
+            }
+            guard let lineNumber = exception.objectForKeyedSubscript("line").toString() else {
+                print("Exception handler error: could not get line number")
```
bchrisarriola: There is a valid `message` & `stackTrace` at this point, should be printed as well.

Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+                return
+            }
+            print("\(message) \nstack: \(stackTrace)\nline number: \(lineNumber)")
+        }
+        self.exceptionHandler = exceptionHandler
+    }
+
+    var timers = [String: Timer]()
```
bchrisarriola: `private`

Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+        consoleNames.forEach { name in
+            let consoleLog: (String) -> Void = { message in
+                print("Javascript \(name): \(message)")
+            }
+            self.objectForKeyedSubscript("console").setObject(consoleLog, forKeyedSubscript: name as NSString)
+        }
+    }
+
+    func setupExceptionHandler() {
+        let exceptionHandler: (JSContext?, JSValue?) -> Void = { context, exception in
```
bchrisarriola: Doesn't seem like we need an extra variable here. Can set this directly to `self.exceptionHandler`

Blockchain/Wallet/JSContextWithDOM.swift
bchrisarriola: Also would be helpful to document what the `key` of this dictionary is supposed to represent (i.e. an identifier for the timer)

Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+        if let timer = self.timers[identifier] {
+            timer.invalidate()
+            self.timers.removeValue(forKey: identifier)
+        }
+    }
+
+    func getAddTimer(repeats: Bool) -> (JSValue?, Double) -> String {
+        return { [unowned self] callback, timeout in
+            let uuid = NSUUID().uuidString
+            let blockOperation = BlockOperation.init(block: {
```
bchrisarriola: Not very Swift-life to invoke the initializer as a method call. Better to use the pattern:
```
BlockOperation(block: {
  // ...
})
```

Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+                DispatchQueue.main.async {
```
bchrisarriola: Is dispatching to the main queue necessary? I think there's a retain cycle here as well, use `[unowned/weak self]`

Blockchain/Wallet/JSContextWithDOM.swift
```diff
@@ -0,0 +1,124 @@
...
+//  Blockchain
+//
+//  Created by kevinwu on 4/27/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+import JavaScriptCore
+
+class JSContextWithDOM: JSContext {
```
bchrisarriola: I feel like a subclass is potentially limiting and it might be a better implementation to define these as an extension to `JSContext`. It also allows us to modularly inject these functions as needed. 

I imagine we can do something like:

```
let context = JSContext()
context.injectNativeConsoleFunctions()
context.injectNativeTimeout()
// etc.
```

woudini: It was originally an extension but when I added `timers` as a property the compiler errored with `Extensions must not contain stored properties`

bchrisarriola: Oh right. I think the right approach then would be to:

(1) create a subclass that is specific to timing (i.e. something like `JSTimingContext`) which contain the `timers` property + associated methods
(2) create an extension on `JSContext` that contains the methods that inject `console.*` implementations, exception handling, etc.

Blockchain/Wallet/JSContextWithDOM.swift
achtenhagen: What is specifically meant by converting them correctly?

bchrisarriola: My understanding is that we want a `console.*` function that is invoked in Javascript behave as if it were invoked natively. So, if you call something like `console.group`, it will actually indent the next several invocations of `console.log/info/debug/etc.`. Right now, this isn't the case and we're just logging that a console function was invoked. If that's actually the behavior we want, then we should make that clear both in the documentation of this method and also the method name should clearly communicate that intent.

Blockchain/Wallet/JSContextWithDOM.swift
woudini: What advantages does using an extension offer in this case? To me a single class is more cohesive and abstract. I don't see why whatever is initializing this `JSContext` would need to know what to inject or only inject some functions and not others.

bchrisarriola: One is more modular and reusable in contexts outside of `Wallet`, while the other is not. In addition, it's also a lot more clear and concise what the component creating this `JSContext` needs.

Leaving the responsibility of what to inject to the component creating this `JSContext` prevents bloating up the `JSContext` with functions it might not even need. While `Wallet` (and perhaps the future rearchitected `Wallet`) needs these functions, this doesn't have to be the case.

mpfluger: Just my 2 cents: I don't think we will use the JSContext except for the communication with the JS wallet. In that case I'm not sure that decoupling the class from the setup is a better architecture. We can always revisit that later on if it does become an issue.

bchrisarriola: Sounds good 👍

Blockchain/Wallet/JSContextWithDOM.swift
woudini: Changed to private

Blockchain/Wallet/JSContextWithDOM.swift
woudini: Changed to TODO

Blockchain/Wallet/JSContextWithDOM.swift
woudini: Printed message

Blockchain/Wallet/JSContextWithDOM.swift
woudini: Printed stack trace

Blockchain/Wallet/JSContextWithDOM.swift
woudini: Made private and added comment

Blockchain/Wallet/JSContextWithDOM.swift
woudini: set directly

Blockchain/Wallet/JSContextWithDOM.swift
woudini: changed to `BlockOperation(block: {})`

Blockchain/Wallet/JSContextWithDOM.swift
woudini: Changed to separate queue `jsQueue` based on https://stackoverflow.com/a/34520575

bchrisarriola: Interesting. Was this related to tests failing when trying to call the interval/timeout functions?

PR #130, opened 2018-04-28
refactor(password): password entry

bchrisarriola:
* Moving password entry flow from RootService into AuthenticationCoordinator/AuthenticationManager
* Also consolidated the auth handler for new + existing wallets.

Note: Please pay close attention to `WalletAuthDelegate` and `AuthenticationManager.Handler` - when used in conjunction it allows receiving auth-specific wallet delegate methods.


bchrisarriola:
@woudini yep, this is a flow that should for sure be tested. If it makes it easier for reviewing, I can address that in a separate pull request. Any feedback on `WalletAuthDelegate`? Does that delegate capture all of the auth-related methods from `WalletDelegate`?

bchrisarriola:
Added tests in #132

woudini:
Looks good. We should keep in mind that parts of the implementation of `walletFailedToDecrypt` and `walletFailedToLoad` may eventually need to be in `WalletAuthDelegate`

bchrisarriola:
> Looks good. We should keep in mind that parts of the implementation of walletFailedToDecrypt and walletFailedToLoad may eventually need to be in WalletAuthDelegate

Yep, I'll move those in a separate PR (haven't commented out the implementations in `RootService` as a reminder)

Blockchain/Authentication/AuthenticationManager.swift
```diff
@@ -102,11 +111,27 @@ final class AuthenticationManager: NSObject {
     /**
      The function used to authenticate the user using a provided passcode.
      - Parameters:
-        - passcode: The passcode used for authenticating the user.
+        - payload: The passcodepayload used for authenticating the user.
```
achtenhagen: Should this be passcode payload instead of passcodepayload?

bchrisarriola:
Addressed comment @achtenhagen.

I'll have to update the base for this PR once #123 is merged.

PR #132, opened 2018-04-30
test(AuthenticationManager): tests for auth with passcode.

bchrisarriola:


Blockchain/Alerts/AlertViewPresenter.swift
```diff
@@ -24,7 +24,7 @@ import Foundation
     @objc func showKeychainReadError() {
         // TODO
 //        UIAlertController *alert = [UIAlertController alertControllerWithTitle:BC_STRING_FAILED_TO_LOAD_WALLET_TITLE
-        // message:BC_STRING_ERROR_LOADING_WALLET_IDENTIFIER_FROM_KEYCHAIN preferredStyle:UIAlertControllerStyleAlert];
+        // message:Blockchain.xcodeproj/project.pbxproj preferredStyle:UIAlertControllerStyleAlert];
```
woudini: Does this change mean anything?

Blockchain/Authentication/Views/PasswordRequiredView.swift
```diff
@@ -14,7 +14,7 @@ protocol PasswordRequiredViewDelegate: class {
 
 class PasswordRequiredView: UIView {
     @IBOutlet weak var labelHeader: UILabel!
-    @IBOutlet weak var textFieldPassword: BCTextField!
+    @IBOutlet weak var textFieldPassword: UITextField!
```
woudini: I think I was seeing a compiler issue with this line when trying to run tests, but it's extremely important for the password field to be a `BCTextField` because it disables autocorrect. See https://github.com/nowsecure/secure-mobile-development/blob/master/en/caching-logging/be-aware-of-the-keyboard-cache.md cc @achtenhagen

Blockchain/Alerts/AlertViewPresenter.swift
bchrisarriola: Not even sure how that got in there. Will revert.

Blockchain/Authentication/Views/PasswordRequiredView.swift
bchrisarriola: Yeah, I change this class exactly for that reason (i.e. tests failing). But, the actual view in memory is still a `BCTextField` (as specified in the .xib)

bchrisarriola: Ok figured it out. Problem was that `BCTextField` and `BCSecureTextField` were not in `BlockchainTests` target membership.

woudini: Awesome thanks! Was scratching my head for a while on that one.

PR #133, opened 2018-05-01
refactor(pin): Complete pin entry flow on new wallet creation.

bchrisarriola:
* Moved pin creation from RootService to AuthenticationCoordinator
* Created `GetPinResponse` & `PutPinResponse` model objects to wrap pin-related API responses
* Created `WalletPinEntryDelegate` to wrap `WalletDelegate` pin-related methods

Please review/merge PRs all PRs < #132 before reviewing this

PR #135, opened 2018-05-02
Check for maintenance

achtenhagen:
This PR migrates the `checkForMaintenance` function to Swift, updates localizations and introduces the wallet options enum.


Blockchain/Alerts/AlertViewPresenter.swift
```diff
@@ -53,6 +53,10 @@ import Foundation
         }
     }
 
+    func showMaintenanceAlert(withTitle title: String, _ message: String) {
```
bchrisarriola: This method might not actually be necessary. Can call `standardNotify(message,title)` directly

Blockchain/Authentication/AuthenticationCoordinator+TouchID.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    // TODO: migrate to the responsible controller that prompts for authentication
+    func handleBiometricAuthenticationError(with error: AuthenticationError) {
+        if let description = error.description {
+            let alert = UIAlertController(title: LocalizationConstants.Errors.error, message: description, preferredStyle: .alert)
+            let action = UIAlertAction(title: LocalizationConstants.ok, style: .default, handler: nil)
+            alert.addAction(action)
+            DispatchQueue.main.async {
+                UIApplication.shared.keyWindow?.rootViewController?.present(alert, animated: true, completion: nil)
+            }
```
bchrisarriola: Can also just call `AlertViewPresenter.shared.standardNotify` directly

Blockchain/Authentication/AuthenticationCoordinator.swift
```diff
@@ -133,11 +138,21 @@ import Foundation
...
+//                [self authenticateWithTouchID];
+//            }
+//            #endif
+            authenticateWithBiometrics()
+            //
         } else {
-            // TODO: check for maintenance
-//            [self checkForMaintenance];
+            NetworkManager.shared.checkForMaintenance(withCompletion: { response in
+                if let message = response {
```
bchrisarriola: Nit: use `guard` instead?

Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
```diff
@@ -27,8 +27,8 @@ extension AuthenticationCoordinator: PEPinEntryControllerDelegate {
 
         showVerifyingBusyView(withTimeout: 30)
 
-        let pinKey = BlockchainSettings.App.shared.pinKey
-        let pinString = pin.toString
+        // let pinKey = BlockchainSettings.App.shared.pinKey
+        // let pinString = pin.toString
```
bchrisarriola: Remove?

Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
```diff
@@ -37,12 +37,20 @@ extension AuthenticationCoordinator: PEPinEntryControllerDelegate {
...
-        //            [self checkForMaintenanceWithPinKey:pinKey pin:pin];
-        //        });
+        guard let pinKey = BlockchainSettings.App.shared.pinKey else {
+            return
+        }
 
+        // Check for maintenance before allowing pin entry
+        NetworkManager.shared.checkForMaintenance(withCompletion: { response in
+            LoadingViewPresenter.shared.hideBusyView()
+            guard let message = response else {
```
bchrisarriola: A bit hard to follow the flow in this block. How about flipping the `guard` to check that the response is nil, otherwise, display the error? I think `guard` is best used to check success cases and not the other way around.

So:

```
guard response == nil else {
    print("Error checking for maintenance in wallet options: %@", message)
    self.pinEntryViewController?.reset()
    AlertViewPresenter.shared.showMaintenanceAlert(withTitle: LocalizationConstants.Errors.error, message)
    return
}
self.walletManager.wallet.apiGetPINValue(pinKey, pin: pin.toString)
```

Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
```diff
@@ -37,12 +37,20 @@ extension AuthenticationCoordinator: PEPinEntryControllerDelegate {
...
-
-        //        dispatch_async(dispatch_get_main_queue(), ^{
-        //            [self checkForMaintenanceWithPinKey:pinKey pin:pin];
-        //        });
+        guard let pinKey = BlockchainSettings.App.shared.pinKey else {
+            return
+        }
 
+        // Check for maintenance before allowing pin entry
+        NetworkManager.shared.checkForMaintenance(withCompletion: { response in
```
bchrisarriola: Retain cycle here—add `[weak self]`

i.e.
```
    NetworkManager.shared.checkForMaintenance(withCompletion: { [weak self] response in
        guard let strongSelf = self else { return }
        // ...do stuff
    })
```

Blockchain/Coordinators/OnboardingCoordinator.swift
```diff
@@ -12,21 +12,24 @@ import Foundation
...
 
-    private init() {
-    }
+    private init() {}
 
     // MARK: Public Methods
 
     func start() {
-        showWelcomeScreen()
+        NetworkManager.shared.checkForMaintenance(withCompletion: { response in
```
bchrisarriola: Retain cycle—add `[weak self]`

Blockchain/Network/NetworkManager.swift
```diff
@@ -53,6 +53,36 @@ final class NetworkManager: NSObject, URLSessionDelegate {
...
     }
 
+    func checkForMaintenance(withCompletion handler: @escaping (_ response: String?) -> Void) {
+        guard
+            let walletOptionsUrl = BlockchainAPI.shared.walletOptionsUrl,
+            let url = URL(string: walletOptionsUrl) else {
+                fatalError("Failed to get wallet options url from Bundle.")
+        }
+        NetworkManager.shared.session.sessionDescription = url.host
+        let task = NetworkManager.shared.session.dataTask(with: url) { data, _, error in
```
bchrisarriola: `[weak self]`

Blockchain/Network/NetworkManager.swift
```diff
@@ -53,6 +53,36 @@ final class NetworkManager: NSObject, URLSessionDelegate {
...
+            if error != nil {
```
bchrisarriola: `guard` instead?

Blockchain/Wallet/WalletOptions.swift
```diff
@@ -0,0 +1,14 @@
...
+//  WalletOptions.swift
+//  Blockchain
+//
+//  Created by Maurice A. on 5/1/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+enum WalletOptions: String {
```
bchrisarriola: Should this actually be a `struct`?

Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
achtenhagen: 👍

Blockchain/Alerts/AlertViewPresenter.swift
achtenhagen: 👍

Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
achtenhagen: 👍

Blockchain/Wallet/WalletOptions.swift
achtenhagen: I made it an enum because it isn't yet clear how we are going to use this, but making it a struct is probably what will end up happening.

Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
achtenhagen: In the case of Authentication, I think `[unowned self]` would work. Agree?

Blockchain/Wallet/WalletOptions.swift
achtenhagen: Converted to `struct`.

Blockchain/Network/NetworkManager.swift
achtenhagen: 👍

Blockchain/Authentication/AuthenticationCoordinator.swift
achtenhagen: 👍

achtenhagen:
Thanks for your feedback! Please let me know what you think about my comments and merge when ready!

bchrisarriola:
LGTM. Good to merge into `swift`, waiting to merge #133 first.

PR #136, opened 2018-05-02
refactor(ManualPairing): move manual pairing away from RootService.

bchrisarriola:
Moved the manual pairing flow away from `RootService`. Still need to implement scanning QR and verifying all the 2FA flows so will get to that next.

Summary of changes:
* Removed some of the .xib files I created and instead using `MainWindow.xib` to instantiated them (affected views: `PasswordRequiredView`, `BCCreateWalletView` and `BCFadeView`). The reason for this is because localized strings are coupled with `MainWindow.xib` and is pretty hard to move them over at the moment. Will need to get to that on a later milestone.
* Created `AuthenticationTwoFactorType` model
* Updated `AuthenticationManager.Handler` to also accept a `AuthenticationTwoFactorType?`

Please review/merge previous PRs before getting to this.


Blockchain.xcodeproj/project.pbxproj
```diff
@@ -2544,7 +2545,7 @@
 						TestTargetID = 9FB3637014B60128004BEA02;
 					};
 					9FB3637014B60128004BEA02 = {
-						DevelopmentTeam = NZ6PH75U7K;
+						DevelopmentTeam = JX4379MUXB;
```
woudini: Should stay as `NZ6PH75U7K`. If you are having trouble running locally you should be given the dev role in iTunes Connect.

Also applies to `DEVELOPMENT_TEAM` changes below.

PR #138, opened 2018-05-03
refactor(pairing): move wallet pairing via QR code

bchrisarriola:
Tested that paring a wallet via QR code works. Will work on 2FA flow next.

### DIFF
--- a/Blockchain/Authentication/AuthenticationCoordinator.swift
+++ b/Blockchain/Authentication/AuthenticationCoordinator.swift
@@ -45,16 +45,17 @@ import Foundation
 
         ModalPresenter.shared.closeAllModals()
 
-        if BlockchainSettings.App.shared.isPinSet {
-            AppCoordinator.shared.showHdUpgradeViewIfNeeded()
-        }
-
-        // New wallet set-up. This will guide the user to create a pin & optionally
-        // enable touch ID and email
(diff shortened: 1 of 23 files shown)


Blockchain/Authentication/AuthenticationCoordinator.swift
```diff
@@ -280,6 +335,7 @@ import Foundation
             topMostViewController?.present(pinViewController, animated: true) { [weak self] in
                 guard let strongSelf = self else { return }
 
+                // CA: Can both of these alerts be moved elsewhere?
```
woudini: What does `CA:` mean?

bchrisarriola: "CA" are my initials. I did that so it's clear who wrote the comment (but apparently it's not).

@achtenhagen @woudini what do you guys think about having a convention like this so when you're reading through the code you know who to talk to if you have a question about a specific comment (another alternative is to just use our first name)? Using `git blame` is another option but this is easier

woudini: I'd rather not use it. In Xcode you can `control` + click and go to `Show Last Change for Line`, which shows the author.

bchrisarriola: Ah, good to know. Let me update

PR #139, opened 2018-05-04
fix(network): completion should be invoked on main thread.

bchrisarriola:
Fixes exception related to making UI related updates on a background thread.


bchrisarriola:
@achtenhagen going to merge this since this is blocking some of the work I'm doing now. If there are any comments, will address it in a separate PR.

PR #142, opened 2018-05-04
 refactor(auth): migrate second password flow

bchrisarriola:
Moved `getPrivateKeyPassword` and `getSecondPassword` away from `RootService` into `AuthenticationCoordinator`.


Blockchain/Authentication/AuthenticationCoordinator.swift
```diff
@@ -379,6 +399,44 @@ import Foundation
...
+            guard !passwordConfirmView.validateSecondPassword || self.walletManager.wallet.validateSecondPassword(password) else {
+                AlertViewPresenter.shared.standardNotify(message: LocalizationConstants.Authentication.secondPasswordIncorrect)
+                return
+            }
+
+            confirmHandler(password)
+            
+            ModalPresenter.shared.closeModal(withTransition: kCATransitionFade)
+        }
+        ModalPresenter.shared.showModal(
```
woudini: This needs to handle the scenario of when a `topMostViewController` is presented. Currently when going to Addresses and creating a new HD account the second password modal appears under the Addresses view controller. See https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/RootService.m#L1413-L1422

Generally `showModal` is used when a second password is requested outside of the side menu. Sorry I don't think that was documented anywhere

woudini: To test, you probably have to temporarily change this line to return `nil` to get to the addresses screen: https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/NSNumberFormatter+Currencies.m#L77

bchrisarriola: Ah, in the case, let me update `ModalPresenter` to always present the modal in `topMostViewController` (at the moment, it looks like it's always using `rootViewController`). 

`ModalPresenter` should be agnostic as to what view controller is presented, this way, callee's don't need to check which view controller is showing before presenting a modal.

woudini: Also had to make a change in `ModalPresenter`, please see https://github.com/blockchain/My-Wallet-V3-iOS/pull/143 which should be merged into `password`

bchrisarriola: K, verified that `ce5a356` will present the second password view. There is a bug though where calling into `wallet.validateSecondPassword` will always return false and this is because `wallet.isInitialized` returns false. Any idea why that may happen?

bchrisarriola:
Heads up: XCode also removed some unrelated lines in `Blockchain.xcodeproj/project.pbxproj`. I ran tests and ran the app which all work 💯. Not sure why those lines were there or why XCode decided to remove them. I actually manually added those lines back (which also cause the app + tests to work), but they get removed again.

PR #143, opened 2018-05-07
fix(ModalPresenter): post notification before checking for existing m…

woudini:
…odal view

The `BCModalViewController` is the way the app historically reused modal views by presenting them inside its own view controller. This was how modal views were shown for side menu items, which were presented by the tab view controller.

### DIFF
--- a/Blockchain/ModalPresenter.swift
+++ b/Blockchain/ModalPresenter.swift
@@ -69,12 +69,13 @@ typealias OnModalResumed = () -> Void
     }
 
     @objc func closeModal(withTransition transition: String) {
+
+        NotificationCenter.default.post(name: Constants.NotificationKeys.modalViewDismissed, object: nil)
+
         guard let modalView = modalView else {
             print("Cannot close modal. modalView is nil.")
             return
         }
-
-        NotificationCenter.default.post(name: Constants.NotificationKeys.modalViewDismissed, object: nil)
 
         modalView.removeFromSuperview()
 


Blockchain/ModalPresenter.swift
```diff
@@ -69,13 +69,14 @@ typealias OnModalResumed = () -> Void
     }
 
     @objc func closeModal(withTransition transition: String) {
+
+        NotificationCenter.default.post(name: Constants.NotificationKeys.modalViewDismissed, object: nil)
```
bchrisarriola: Shouldn't this be after the guard statement (i.e. when it has been confirmed that a `modalView` exists and was actually dismissed)?

woudini: In cases where the user is viewing a side menu item (e.g. Settings, Addresses), there isn't a `modalView` because it isn't set when presenting a modal view inside a view controller as in https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/AccountsAndAddressesViewController.m#L134. `modalView` is only set when outside a side menu item.

bchrisarriola: If I'm understanding this correctly, it looks like the method `closeModal` may not actually close its own reference of a modal and instead will post a notification that a _different_ modal view should be dismissed? It could also be the case that 2 modals are closed (i.e. if `modalView` is non-nill, and there is a presented `BCModalViewController`).

Some short-term solutions: 
* can this notification be removed from `ModalViewPresenter` and instead posted at the site that actually needs to close a presented `BCModalViewController`?
* can the call site that requires a `BCModalViewController` call `.dismiss` directly?
* instead of using `BCModalViewController`, can `ModalViewPresenter` be used?

I think the current design is very error-prone as the term "modal" is conflated and managed in different ways (in 1 case, it is managed by `ModalViewPresenter`, in the other, it is managed by a `BCModalViewController`). A longer-term solution will be more involved.

woudini: Yes, to be more specific it posts a notification for a view controller to be dismissed (`BCModalViewController`) - in these cases no modal view is involved.
- The issue is mainly that the close button is part of the `ModalView` and somehow needs to communicate its action with the view controller. The button's default action is to close the modal.
- Same issue as above
- `ModalViewPresenter` could check if there is a top-most view controller or a top view controller and present the modal inside of a `BCModalViewController` however this would be a nontrivial change.

bchrisarriola: > The issue is mainly that the close button is part of the ModalView and somehow needs to communicate its action with the view controller. The button's default action is to close the modal.

Ah, if this is the case, can we do the following:
(1) create a `BCModalViewDelegate`
(2) have the delegate define a method like `didCloseModal`
(3) have `BCmodalViewController` be a delegate of `BCModalView`

With this design, can we get rid of the notification altogether?

If you think this can work, let's merge this and address that in a separate pull request.

woudini: I think that can work. Would also be able to fix this hack I just remembered - the private key prompt's `Continue` button is a pseudo-button that is added on top of the actual `Continue` button, so a delegate would also help clean up that implementation: https://github.com/blockchain/My-Wallet-V3-iOS/blame/UIWebView/Blockchain/AppDelegate.m#L911-L913

PR #147, opened 2018-05-08
feat(Alert View Presenter): support custom alert actions

achtenhagen:
This PR adds support for custom alert actions for the `standardNotify` functions. It also fixes an issue that prevented alerts from showing while the application was transitioning to the active state.

### DIFF
--- a/Blockchain/Alerts/AlertViewPresenter.swift
+++ b/Blockchain/Alerts/AlertViewPresenter.swift
@@ -94,9 +94,8 @@ import Foundation
     }
 
     @objc func standardNotify(message: String, title: String = LocalizationConstants.Errors.error, handler: AlertConfirmHandler? = nil) {
-        let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
-        alert.addAction(UIAlertAction(title: LocalizationConstants.okString, style: .cancel, handler: handler))
-        standardNotify(alert: alert)
+        let standardAction = UIAlertAction(title: LocalizationConstants.okString, style: .cancel, handler: handler)
+        standardNotify(message: message, title: title, actions: [standardAction])
     }
 
     /// Allows custom actions to be included in the standard alert presentation
--- a/Blockchain/Alerts/AlertViewPresenter.swift
+++ b/Blockchain/Alerts/AlertViewPresenter.swift
@@ -93,17 +93,24 @@ import Foundation
         }
     }
 
-    @objc func standardNotify(
-        message: String,
-        title: String = LocalizationConstants.Errors.error,
-        handler: AlertConfirmHandler? = nil
-    ) {
+    @objc func standardNotify(message: String, title: String = LocalizationConstants.Errors.error, handler: AlertConfirmHandler? = nil) {
+        let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: LocalizationConstants.okString, style: .cancel, handler: handler))
+        standardNotify(alert: alert)
+    }
+
+    /// Allows custom actions to be included in the standard alert presentation
+    @objc func standardNotify(message: String, title: String, actions: [UIAlertAction]) {
+        let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
+        actions.forEach { alert.addAction($0) }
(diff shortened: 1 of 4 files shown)


Blockchain/Alerts/AlertViewPresenter.swift
```diff
@@ -93,17 +93,24 @@ import Foundation
...
+    @objc func standardNotify(message: String, title: String = LocalizationConstants.Errors.error, handler: AlertConfirmHandler? = nil) {
+        let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: LocalizationConstants.okString, style: .cancel, handler: handler))
+        standardNotify(alert: alert)
+    }
 
-            let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
-            alert.addAction(UIAlertAction(title: LocalizationConstants.okString, style: .cancel, handler: handler))
+    /// Allows custom actions to be included in the standard alert presentation
+    @objc func standardNotify(message: String, title: String, actions: [UIAlertAction]) {
```
woudini: This is pretty confusing since the first `standardNotify`'s title is always "Error" but this `standardNotify` accepts a custom one. I'd change the naming of one or the other

Blockchain/Alerts/AlertViewPresenter.swift
```diff
@@ -93,17 +93,24 @@ import Foundation
...
     }
 
-    @objc func standardNotify(
-        message: String,
-        title: String = LocalizationConstants.Errors.error,
-        handler: AlertConfirmHandler? = nil
-    ) {
-        DispatchQueue.main.async {
-            guard UIApplication.shared.applicationState == .active else { return }
+    @objc func standardNotify(message: String, title: String = LocalizationConstants.Errors.error, handler: AlertConfirmHandler? = nil) {
```
woudini: Can you indent these like they were before, I think it's easier to read

Blockchain/Alerts/AlertViewPresenter.swift
```diff
@@ -93,17 +93,24 @@ import Foundation
...
+        let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: LocalizationConstants.okString, style: .cancel, handler: handler))
+        standardNotify(alert: alert)
```
bchrisarriola: I think this method can just call `standardNotify(message: String, title: String, actions: [UIAlertAction])`

So: 
```
standardNotify(message: message, title: title, actions: [UIAlertAction(title: LocalizationConstants.okString, style: .cancel, handler: handler)])
```

Blockchain/Alerts/AlertViewPresenter.swift
achtenhagen: It's not always "Error", it's just the default parameter. I think we should remove the default param and instead create an error alert function. Thoughts? But I agree, the signatures should match up until the last argument.

achtenhagen: Because I do not think we should assume the general context of a standard alert to be failure, unless @bchrisarriola found that to be the case when writing this function.

woudini: Then why not remove the default "Error" `title` parameter for both

bchrisarriola: I think either approach can work. Another improvement would be to have `standardNotify` accept a `Severity` enum (i.e. info, error, warning, etc.) and display a title based on that.

Agreed with @achtenhagen though that the method should probably not assume the alert is an error. Let's go with what you suggested:

> I think we should remove the default param and instead create an error alert function. 

If this is blocking you with another change, it's ok to merge this and just do another PR with the `AlertViewPresenter` change.

achtenhagen: I like the severity enum idea and we should implement that in a subsequent pr. For now I will remove the default param and add a lightweight `standardError` function. After that we should be good to go.

achtenhagen: Now we will have something like:
`AlertViewPresenter.shared.standardError(message: errorMessage)`

PR #148, opened 2018-05-09
refactor(ShowType): Migrating show type to PostAuthenticationRoute.

bchrisarriola:
* Moved `showType` away from `RootService` and renamed this enum to `PostAuthenticationRoute`.
* Added back `userNotificationCenter:didReceive:withCompletionHandler` back to `PushNotificationManager` since it was accidentally removed on b79a4b35 (@woudini can you confirm that `ShowTypeNewPayment` was actually used and not part of contacts in which case we should remove it?)
* Commented out some more unused code in `RootService`

### DIFF
--- a/Blockchain/Notifications/PushNotificationManager.swift
+++ b/Blockchain/Notifications/PushNotificationManager.swift
@@ -35,13 +35,9 @@ class PushNotificationManager: NSObject {
         return shared
     }
 
-    /// UNNotification that is to be presented to the user
-    @objc var presentingPushNotification: UNNotification?
-
     /// Requests permission from the user to grant access to receive push notifications
     func requestAuthorization() {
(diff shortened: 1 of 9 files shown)


woudini:
`ShowTypeNewPayment` was part of contacts. It was added to setup a transaction in the send screen ready to be sent when you opened the app through a push notification that requested a payment.

Blockchain/Notifications/PushNotificationManager.swift
```diff
@@ -59,6 +59,23 @@ class PushNotificationManager: NSObject {
 @available(iOS 10.0, *)
 extension PushNotificationManager: UNUserNotificationCenterDelegate {
```
woudini: can remove this extension, or if it's required just leave the methods empty

PR #150, opened 2018-05-09
refactor(BuySell): setup trade from dictionary in wallet delegate method

woudini:
To address comment in https://github.com/blockchain/My-Wallet-V3-iOS/pull/149

### DIFF
--- a/Blockchain/Trade.swift
+++ b/Blockchain/Trade.swift
@@ -12,7 +12,7 @@ import Foundation
 // TODO: integrate with Exchange trades.
 struct Trade {
 
-    private struct Keys {
+    struct Keys {
         static let created = "createdAt"
         static let receiveAddress = "receiveAddress"
         static let tradeHash = "txHash"
--- a/Blockchain/Trade.swift
+++ b/Blockchain/Trade.swift
@@ -12,7 +12,7 @@ import Foundation
 // TODO: integrate with Exchange trades.
 struct Trade {
 
-    struct Keys {
+    private struct Keys {
         static let created = "createdAt"
         static let receiveAddress = "receiveAddress"
         static let tradeHash = "txHash"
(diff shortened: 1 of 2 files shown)


Blockchain/Trade.swift
```diff
@@ -12,7 +12,7 @@ import Foundation
 // TODO: integrate with Exchange trades.
 struct Trade {
 
-    private struct Keys {
+    struct Keys {
```
bchrisarriola: Better to create a failable initializer that accepts a dictionary so that this can be private (and so WalletManager doesn't need to know the details of how to construct a `Trade` object given a dictionary.

So:

```
struct Trade {
    private struct Keys {
         static let created = "createdAt"
         static let receiveAddress = "receiveAddress"
         static let tradeHash = "txHash"
    }
}

extension Trade {
    init?(dict: [String: String]) {
         guard let hash = trade[Keys.tradeHash] else {
             print("trade hash not found")
             return nil
         }
         guard let date = trade[Keys.created] else {
             print("trade date not found")
             return nil
         }
         return Trade(date: date, hash: hash)
     }
}
```

In WalletManager:

```
func didCompleteTrade(_ tradeDict: [String: String]) {
    guard let trade = Trade(dict: tradeDict) else {
        print("Failed to create Trade object.")
        return    
    }
    buySellDelegate?.didCompleteTrade(trade: trade)
}
```

bchrisarriola: Also noticed that there is a `createdAt` key, should `Trade` also have an optional `createdAt` property?

woudini: For now I think we should leave `createdAt` as is instead of making it optional. Without a date the messaging would show "The trade you created on has been completed" - that then becomes a product question for a case I don't think will happen often.

PR #151, opened 2018-05-09
refactor(AppFeature): Removing #defines for app feature enabling.

bchrisarriola:
Removing `#defines`  for enabling/disabling app features such as touch ID, swipe to receive, etc. If a component requires a configuration for a specific feature, it will request it by invoking `AppFeatureConfigurator.shared.configuration(for: feature)`. The benefit of using a design like this is that we can eventually add more specific feature configurations into an instance of `AppFeatureConfiguration`. At the moment, all it does is tell you if a feature is enabled or not (by default, all features are enabled).

Other changes:
* Renamed extension class to `AuthenticationCoordinator+Biometrics`
* Implemented TODOs that depended on the `#define ENABLE_TOUCH_ID`
* Removed #defines

### DIFF
--- a/Blockchain/SettingsTableViewController.m
+++ b/Blockchain/SettingsTableViewController.m
@@ -37,16 +37,6 @@ const int securityTwoStep = 0;
 const int securityPasswordChange = 1;
 const int securityWalletRecoveryPhrase = 2;
 const int PINChangePIN = 3;
-#if defined(ENABLE_TOUCH_ID) && defined(ENABLE_SWIPE_TO_RECEIVE)
-const int PINTouchID = 4;
-const int PINSwipeToReceive = 5;
-#elif ENABLE_TOUCH_ID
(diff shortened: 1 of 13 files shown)


Blockchain/SettingsTableViewController.m
```diff
@@ -70,10 +60,37 @@ @interface SettingsTableViewController () <UITextFieldDelegate, EmailDelegate, M
...
 @property (nonatomic) BackupNavigationViewController *backupController;
 
+@property (readonly, nonatomic) int PINTouchID;
+@property (readonly, nonatomic) int PINSwipeToReceive;
+
 @end
 
 @implementation SettingsTableViewController
 
+- (int)ujPINTouchID
```
woudini: typo?

bchrisarriola: fixed

PR #152, opened 2018-05-10
refactor(Settings): set App Coordinator as settings delegate

woudini:
Most of the Settings-related methods use a notification to communicate with the `SettingsViewController`.

Not sure where the best place is to set the `WalletSettingsDelegate` inside `AppCoordinator`. I thought about moving it into `showSettingsView` but it wouldn't make sense to set a delegate each time settings is opened.

### DIFF
--- a/Blockchain/TabControllerManager.h
+++ b/Blockchain/TabControllerManager.h
@@ -23,7 +23,9 @@
 - (void)toggleSideMenu;
 @end
 
-@interface TabControllerManager : NSObject <AssetDelegate>
+@protocol WalletSettingsDelegate;
+
+@interface TabControllerManager : NSObject <AssetDelegate, WalletSettingsDelegate>
 @property (nonatomic) AssetType assetType;
 @property (nonatomic) NSDecimalNumber *latestEthExchangeRate;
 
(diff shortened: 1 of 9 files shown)


bchrisarriola:
@woudini just realized this: what about having `TabControllerManager` conform to `WalletSettingsDelegate` and set itself as the wallet's `settingsDelegate` (can be in the constructor of TabControllerManager)?

woudini:
Sure, changed to set `TabControllerManager` as `settingsDelegate`.

Blockchain/TabControllerManager.h
```diff
@@ -23,7 +23,9 @@
 - (void)toggleSideMenu;
 @end
 
-@interface TabControllerManager : NSObject <AssetDelegate>
+@protocol WalletSettingsDelegate;
+
+@interface TabControllerManager : NSObject <AssetDelegate, WalletSettingsDelegate>
```
woudini: I'm getting a warning `Cannot find protocol definition for 'WalletSettingsDelegate'` here

woudini: Removed forward declarations in https://github.com/blockchain/My-Wallet-V3-iOS/pull/152/commits/1e5b978e822b0caa1a99e97cc0bdbc1f738528d8.

PR #154, opened 2018-05-11
refactor(reminder): move all reminder code to ReminderCoordinator.

bchrisarriola:
Note that I did not branch off of the `touch_id` branch (#153) so this can be reviewed/merged in parallel with it.


bchrisarriola:
@woudini when I tried this, I was able to get that part of the code executing (didn't show a reminder on my account though since I had 2FA and email verified). Will test again and try different scenarios

bchrisarriola:
Looks like there were some authentication flows (pin entry + touch ID) wherein `authHandler` within `WalletManager` was not set. Good catch, @woudini 

Some adjustments I made:
* Created a method in `AuthenticationManager` to allow authenticating with a pin (check for maintenance is wrapped within here), this method accepts an `WalletAuthHandler`
* Remove an extra call to `authenticateWithBiometrics` in AppDelegate's `applicationWillEnterForeground` method
* `AuthenticationManager.Handler` typealias was previous conflated to mean two things: (1) handler for biometrics authentication, and (2) handler for password or pin authentication. To avoid the confusion, I created 2 type aliases (i.e. `WalletAuthHandler` and `BiometricsAuthHandler`) to capture these two different use cases

woudini:
I'm now seeing multiple PIN screens being added after each time the app is backgrounded (so once one is dismissed, the one underneath it shows). Changing the `isPinEntryModalPresented` var to something like this worked for me:
```
private var isPinEntryModalPresented: Bool {
        let rootViewController = UIApplication.shared.keyWindow!.rootViewController!
        let tabControllerManager = AppCoordinator.shared.tabControllerManager
        guard let pinEntryViewController = pinEntryViewController else {
            return false
        }
        return (tabControllerManager.tabViewController.presentedViewController == pinEntryViewController &&
            !pinEntryViewController.isBeingDismissed) || pinEntryViewController.view.isDescendant(of: rootViewController.view)
    }
```
I'm also still not seeing the reminders, but they can't really be tested until `getAccountInfo` is properly in place.

bchrisarriola:
Fixed the double pin entry view bug from your suggestion.

> I'm also still not seeing the reminders, but they can't really be tested until getAccountInfo is properly in place.

Confirmed that the reminder logic is actually being executed—I plan on moving the account info-related code in a subsequent PR.

bchrisarriola:
@woudini can this be merged?

PR #159, opened 2018-05-17
refactor(RootService): clean up API calls on startup

woudini:
Clean up the flow of methods on startup, which was previously the following:
```
Multiaddr
did_multiaddr
didGetMultiAddressResponse
getAccountInfo
reloadAfterGettingCurrencySymbols -> updateSymbols (BTC rates)
fetchBitcoinCashExchangeRates
getEthExchangeRate
```

### DIFF
--- a/Blockchain/Wallet.m
+++ b/Blockchain/Wallet.m
@@ -694,13 +694,17 @@
     };
 
 #pragma mark Settings
-
+    
+    self.context[@"objc_on_get_account_info_and_exchange_rates"] = ^() {
+        [weakSelf on_get_account_info_and_exchange_rates];
+    };
+    
     self.context[@"objc_on_get_account_info_success"] = ^(NSString *accountInfo) {
         [weakSelf on_get_account_info_success:accountInfo];
     };
-
-    self.context[@"objc_on_get_all_currency_symbols_success"] = ^(NSString *currencies) {
-        [weakSelf on_get_all_currency_symbols_success:currencies];
+    
+    self.context[@"objc_on_get_btc_exchange_rates_success"] = ^(NSString *currencies) {
+        [weakSelf on_get_btc_exchange_rates_success:currencies];
     };
 
     self.context[@"objc_on_error_creating_new_address"] = ^(NSString *error) {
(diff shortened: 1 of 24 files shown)


Blockchain/Wallet.m
```diff
@@ -4279,6 +4288,13 @@ - (void)on_build_exchange_trade_success_from:(NSString *)from depositAmount:(NSS
     }
 }
 
+- (void)on_get_account_info_and_exchange_rates
+{
+    [LoadingViewPresenter.sharedInstance hideBusyView];
```
bchrisarriola: These actions are UI-specific and the wallet-layer should not interact with that. Instead, we should follow the [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) here and invoke a delegate method to propagate t

So:
1. Define another method in `WalletDelegate` (i.e. `- (void)walletDidGetAccountInfoAndExchangeRates:(Wallet *)wallet`)
2. Have WalletManager implement `walletDidGetAccountInfoAndExchangeRates()`
3. Create a protocol called `WalletExchangeRatesDelegate`
4. AppCoordinator conforms to `WalletExchangeRatesDelegate`

Blockchain/Wallet/WalletManager.swift
```diff
@@ -233,13 +259,46 @@ extension WalletManager: WalletDelegate {
...
     // MARK: - Currency Symbols
 
     func walletDidGetAllCurrencySymbols(_ wallet: Wallet!) {
         updateSymbols()
-        wallet.fetchBitcoinCashExchangeRates()
+    }
+
+    // MARK: - BTC Multiaddress
+    func didSetLatestBlock(_ block: LatestBlock!) {
+        AppCoordinator.sharedInstance().tabControllerManager.didSetLatestBlock(block)
```
bchrisarriola: This can be `AppCoordinator.shared`. 

`.sharedInstance()` is mostly for obj-c compat

Blockchain/Wallet/WalletManager.swift
```diff
@@ -233,13 +259,46 @@ extension WalletManager: WalletDelegate {
...
+        let newDefaultAccountLabeledAddressesCount = self.wallet.getDefaultAccountLabelledAddressesCount()
+        if let lastCount = BlockchainSettings.App.shared.defaultAccountLabelledAddressesCount?.intValue,
+            lastCount != newDefaultAccountLabeledAddressesCount {
+            KeychainItemWrapper.removeAllSwipeAddresses(for: .bitcoin)
+        }
+        let newCount = NSNumber(integerLiteral: Int(newDefaultAccountLabeledAddressesCount))
+        BlockchainSettings.App.shared.defaultAccountLabelledAddressesCount = newCount
+    }
+
+    func reloadAfterMultiaddressResponse() {
```
bchrisarriola: private

PR #162, opened 2018-05-19
refactor(RootService): create recovery delegate

woudini:


Blockchain/Coordinators/OnboardingCoordinator.swift
```diff
@@ -12,6 +12,8 @@ import Foundation
 class OnboardingCoordinator: Coordinator {
     static let shared = OnboardingCoordinator()
 
+    var createWallet: BCCreateWalletView?
```
bchrisarriola: Make this `private`

PR #167, opened 2018-05-21
refactor(transaction): moving transaction-related delegate methods.

bchrisarriola:


bchrisarriola:
Lmk if I missed any other transaction-related methods.

If so, I'll address it in a separate PR.

PR #168, opened 2018-05-21
refactor(RootService): set app coordinator as transfer all delegate

woudini:
Some other changes:
- Re-implemented `reloadAfterMultiAddressResponse` inside `AppCoordinator`
- If `symbol_local` is nil, changed to return nil for local currency conversion. This seems to be happening due to `backup_wallet_success` calling `reload` before `getAccountInfoAndExchangeRates` has a chance to set `symbol_local` after `latestMultiAddressResponse` has been reset. If possible we should store `symbol_local` separately from `latestMultiAddressResponse`.
- Created `DestinationAddressSource.h` and added to `SendEtherViewController`. Tried making a swift enum multiple times (added to both main and test target) and the compiler couldn't find `Blockchain-Swift.h`

### DIFF
--- a/Blockchain/AccountsAndAddressesNavigationController.m
+++ b/Blockchain/AccountsAndAddressesNavigationController.m
@@ -221,7 +221,7 @@
         [[AppCoordinator sharedInstance] closeSideMenu];
     }];
     
-    [app setupTransferAllFunds];
+    [[AppCoordinator sharedInstance] setupTransferAllFunds];
 }
 
 #pragma mark - Navigation
--- a/Blockchain/AccountsAndAddressesNavigationController.m
+++ b/Blockchain/AccountsAndAddressesNavigationController.m
@@ -221,7 +221,7 @@
         [[AppCoordinator sharedInstance] closeSideMenu];
     }];
     
-    [[AppCoordinator sharedInstance] setupTransferAllFunds];
+    [[TransferAllCoordinator sharedInstance] startWithSendScreen];
 }
 
 #pragma mark - Navigation
--- a/Blockchain/BackupViewController.swift
+++ b/Blockchain/BackupViewController.swift
@@ -65,8 +65,7 @@ import UIKit
                 let alertToTransferAll = UIAlertController(title: NSLocalizedString("Transfer imported addresses?", comment: ""), message: NSLocalizedString("Imported addresses are not backed up by your Recovery Phrase. To secure these funds, we recommend transferring these balances to include in your backup.", comment: ""), preferredStyle: .alert)
                 alertToTransferAll.addAction(UIAlertAction(title: NSLocalizedString("Cancel", comment: ""), style: .cancel, handler: nil))
                 alertToTransferAll.addAction(UIAlertAction(title: NSLocalizedString("Transfer all", comment: ""), style: .default, handler: { _ in
-                    TransferAllCoordinator.shared.start()
-                    TransferAllCoordinator.shared.transferAllController?.delegate = self
+                    TransferAllCoordinator.shared.start(withDelegate: self)
                 }))
                 present(alertToTransferAll, animated: true, completion: nil)
             }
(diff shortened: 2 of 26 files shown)


bchrisarriola:
Can `TabControllerManager` conform to `WalletTransferAllDelegate`? Looks like most of these methods just call into `TabControllerManager`.

woudini:
It can but at the time of writing it I got the impression that `TabControllerManager` was getting a bit big. Is there a stricter definition of what should go in `TabControllerManager` vs `AppCoordniator`? Seems like it should go in its own `TransferAllCoordinator`.

Blockchain/AccountsAndAddressesNavigationController.m
```diff
@@ -221,7 +221,7 @@ - (void)transferAllFundsClicked
         [[AppCoordinator sharedInstance] closeSideMenu];
     }];
     
-    [app setupTransferAllFunds];
+    [[AppCoordinator sharedInstance] setupTransferAllFunds];
```
bchrisarriola: I think we should move this to `TransferAllCoordinator`. So:

```
[[TransferAllCoordinator sharedInstance] setup];
```

Blockchain/BackupViewController.swift
```diff
@@ -65,15 +65,8 @@ import UIKit
...
-                    transferAllController.delegate = self
-                    let navigationController = BCNavigationController(
-                        rootViewController: transferAllController,
-                        title: NSLocalizedString("Transfer All Funds",
-                        comment: "")
-                    )
-                    app.transferAllFundsModalController = transferAllController
-                    self.present(navigationController!, animated: true, completion: nil)
+                    TransferAllCoordinator.shared.start()
+                    TransferAllCoordinator.shared.transferAllController?.delegate = self
```
bchrisarriola: Let's hide `transferAllController` to external components by making it `private`. We can then pass the `TransferAllPromptDelegate` to the `start()` method.

So it would be something like:

```
TransferAllCoordinator.shared.start(withDelegate: self)
```

...and update `start` to:

```
func start(withDelegate delegate: TransferAllPromptDelegate)
```

Blockchain/Coordinators/AppCoordinator.swift
```diff
@@ -202,6 +202,12 @@ import Foundation
         slidingViewController.resetTopView(animated: true)
     }
 
+    @objc func setupTransferAllFunds() {
```
bchrisarriola: Can remove this (see comment above)

Blockchain/Coordinators/TransferAllCoordinator.swift
```diff
@@ -0,0 +1,63 @@
...
+
+/// Coordinator for the transfer all flow.
+class TransferAllCoordinator: Coordinator {
+    static let shared = TransferAllCoordinator()
+
+    private init() {
+        WalletManager.shared.transferAllDelegate = self
+    }
+
+    weak var transferDelegate: TransferAllPromptDelegate?
```
bchrisarriola: Remove?

Blockchain/Coordinators/TransferAllCoordinator.swift
```diff
@@ -0,0 +1,63 @@
...
+    var transferAllController: TransferAllFundsViewController?
```
bchrisarriola: `private`

woudini: `transferAllController` needs to be set to nil here https://github.com/blockchain/My-Wallet-V3-iOS/pull/168/files#diff-68785bd643dd937c322e1223b73018c9R207 otherwise it may branch incorrectly through the `if` statements inside the `WalletTransferAllDelegate` implementation.

Blockchain/BackupViewController.swift
woudini: How should https://github.com/blockchain/My-Wallet-V3-iOS/pull/168#discussion_r189754135 be addressed? Maybe a public method that sets the `transferAllController` to nil?

bchrisarriola:
🔥

PR #170, opened 2018-05-21
refactor(TabControllerManager): move exchange delegate methods to exc…

woudini:
…hange overview controller

It doesn't seem to make sense to move `didCreateEthAccountForExchange` to `ExchangeOverviewController` because it won't exist at that point. `showGetAssetsAlert` requires a lot of navigation so it would also be a stretch to move to `ExchangeOverviewController`. Should these be in a separate protocol?


bchrisarriola:
What action triggers (and where do they get triggered) the methods `didCreateEthAccountForExchange` and `showGetAssetsAlert`?

woudini:
`didCreateEthAccountForExchange` is called when an eth account needs to be created when exchange is first used (when a wallet is created on iOS, an eth account is not automatically created).

`showGetAssetsAlert` appears when you have a zero balance for all assets and prompts the user to buy BTC or receive BTC/ETH/BCH.

woudini:
For `didCreateEthAccountForExchange`, eth account creation is done in the JS so it is called by `createEthAccountForExchange:`

`showGetAssetsAlert` isn't actually a `WalletDelegate` method, so I will remove it from the protocol.

bchrisarriola:
Going to rebase this against `swift` so that I can merge it

PR #171, opened 2018-05-21
refactor(RootService): set app coordinator as wallet history delegate

woudini:


Blockchain/Coordinators/AppCoordinator.swift
```diff
@@ -404,8 +405,20 @@ extension AppCoordinator: WalletBackupDelegate {
...
+
     func didFailBackupWallet() {
         walletManager.wallet.getAndHistory()
     }
 }
+
+extension AppCoordinator: WalletHistoryDelegate {
+    func didFailGetHistory(error: String?) {
+        guard let errorMessage = error, errorMessage.count > 0 else {
+            AlertViewPresenter.shared.standardError(message: LocalizationConstants.Errors.noInternetConnectionPleaseCheckNetwork,
```
bchrisarriola: The default title is `LocalizationConstants.Errors.error` and handler is `nil` so you can just do:

```
AlertViewPresenter.shared.standardError(message: LocalizationConstants.Errors.noInternetConnectionPleaseCheckNetwork)
```

Same for the method below.

PR #172, opened 2018-05-21
fix(WalletManager): prevent crash by avoiding reload until startup co…

woudini:
…mplete

### DIFF
--- a/Blockchain/Wallet/WalletManager.swift
+++ b/Blockchain/Wallet/WalletManager.swift
@@ -350,7 +350,7 @@ extension WalletManager: WalletDelegate {
 
     // MARK: ETH Exchange Rate
     func didFetchEthExchangeRate(_ rate: NSNumber!) {
-        reloadAfterMultiaddressResponse()
+        if latestMultiAddressResponse?.symbol_local != nil { reloadAfterMultiaddressResponse() }
         AppCoordinator.shared.tabControllerManager.didFetchEthExchangeRate(rate)
     }
 


Blockchain/Wallet/WalletManager.swift
```diff
@@ -350,7 +350,7 @@ extension WalletManager: WalletDelegate {
 
     // MARK: ETH Exchange Rate
     func didFetchEthExchangeRate(_ rate: NSNumber!) {
-        reloadAfterMultiaddressResponse()
+        if latestMultiAddressResponse?.symbol_local != nil { reloadAfterMultiaddressResponse() }
```
bchrisarriola: Can we just remove this method altogether? If I understand correctly, this method will get invoked upon logging in after retrieving the ether exchange rate. But in 93652d14, a reload will be invoked again anyway when all the exchange rates have been retrieved.

bchrisarriola: I think there are several issues here:

* it's unclear why `reloadAfterMultiaddressResponse` needs to be invoked after `didFetchEthExchangeRate`. The method `reloadAfterMultiaddressResponse` implies that it's invoked after receiving a a multiaddress response.
* there are plenty of places throughout the code where we invoke some sort of reload method as a result of a state change. as part of the wallet re-arch, I suggest we use some sort of reactive solution (like RxSwift) so that the UI-layer can react to state changes as needed and we no longer need to reload every single view controller such as in this case

woudini: This is actually removed in https://github.com/blockchain/My-Wallet-V3-iOS/pull/168/files#diff-621d3cb799b89cef85d798a067985656L355 so you can close this PR at your discretion. I'd say we should prioritize deprecating this as part of the next milestone.

bchrisarriola: Ok let's close PR since #168 removes this call anyway.

Agree that we should prioritize this issue in the next milestone (I think the wallet rearch should address this).

PR #176, opened 2018-05-22
fix(warning): fixing warnings in WalletManager.

bchrisarriola:
* Not necessary to do conditional casting for `NSMutableDictionary` since it will always succeed
* Updated signature to `didCompleteTrade(_ tradeDict: [AnyHashable : Any]!)` so that it matches the signature defined in `WalletDelegate`

### DIFF
--- a/Blockchain/Wallet/WalletManager.swift
+++ b/Blockchain/Wallet/WalletManager.swift
@@ -133,10 +133,7 @@ class WalletManager: NSObject {
             print("Currency symbols dictionary is nil")
             return
         }
-        guard let symbolLocalDict = NSMutableDictionary(dictionary: currencySymbols) as? NSMutableDictionary else {
-            print("Could not create symbolLocalDict")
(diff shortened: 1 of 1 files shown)


Blockchain/Wallet/WalletManager.swift
```diff
@@ -220,8 +217,8 @@ extension WalletManager: WalletDelegate {
         buySellDelegate?.initializeWebView()
     }
 
-    func didCompleteTrade(_ tradeDict: [String: String]) {
-        guard let trade = Trade(dict: tradeDict) else {
+    func didCompleteTrade(_ tradeDict: [AnyHashable : Any]!) {
+        guard let trade = Trade(dict: tradeDict as! [String : String]) else {
```
woudini: Swiftlint warns "Colons should be next to the identifier when specifying a type and next to the key in dictionary literals" for this line and above

PR #177, opened 2018-05-22
Migrate Private Key Reader

achtenhagen:
This is a WIP PR for ~adding a legacy Private Key Reader accessible from Objective-C~. The Key Reader itself has also been moved into its own file instead of living as an extension to the wallet. Bitcoin addresses also validate via the JS bridge, which should be refactored to do native validation.


Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -0,0 +1,229 @@
...
+
+@objc
+final class PrivateKeyReader: UIViewController & AVCaptureMetadataOutputObjectsDelegate {
+
+    // MARK: Properties
+
+    private let assetType: AssetType?
+
+    //: Temporary bridging asset type
+    @objc private let legacyAssetType: LegacyAssetType
```
bchrisarriola: This can be removed

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -0,0 +1,229 @@
...
+        self.modalTransitionStyle = .crossDissolve
+        guard let deviceInput = getDeviceInput() else { return nil }
+        captureSession = AVCaptureSession()
+        captureSession?.addInput(deviceInput)
+    }
+
+    // TODO: remove once AccountsAndAddresses and SendBitcoinViewController are migrated to Swift
+    @objc init?(assetType: LegacyAssetType, acceptPublicKeys: Bool = false) {
+        legacyAssetType = assetType
+        self.assetType = nil; delegate = nil; captureSession = nil; videoPreviewLayer = nil
```
bchrisarriola:
```
self.assetType = AssetType.from(legacyAssetType:  assetType)
```

Can just remove `legacyAssetType` property, too.

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -0,0 +1,229 @@
...
+
+    weak var delegate: PrivateKeyReaderDelegate?
+
+    //: Legacy Objc delegate to support Legacy asset types
+    @objc weak var legacyDelegate: LegacyPrivateKeyDelegate?
+
+    //: Internal so that the same class instance can be reused to scan public keys
+    var acceptPublicKeys: Bool!
+
+    @objc var loadingText: String!
```
bchrisarriola: Can make this `private` and have it provided optionally in the constructor

bchrisarriola:
Some other comments:

* `AccontsAndAddressesDetailViewController` is still calling `[app scanPrivateKeyForWatchOnlyAddress]`. Could be good to create a coordinator object for the private/public key scanning

achtenhagen:
Great feedback @bchrisarriola, thanks! At which point would you call `KeyImportCoordinator.shared.start()` ? I currently have it in the initializer of the WalletManager.

achtenhagen:
Could you elaborate on why we should implement the methods of each delegate in the WalletManager just to call out to the delegates? Is the plan to have everything go through the WalletManager which then delegates the calls to their respective coordinators?

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
achtenhagen: `acceptPublicKeys` or `loadingText` ?

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
achtenhagen: It looks like `AssetType.from`... is not available to me yet. I'll have to rebase this branch first.

bchrisarriola:
> Could you elaborate on why we should implement the methods of each delegate in the WalletManager just to call out to the delegates?

The reason for this is because `WalletManager` should not know the details of what actually happens when a `WalletDelegate` method is invoked. For example, the `WalletDelegate` method `getPinSuccess` does a bunch of actions upon invocation (i.e. it invokes the `pinViewControllerCallback` with a success, it optionally displays an alert, it invokes `load` on the wallet, etc.). These actions should be abstracted from the `WalletManager` and instead be the responsibility of a higher-level abstraction (in that case, the `AuthenticationCoordinator`). Propagating up those actions is tedious but it allows delegation to the appropriate higher-level component.

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
bchrisarriola: Oh, both can be `private`.

bchrisarriola:
> Okay that makes sense to me. How do we intend to control the file size of the WalletManager? Wouldn't it eventually to be the same size as RootService?

Can't avoid that right now because `Wallet.h/.m` defines so many methods. However, it won't ever get as big as `RootService` because all it does it delegate up. 

Best solution is to rearchitect the wallet 😬

bchrisarriola:
@achtenhagen is this ready to be reviewed?

achtenhagen:
@bchrisarriola Yes it's ready.

achtenhagen:
Although #184 should be reviewed first.

Blockchain/Coordinators/KeyImportCoordinator.swift
```diff
@@ -0,0 +1,215 @@
...
+    private let walletManager: WalletManager
+
+    // TODO: Refactor class to support other asset types (currently assumed to be Bitcoin)
+    private init(walletManager: WalletManager = WalletManager.shared) {
+        self.walletManager = walletManager
+        super.init()
+    }
+
+    func start() {
+        walletManager.keyImportDelegate = self
```
bchrisarriola: Move inside `init()`

bchrisarriola: I think `start()` should mean it should launch `PrivateKeyReader`. Is that what you had intended?

Blockchain/Coordinators/KeyImportCoordinator.swift
```diff
@@ -0,0 +1,215 @@
...
+
+        // TODO: `lastScannedWatchOnlyAddress` needs to be of type AssetAddress, not String
+        walletManager.wallet.lastScannedWatchOnlyAddress = address.description
+    }
+}
+
+// MARK: - PrivateKeyReaderDelegate
+
+extension KeyImportCoordinator: PrivateKeyReaderDelegate {
+    func didFinishScanningWithError(_ error: PrivateKeyReaderError) {
```
bchrisarriola: TODOs?

Blockchain/Coordinators/KeyImportCoordinator.swift
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        AlertViewPresenter.shared.standardNotify(message: error, title: LocalizationConstants.Errors.error, handler: nil)
+    }
+
+    func failedToImportPrivateKeyForWatchOnlyAddress(errorDescription: String) {
+        LoadingViewPresenter.shared.hideBusyView()
+        walletManager.wallet.isSyncing = false
+
+        // TODO: improve JS error handling to avoid string comparisons
```
bchrisarriola: ++

Blockchain/Coordinators/KeyImportCoordinator.swift
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// MARK: - WalletKeyImportDelegate
+
+extension KeyImportCoordinator: WalletKeyImportDelegate {
+    @objc func alertUserOfImportedIncorrectPrivateKey() {
+        NotificationCenter.default.removeObserver(self, name: backupKey, object: nil)
+        let importedKeyButForIncorrectAddress = LocalizationConstants.AddressAndKeyImport.importedKeyButForIncorrectAddress
+        let importedKeyDoesNotCorrespondToAddress = LocalizationConstants.AddressAndKeyImport.importedKeyDoesNotCorrespondToAddress
+        let message = String(format: "%@\n\n%@", importedKeyButForIncorrectAddress, importedKeyDoesNotCorrespondToAddress)
```
bchrisarriola: Can also be written as "\(importedKeyButForIncorrectAddress)\n\n\(importedKeyDoesNotCorrespondToAddress)"

bchrisarriola:
Finished reviewing @achtenhagen. So #184 should be merged first?

achtenhagen:
Yes that's correct.

Blockchain/Coordinators/KeyImportCoordinator.swift
achtenhagen: What's preventing this from working right now is that whatever calls the `PrivateKeyReader` must implement its delegate because it specifies different behavior on success and failure of the scan. This is why the Key Reader and Importer have been decoupled.

PR #179, opened 2018-05-22
fix(WalletManager): re-implement payment notice delegate method

woudini:


Blockchain/Wallet.h
```diff
@@ -104,7 +104,7 @@
 - (void)updateLoadedAllTransactions:(NSNumber *)loadedAll;
 - (void)receivedTransactionMessage;
 - (void)paymentReceivedOnPINScreen:(NSString *)amount assetType:(LegacyAssetType)assetType;
-- (void)didReceivePaymentNotice:(NSString *)notice;
+- (void)didReceivePaymentNotice:(NSString *_Nullable)notice;
```
bchrisarriola: Should this be `_Nonnull`?

Blockchain/TabControllerManager.m
```diff
@@ -343,6 +343,13 @@ - (void)receivedTransactionMessage
     }
 }
 
+- (void)didReceivePaymentNoticeWithNotice:(NSString *)notice
+{
+    if (self.tabViewController.selectedIndex == TAB_SEND && LoadingViewPresenter.sharedInstance.busyView.alpha == 0 && !AuthenticationCoordinator.shared.pinEntryViewController && !self.tabViewController.presentedViewController) {
```
bchrisarriola: If notice is nullable (in this case it is), check for nullability before displaying an alert

PR #186, opened 2018-05-22
fix(Transactions-Bch): re-implement didGetBitcoinCashHistory

woudini:


bchrisarriola:
What delegate method gets invoked after BTC history?

woudini:
`on_get_history_success` is called after BTC history. Now that `getMessages` is no longer being called, it may be redundant since `didGetMultiaddressResponse` is called after it.

PR #187, opened 2018-05-22
fix(Login): present alert after adding password modal to view hierarchy

woudini:


woudini:
No, it fixed the offset password modal

PR #188, opened 2018-05-22
refactor(swipe): creating AssetAddressRepository and WalletSwipeAddressDelegate.

bchrisarriola:
* Migrated out `didGetSwipeAddresses` from `RootService`.
* Created `AssetAddressRepository` component for retrieving addresses—would like to move latestMultiAddress response (currently in WalletManager) in there as well.
* Removed all direct references to the KeychainItemWrapper and instead reference the new `AssetAddressRepository`

### DIFF
--- a/PinEntry/Classes/PEPinEntryController.m
+++ b/PinEntry/Classes/PEPinEntryController.m
@@ -193,15 +193,7 @@ static PEViewController *VerifyController()
     AssetAddressRepository *assetAddressRepository = AssetAddressRepository.sharedInstance;
     if (assetType == LegacyAssetTypeBitcoin || assetType == LegacyAssetTypeBitcoinCash) {
 
-        AssetType type;
-        switch (assetType) {
-            case LegacyAssetTypeBitcoin:
-                type = AssetTypeBitcoin;
-            case LegacyAssetTypeBitcoinCash:
-                type = AssetTypeBitcoinCash;
-            case LegacyAssetTypeEther:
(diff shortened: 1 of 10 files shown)


PinEntry/Classes/PEPinEntryController.m
```diff
@@ -190,8 +190,20 @@ - (void)setupQRCode
...
+
+        AssetType type;
+        switch (assetType) {
+            case LegacyAssetTypeBitcoin:
+                type = AssetTypeBitcoin;
+            case LegacyAssetTypeBitcoinCash:
+                type = AssetTypeBitcoinCash;
+            case LegacyAssetTypeEther:
+                type = AssetTypeEthereum;
+        }
```
woudini: Any reason to not have a function converting `LegacyAssetType` to `AssetType`?

woudini:
Not compiling for me at 6cac6638a74c67c5c3f3c80b6859e0887180e109
screen shot 2018-05-22 at 9 00 02 pm

PinEntry/Classes/PEPinEntryController.m
bchrisarriola: Mostly because it's going to be deprecated. But let me add a method within `PEPinEntryController`.

bchrisarriola:
Hmm try doing a clean build? I'm not getting that issue

woudini:
It compiles now 👍

PR #189, opened 2018-05-23
fix(pairing): ask to set pin after pairing.

bchrisarriola:
Bug was because `topMostViewController` is non-deterministic—sometimes it was `ECSlidingViewController`, other times it was `PairingCodeParser`. Fix is to always use `rootViewController`.


bchrisarriola:
> Not related to these changes specifically, but it looks like the authHandler is nil when manual pairing, so I don't get the set PIN prompt - remind me why it's set to nil after execution?

For it not to be nil, authentication needs to go through one of the `authenticate` methods in `AuthenticationManager`

bchrisarriola:
@woudini also, `authHandler` will be `nil` after authentication completes.

PR #191, opened 2018-05-23
fix(FiatAtTime): fix conflicting parameter types

woudini:
The order of the LegacyAssetType enum (BTC, BCH, ETH) being different from the original (BTC, ETH, BCH) was causing the fiat at time to fail because of conflicting parameter types. Can we change that?


bchrisarriola:
👍 that we should change the order

PR #192, opened 2018-05-23
fix(Swipe-to-Receive): move and finish implementation of check for un…

woudini:
…used addresses

### DIFF
--- a/Blockchain/Addresses/AssetAddressRepository.swift
+++ b/Blockchain/Addresses/AssetAddressRepository.swift
@@ -66,7 +66,7 @@ import Foundation
     ///
     /// - Parameter assetType: the AssetType
     /// - Returns: the swipe addresses
-    @objc func swipeToReceiveAddresses(for assetType: AssetType) -> [String] {
+    @objc func swipeToReceiveAddresses(for assetType: AssetType) -> [Any] {
         if assetType == .ethereum {
             if let swipeAddressForEther = BlockchainSettings.App.shared.swipeAddressForEther {
                 return [swipeAddressForEther]
@@ -75,7 +75,7 @@ import Foundation
             }
         }
 
-        return KeychainItemWrapper.getSwipeAddresses(for: assetType.legacy) as? [String] ?? []
+        return KeychainItemWrapper.getSwipeAddresses(for: assetType.legacy) ?? []
     }
 
     /// Removes the first swipe address for assetType.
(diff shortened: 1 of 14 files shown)


Blockchain/Addresses/AssetAddressRepository.swift
```diff
@@ -66,7 +66,7 @@ import Foundation
     ///
     /// - Parameter assetType: the AssetType
     /// - Returns: the swipe addresses
-    @objc func swipeToReceiveAddresses(for assetType: AssetType) -> [String] {
+    @objc func swipeToReceiveAddresses(for assetType: AssetType) -> [Any] {
```
bchrisarriola: Why was this changed to `Any`? I think this should be reverted to `String` since addresses won't ever be any other type.

Blockchain/Addresses/AssetAddressRepository.swift
```diff
@@ -99,9 +99,51 @@ import Foundation
...
         }
     }
 }
+
+extension AssetAddressRepository {
+    @objc func checkForUnusedAddress(_ address: String,
+                                     displayAddress: String,
+                               legacyAssetType: LegacyAssetType,
+                               successHandler: @escaping ((_ address: String, _ isUnused: Bool) -> Void),
+                               errorHandler: @escaping (() -> Void)) {
```
bchrisarriola: Alignment is a bit weird here. I think this structure is much more readable:

```
@objc func checkForUnusedAddress(
    _ address: String,
    displayAddress: String,
    legacyAssetType: LegacyAssetType,
    successHandler: @escaping ((_ address: String, _ isUnused: Bool) -> Void),
    errorHandler: @escaping (() -> Void)
) {
// ...
}
```

PinEntry/Classes/PEPinEntryController.m
```diff
@@ -222,7 +222,19 @@ - (void)addAddressToSwipeView:(BCSwipeAddressView *)swipeView assetType:(LegacyA
...
                 }
             };
-            [app checkForUnusedAddress:nextAddress success:success error:error assetType:assetType];
+            
+            NSString *addressToCheck = nextAddress;
+            NSString *displayAddress = nextAddress;
+
+            if (assetType == LegacyAssetTypeBitcoinCash) {
+                if ([nextAddress isKindOfClass:[NSString class]]) {
+                    [KeychainItemWrapper removeAllSwipeAddressesForAssetType:LegacyAssetTypeBitcoinCash];
```
bchrisarriola: Use `AssetAddressRepository`. UIViewControllers shouldn't interface with low-level constructs like the keychain/nsuserdefaults etc.

Blockchain/Addresses/AssetAddressRepository.swift
```diff
@@ -99,9 +99,51 @@ import Foundation
...
-    func onRetrievedSwipeToReceive(addresses: [String], assetType: AssetType) {
+    func onRetrievedSwipeToReceive(addresses: [Any], assetType: AssetType) {
         addresses.forEach {
             KeychainItemWrapper.addSwipeAddress($0, assetType: assetType.legacy)
         }
     }
 }
+
+extension AssetAddressRepository {
+    @objc func checkForUnusedAddress(_ address: String,
```
bchrisarriola: This used to accept `AssetAddress`, any reason that was changed?

Blockchain/Addresses/AssetAddressRepository.swift
woudini: Sorry I should have clarified in the PR description or added a comment and asked for feedback. Currently in the `swift` branch BCH addresses are stored in the BCH format `bitcoincash:address`. However the BCH lookup URL for `checkForUnusedAddress` requires the corresponding BTC address. Before this refactor,`checkForUnusedAddress` was calling `[app.wallet fromBitcoinCash:address]`. Here however the `WalletManager` singleton or a part of it is not yet initialized at that point, so I chose to do the following:
1. store a dictionary of both the BCH address and corresponding BTC address when retrieving the addresses
2. if the current BCH address is not in dictionary format, remove swipe addresses for BCH

I did this under the impression that we were on schedule to get a build done today.

Blockchain/Addresses/AssetAddressRepository.swift
woudini: I can't create an `AssetAddress` in obj-c

bchrisarriola: For it to be accessible in objective-c, you'll need to add the `@objc` annotation to the class.

i.e. 
```
@objc public protocol AssetAddress
```

bchrisarriola:
It's generally not a good pattern to return very generic objects (such as `id`). This is why we've been refactoring NSDictionary references to very specific concrete objects (i.e. `AssetAddress`, `BitcoinUrlPayload`, etc.). It's easier to work with typed objects since you know what contract you can use/expect.

Since we're mostly trying to finish `RootService` deprecation, can we break this PR up in 2? (1) migrate to use the existing `checkForUnusedAddress` and (2) implement BTC/BCH

bchrisarriola:
What's the status with this PR? Should we close this if you're creating a new one or do you plan on updating it?

woudini:
I'll do a new one since #199 has some of these changes

PR #194, opened 2018-05-23
Improve Address & Key Importing

achtenhagen:


Blockchain/AccountsAndAddressesViewController.m
```diff
@@ -200,13 +200,10 @@ - (void)importAddress
         return;
     }
 
-    PrivateKeyReader *privateKeyReader = [[PrivateKeyReader alloc] initWithAssetType:self.assetType acceptPublicKeys:YES];
+    [[KeyImportCoordinator sharedInstance] startWith:self assetType:self.assetType];
+    PrivateKeyReader *privateKeyReader = [[KeyImportCoordinator sharedInstance] privateKeyReader];
```
bchrisarriola: I think `privateKeyReader` should actually be `private` and only internal to `KeyImportCoordinator`.

In the `start` method, you can pass in a `UIViewController` which will be in charge of presenting `privateKeyReader` which is all internal to `KeyImportCoordinator`.

Blockchain/Coordinators/KeyImportCoordinator.swift
```diff
@@ -23,28 +23,45 @@ import Foundation
         case privateKeyOfAnotherNonWatchOnlyAddress
     }
 
-    static let shared = KeyImportCoordinator()
+    //: Nil if device input is unavailable
+    @objc var privateKeyReader: PrivateKeyReader?
```
bchrisarriola: `private`

Blockchain/Coordinators/KeyImportCoordinator.swift
```diff
@@ -23,28 +23,45 @@ import Foundation
...
         self.walletManager = walletManager
         super.init()
         self.walletManager.keyImportDelegate = self
     }
 
-    func start() {}
+    func start() { /* Tasks which do not require scanning should call this */ }
 
-    // MARK: - Temporary Objective-C bridging methods for backwards compatibility
+    func start(with delegate: PrivateKeyReaderDelegate, assetType: AssetType = .bitcoin) {
```
bchrisarriola: From my suggestion above, we can change this signature to:

```
func start(with delegate: PrivateKeyReaderDelegate & UIViewController, assetType: AssetType = .bitcoin) { ... }
```

or

```
func start(with delegate: PrivateKeyReaderDelegate, in viewController: UIViewController, assetType: AssetType = .bitcoin) { ... }
```

I think the latter is cleaner though

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -50,17 +50,21 @@ final class PrivateKeyReader: UIViewController & AVCaptureMetadataOutputObjectsD
     //: Legacy Objc delegate to support Legacy asset types
     @objc weak var legacyDelegate: LegacyPrivateKeyDelegate?
 
-    private var acceptPublicKeys: Bool!
+    @objc var acceptPublicKeys = false
```
bchrisarriola: revert

Blockchain/AccountsAndAddressesViewController.m
bchrisarriola: Coordinators in general should not leak internal state such as UIViewControllers they present. We did have to make some exceptions in App/Auth Coordinator though because of existing designs—something that needs to be addressed.

Blockchain/Coordinators/KeyImportCoordinator.swift
achtenhagen: Would this not lead to self being passed in twice?
```
[[KeyImportCoordinator sharedInstance] startWith:self in:self assetType:self.assetType];
```

achtenhagen: Your first suggestion would result in:
```
[[KeyImportCoordinator sharedInstance] startWith:self assetType:self.assetType];
```

which I would prefer.

bchrisarriola: There are trade offs for both. One is more configurable and lends itself to an easier unit test (passing in self twice), the other is more concise. I'm fine with either.

PR #196, opened 2018-05-23
refactor(RootService): remove app references.

bchrisarriola:
There are still a couple remaining but those methods still have to be migrated

### DIFF
--- a/Blockchain/URL/BitcoinURLPayload.swift
+++ b/Blockchain/URL/BitcoinURLPayload.swift
@@ -33,28 +33,24 @@ extension BitcoinURLPayload {
             return nil
         }
 
-<<<<<<< Updated upstream
-        let queryArgs = url.queryArgs
-
-        let address = url.host ?? queryArgs["address"]
-        let amount = queryArgs["amount"]
-
-        self.init(address: address, amount: amount)
-=======
+        let address: String?
+        let amount: String?
         let urlString = url.absoluteString
+        
         if urlString.contains("//") {
             let queryArgs = url.queryArgs
 
-            self.address = url.host ?? queryArgs["address"]
-            self.amount = queryArgs["amount"]
+            address = url.host ?? queryArgs["address"]
+            amount = queryArgs["amount"]
         } else if let commaIndex = urlString.index(of: ":") {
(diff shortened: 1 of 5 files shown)


Blockchain/URL/BitcoinURLPayload.swift
```diff
@@ -30,7 +35,9 @@ extension BitcoinURLPayload {
 
         let queryArgs = url.queryArgs
 
-        self.address = url.host ?? queryArgs["address"]
-        self.amount = queryArgs["amount"]
+        let address = url.host ?? queryArgs["address"]
```
woudini: `address` is being set to nil with a `URL` of `bitcoin:1Amu4uPJnYbUXX2HhDFMNq7tSneDwWYDyv` (web format). Should probably add support for the `bitcoin:` format and add it to `BitcoinURLPayload` tests.

bchrisarriola: Good catch. Can ether or bitcoincash also be formatted in that way?

Blockchain/URL/BitcoinURLPayload.swift
```diff
@@ -9,17 +9,22 @@
...
+        self.address = address
+        self.amount = amount
+    }
 }
 
 extension BitcoinURLPayload {
-    init?(url: URL) {
+    @objc convenience init?(url: URL) {
         guard let scheme = url.scheme else {
             return nil
```
woudini: This `guard` statement and the one below it are a bit too strict. Ideally the scanner should be "smart" enough to allow both `bitcoin:address` and just `address`, and the same with BCH. Sorry I should have mentioned earlier.

bchrisarriola: I don't think it's unreasonable to expect that a URL should have a scheme using this initializer. In the case of just `address`, you can always use the main initializer `BitcoinURLPayload(address: address, amount: amount)`.

woudini: It's fine where it is but the scanner is too strict. Users will complain when they can't scan a BTC/BCH address with no `bitcoin:` prefix.

PR #197, opened 2018-05-23
refactor(RootService): move check for ether transaction

woudini:


Blockchain/ExchangeConfirmViewController.m
```diff
@@ -161,7 +161,12 @@ - (void)disableConfirmButton
 
 - (void)confirmButtonClicked
 {
-    if ([[self.trade.depositCurrency uppercaseString] isEqualToString:CURRENCY_SYMBOL_ETH] && [app checkIfWaitingOnEtherTransaction]) return;
+    if ([[self.trade.depositCurrency uppercaseString] isEqualToString:CURRENCY_SYMBOL_ETH]) return;
```
bchrisarriola: This change isn't the same as it was before, both conditions need to be in the same if statement:

```
if ([[self.trade.depositCurrency uppercaseString] isEqualToString:CURRENCY_SYMBOL_ETH] && [[WalletManager sharedInstance].wallet isWaitingOnEtherTransaction]) {
    [[AlertViewPresenter sharedInstance] showWaitingForEtherPaymentAlert];
    return;
}
```

PR #198, opened 2018-05-23
fix(loading): make loading view visible over all views.

bchrisarriola:
* Adding busyView at the top-most window level. This way, no matter what views/view controllers are presented under, the loading view will always be displayed
* Making busyView `private`. We used to leak this state to external components to check if there is a loading view presented. Now, we can check if a loading view is displayed by querying `LoadingViewPresenter.shared.isLoadingShown`

### DIFF
--- a/Blockchain/LoadingViewPresenter.swift
+++ b/Blockchain/LoadingViewPresenter.swift
@@ -13,12 +13,16 @@ import Foundation
 
     static let shared = LoadingViewPresenter()
 
-    @objc private(set) lazy var busyView: BCFadeView = {
+    @objc private lazy var busyView: BCFadeView = {
         let busyView = BCFadeView.instanceFromNib()
         busyView.frame = UIScreen.main.bounds
         busyView.alpha = 0.0
         return busyView
     }()
+
+    @objc var isLoadingShown: Bool {
+        return busyView.superview != nil && busyView.alpha == 1.0
+    }
 
     /// sharedInstance function declared so that the LoadingViewPresenter singleton can be accessed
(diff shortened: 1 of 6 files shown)


achtenhagen:
🙌

achtenhagen:
During the Import Address flow, I am getting this warning:
```
This application is modifying the autolayout engine from a background thread after the engine was accessed from the main thread. This can lead to engine corruption and weird crashes.
```

Blockchain/LoadingViewPresenter.swift
```diff
@@ -53,28 +56,19 @@ import Foundation
...
             return
         }
 
-        busyView.labelBusy.text = text
+        guard !isLoadingShown else {
+            return
+        }
 
-        UIApplication.shared.keyWindow?.rootViewController?.view.bringSubview(toFront: busyView)
+        attachToMainWindow()
```
achtenhagen: Suggested fix: `attachToMainWindow()` needs to be placed on a different thread so the UI can update, otherwise the text label does not appear and the warning is logged.
```
DispatchQueue.main.async {
    self.attachToMainWindow()
}
```

bchrisarriola: LoadingViewPresenter (and any component that touches the UI) shouldn't ever be called from a background thread 😞 But since that's a larger fix, I'll dispatch show/hide/update on the main thread as you suggest. For our wallet rearch, we should guarantee that all callbacks are executed on the main thread and once we have that, we can remove these dispatch calls.

PR #200, opened 2018-05-24
refactor(RootService): Remove RootService

bchrisarriola:
Getting rid of RootService. Need to merge #196 before this


Blockchain/SettingsChangePasswordViewController.m
```diff
@@ -128,7 +127,6 @@ - (void)changePasswordSuccess
     UIAlertController *alertForChangePasswordSuccess = [UIAlertController alertControllerWithTitle:[LocalizationConstantsObjcBridge success] message:BC_STRING_SETTINGS_SECURITY_PASSWORD_CHANGED preferredStyle:UIAlertControllerStyleAlert];
     [alertForChangePasswordSuccess addAction:[UIAlertAction actionWithTitle:BC_STRING_OK style:UIAlertActionStyleCancel handler:nil]];
     [self.navigationController dismissViewControllerAnimated:YES completion:^{
-        app.settingsNavigationController = nil;
```
woudini: This might affect the change password flow, just not sure how.

bchrisarriola: Tested the change password flow and it works just fine.

Side-note for future refactoring:

There are a couple of UIViewControllers that are kept in memory even though they might not be presented (e.g. settingsNavigationController, etc.). Looks like those are there because we want to be able to refresh those view controllers by invoking `reload` when certain state changes. I did not refactor and fix that in the coordinators as solving it is a larger task (we should get to it one the wallet is rearchitected).

bchrisarriola: > This might affect the change password flow, just not sure how.

Looks like this was set to nil to prevent the change password view from being presented when tapping on the settings on the side menu after changing your password. I updated this behavior to always create a new `SettingsNavigationController` when tapping on the settings side menu. Fixed applied on 716fd4e

PR #201, opened 2018-05-25
Show busy view in Manual Pair view and Password Required modal

woudini:


Blockchain/BackupVerifyViewController.swift
```diff
@@ -136,7 +136,7 @@ class BackupVerifyViewController: UIViewController, UITextFieldDelegate, SecondP
             }
             if valid {
                 let backupNavigation = self.navigationController as? BackupNavigationViewController
-                backupNavigation?.busyView?.fadeIn()
+                LoadingViewPresenter.sharedInstance().showBusyView(withLoadingText: LocalizationConstants.syncingWallet)
```
bchrisarriola: Use `LoadingViewPresenter.shared` here

Blockchain/Authentication/Views/PasswordRequiredView.swift
```diff
@@ -76,7 +76,13 @@ class PasswordRequiredView: UIView {
         textFieldPassword.resignFirstResponder()
         textFieldPassword.text = nil
 
-        delegate?.didContinue(with: cleanedPassword)
+        LoadingViewPresenter.shared.showBusyView(withLoadingText: LocalizationConstants.Authentication.downloadingWallet)
+
+        perform(#selector(login), with: cleanedPassword, afterDelay: Constants.Animation.durationLong)
```
bchrisarriola: What is this delay for?

Blockchain/BCManualPairView.m
```diff
@@ -135,6 +135,15 @@ - (IBAction)continueClicked:(id)sender
     BlockchainSettings.sharedAppInstance.hasSeenAllCards = YES;
     BlockchainSettings.sharedAppInstance.shouldHideAllCards = YES;
 
+    [[LoadingViewPresenter sharedInstance] showBusyViewWithLoadingText:[LocalizationConstantsObjcBridge downloadingWallet]];
+    
+    [self performSelector:@selector(pair) withObject:nil afterDelay:[ConstantsObjcBridge animationDurationLong]];
```
bchrisarriola: I don't think we should be adding arbitrary delays

Blockchain/js/wallet-ios.js
```diff
@@ -749,8 +749,6 @@ MyWalletPhone.login = function(user_guid, shared_key, resend_code, inputedPasswo
         objc_show_email_authorization_alert();
     }
 
-    objc_loading_start_download_wallet();
```
bchrisarriola: Remove `loading_start_download_wallet` in obj-c as well?

Blockchain/Authentication/Views/PasswordRequiredView.swift
bchrisarriola: Doesn't the JSContext always execute in its own thread? I don't think adding an arbitrary delay is an acceptable fix

bchrisarriola: I think this bug is probably related to cac69494

woudini: I'm open to other solutions but at this point the busy views are not appearing on the manual pair view and password required view when they should be, and the kind of solution that we know will work is concurrency-related.

PR #202, opened 2018-05-25
fix(loading): Loading presentation

bchrisarriola:
The loading/busy view was not being presented in some areas (manual pairing, password required view, etc.) because the action was dispatched asynchronously to the GCD. So, if that queue is congested (like in the case of password auth since it does), the loading view may not be presented. This change fixes that issue by guaranteeing that an invocation to one of the `LoadingViewPresenter` methods execute on the main thread (previous was async and passed to the main queue).

As I brought up before, we want wallet delegate methods to always be invoked on the main thread (removes all the guesswork whether or not a callback is executed on the main or background thread). I went ahead and guaranteed that by calling the wallet delegate methods from within a block dispatched to GCD on the main queue. Also, we have some other LoadingViewPresenter references in the Wallet.m that I also wrapped inside a block that gets executed on the main thread.

Fixes #201 

Other change: implement the correct method for `WalletTransactionDelegate`.

### DIFF
--- a/Blockchain/Authentication/AuthenticationCoordinator.swift
+++ b/Blockchain/Authentication/AuthenticationCoordinator.swift
@@ -499,6 +499,9 @@ import Foundation
 
 extension AuthenticationCoordinator: ManualPairViewDelegate {
     func manualPairView(_ manualPairView: BCManualPairView!, didContinueWithGuid guid: String!, andPassword password: String!) {
+
+        LoadingViewPresenter.shared.showBusyView(withLoadingText: LocalizationConstants.Authentication.downloadingWallet)
+
         let payload = PasscodePayload(guid: guid, password: password, sharedKey: "")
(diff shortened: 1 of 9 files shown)


bchrisarriola:
I'm not seeing the issue where the busy view is being dismissed prematurely—what specific user flow is that happening on? Does it happen when entering the pin after background the app? I don't think `hideBusyView` should be called in in `PEPinEntryViewController` since that view controller should now know any details about a busyview being presented/hidden (i.e. it should delegate that responsibility to the `PEPinEntryControllerDelegate`).

As far as the animation, that seems like a separate issue. I'm not seeing that but it could be related to using `UIView.animate` vs. calling `UIView.beginAnimations` from cac69494. Will test.

woudini:
It's dismissed early after entering the PIN on login.

woudini:
Just pulled and updated, logging in with PIN looks good but manual pairing is now not showing the busy view.

PR #203, opened 2018-05-26
fix(modal): prevent modal presentation on UIAlertViewController

bchrisarriola:
`PasswordRequiredView` sometimes gets presented in a cut-off way. This change fixes that.

### DIFF
--- a/Blockchain/UIViewController.swift
+++ b/Blockchain/UIViewController.swift
@@ -21,3 +21,11 @@ extension UINavigationController {
         return visibleViewController ?? self
     }
 }
+
+extension UIAlertController {
+
+    /// Overridden so that UIAlertControllers will never show up as the `topMostViewController`.
+    override var topMostViewController: UIViewController? {
+        return presentedViewController?.topMostViewController
+    }
+}


Blockchain/UIViewController.swift
```diff
@@ -21,3 +21,11 @@ extension UINavigationController {
         return visibleViewController ?? self
     }
 }
+
+extension UIAlertController {
+
+    /// Overridden so that UIAlertControllers will never show up as the `topMostViewController`.
+    override var topMostViewController: UIViewController? {
+        return presentedViewController?.topMostViewController
```
woudini: Can you explain how this works? Normally a `UIAlertController` doesn't have a `presentedViewController`, so won't this just return nil?

bchrisarriola: Yep, we want to return nil in that case. Previously we were returning `self` (i.e. the `UIAlertController`) when the  top most view controller is a `UIAlertController`. In the scenario when a `UIAlertController` is actually the top most view controller, this change will simply return `nil` and not self so that the value returned to the caller invoking `topMostViewController` will be the view controller presenting the `UIAlertController`.

PR #204, opened 2018-05-30
About

woudini:


bchrisarriola:
Should we also display Privacy Policy and Cookie Policy in the wallet creation screen (currently only displays TOS)?

simulator screen shot - iphone 5s - 2018-05-30 at 11 56 06

woudini:
It's not on web, but that may be a question for Marketing.

PR #205, opened 2018-05-30
Improved Address Validation

achtenhagen:
Highlights in this PR:

- Encapsulates address validation in `AddressValidator` class.
- Fixes broken unit tests due to previous incorrect address validation.
- Makes `AssetAddress` class Objective-C compatible.
- Migrates `escapedForJS` to String extension in Swift + adds unit tests.
- Adds unit tests for the `AddressValidator` + adds more tests to existing ones.
- Exposes `loadJS` in `wallet.h` for JS access prior to wallet initialization.

Note on the last point:
*This separation should be part of the wallet rearchitecture.*


Blockchain/Addresses/AddressValidator.swift
```diff
@@ -0,0 +1,63 @@
...
+//  AddressValidator.swift
+//  Blockchain
+//
+//  Created by Maurice A. on 5/24/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+// TODO: use associatedtype in protocol once there are no more objc dependents
```
bchrisarriola: I find that getting a warning for TODOs from swiftlint to be a bit aggressive. Thoughts on removing that warning from swiftlint?

Blockchain/Addresses/AddressValidator.swift
```diff
@@ -0,0 +1,63 @@
...
+
+    @objc class func sharedInstance() -> AddressValidator? {
+        return shared
+    }
+
+    // MARK: - Initialization
+
+    @objc
+    private init?(context: JSContext? = WalletManager.shared.wallet.context) {
+        guard let JSContext = context else { return nil }
```
bchrisarriola: 2 thoughts here:

1. The wallet currently invokes `loadJS` a few times which instantiates a new `JSContext`. When that happens, the address validators also need to be updated with the new `JSContext`. Unclear to me why `loadJS` is called multiple times, in the wallet refactor we can dig into that more, but for now, when `loadJS` is invoked, the `AddressValidator`'s context also needs to be updated (i.e. the singleton needs to be reinitialized/reconstructed).

2. I think this constructor should only accept a non-optional JSContext.

bchrisarriola: Also, use lowercase for property names (i.e. s/JSContext/jsContext)

Blockchain/Addresses/AssetAddressRepository.swift
```diff
@@ -128,8 +128,8 @@ extension AssetAddressRepository {
         let assetType = AssetType.from(legacyAssetType: legacyAssetType)
 
         switch assetType {
-        case .bitcoin: assetAddress = BitcoinAddress(string: address)!
-        case .bitcoinCash: assetAddress = BitcoinCashAddress(string: address)!
+        case .bitcoin: assetAddress = BitcoinAddress(string: address)
```
bchrisarriola: 🔥

Blockchain/Addresses/BitcoinAddress.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Created by Maurice A. on 4/26/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+// TODO: convert class to struct once there are no more objc dependents
+
+@objc
+internal class BitcoinAddress: NSObject & AssetAddress {
```
bchrisarriola: `public`?

Blockchain/Addresses/BitcoinCashAddress.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Created by Maurice A. on 4/26/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+// TODO: convert class to struct once there are no more objc dependents
+
+@objc
+internal class BitcoinCashAddress: NSObject & AssetAddress {
```
bchrisarriola: `public`?

Blockchain/Addresses/EthereumAddress.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Created by Maurice A. on 5/24/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+// TODO: convert class to struct once there are no more objc dependents
+
+@objc
+internal class EthereumAddress: NSObject & AssetAddress {
```
bchrisarriola: `public`?

Blockchain/Addresses/AddressValidator.swift
bchrisarriola: Changed my mind about this

Blockchain/Addresses/AddressValidator.swift
achtenhagen: I think it should not be a singleton because for that to hold true, the class needs to be usable entirely in isolation, whereas here we have the class depend on the JS context. I'm not a fan of trying to update the singleton's `context` variable each time `loadJS()` is called, so I have made it a regular class, that when initialized, takes a context object (that way it guarantees to always use the latest `WalletManager.shared.wallet.context` value). So right now it would look something like:
```
let validator = AddressValidator(context: WalletManager.shared.wallet.context)
guard validator.validate(bitcoinAddress: address) else { return }
```

If we assume the default value of the initializer to be `WalletManager.shared.wallet.context`, the tests will break as this will not always the home of the `context`. It should be decoupled from the wallet anyways

PR #208, opened 2018-05-31
fix(Addresses): show second password prompt correctly

woudini:
Changed `UINavigationController`'s `topViewController` to `self` instead of `visibleViewController` to prevent this from happening. No change in busy view behavior in manual pairing, auto pairing, password required login, or recovery behavior compared to `dev`:
simulator screen shot - iphone 6 - 2018-05-31 at 11 42 16

### DIFF
--- a/Blockchain/AccountsAndAddressesViewController.m
+++ b/Blockchain/AccountsAndAddressesViewController.m
@@ -18,7 +18,7 @@
 
 #define CELL_HEIGHT_DEFAULT 44.0f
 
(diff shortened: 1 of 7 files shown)


woudini:
I changed it to not use the `BCModalViewController` and to use the modal instead. The busy view is still an issue though now second password can be used. Thoughts?

bchrisarriola:
@woudini looks good to me. So the busy view isn't presented at all when it should? We can address in a separate PR if so

PR #209, opened 2018-06-01
update(About): add cookie policy and update other About URLs

woudini:
Note that this will be merged into `release` and not `dev`.


woudini:
Yeah, there is no swift localization file or bridge in `release` yet.

PR #211, opened 2018-06-01
fix(LoadingViewPresenter): begin deprecating topViewController and sh…

woudini:
…ow busy view after second password entry if applicable

### DIFF
--- a/Blockchain/Authentication/AuthenticationCoordinator.swift
+++ b/Blockchain/Authentication/AuthenticationCoordinator.swift
@@ -132,6 +132,7 @@ import Foundation
         self.walletManager = walletManager
         super.init()
         self.walletManager.pinEntryDelegate = self
+        self.walletManager.secondPasswordDelegate = self
     }
 
     // MARK: - Start Flows
@@ -415,6 +416,10 @@ import Foundation
         validateSecondPassword: Bool,
         confirmHandler: @escaping PasswordConfirmView.OnPasswordConfirmHandler
     ) {
+        let loadingViewPresenter = LoadingViewPresenter.shared
+        let isLoadingShown = loadingViewPresenter.isLoadingShown
+        if isLoadingShown { loadingViewPresenter.hideBusyView() }
+
         let passwordConfirmView = PasswordConfirmView.instanceFromNib()
         passwordConfirmView.updateLabelDescription(text: displayText)
         passwordConfirmView.validateSecondPassword = validateSecondPassword
@@ -429,9 +434,11 @@ import Foundation
                 return
             }
 
-            confirmHandler(password)
-
             ModalPresenter.shared.closeModal(withTransition: kCATransitionFade)
+
+            if isLoadingShown { loadingViewPresenter.showBusyView(withLoadingText: loadingViewPresenter.currentLoadingText!) }
+
+            confirmHandler(password)
         }
         ModalPresenter.shared.showModal(
             withContent: passwordConfirmView,
@@ -598,6 +605,24 @@ extension AuthenticationCoordinator: SetupDelegate {
             self.lastEnteredPIN?.saveToKeychain()
 
             completion(true)
+        }
+    }
+}
+
+extension AuthenticationCoordinator: WalletSecondPasswordDelegate {
+    func getSecondPassword(success: JSValue) {
+        showPasswordConfirm(withDisplayText: LocalizationConstants.Authentication.secondPasswordDefaultDescription,
+                            headerText: LocalizationConstants.Authentication.secondPasswordRequired,
+                            validateSecondPassword: true) { (secondPassword) in
+                                success.call(withArguments: [secondPassword])
+        }
+    }
+
+    func getPrivateKeyPassword(success: JSValue) {
+        showPasswordConfirm(withDisplayText: LocalizationConstants.Authentication.privateKeyPasswordDefaultDescription,
+                            headerText: LocalizationConstants.Authentication.privateKeyNeeded,
+                            validateSecondPassword: false) { (privateKeyPassword) in
+                                success.call(withArguments: [privateKeyPassword])
         }
     }
 }
--- a/Blockchain/Authentication/AuthenticationCoordinator.swift
+++ b/Blockchain/Authentication/AuthenticationCoordinator.swift
@@ -418,7 +418,9 @@ import Foundation
     ) {
         let loadingViewPresenter = LoadingViewPresenter.shared
         let isLoadingShown = loadingViewPresenter.isLoadingShown
-        if isLoadingShown { loadingViewPresenter.hideBusyView() }
+        if isLoadingShown {
+            loadingViewPresenter.hideBusyView()
+        }
 
         let passwordConfirmView = PasswordConfirmView.instanceFromNib()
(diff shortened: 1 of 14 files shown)


Blockchain/LoadingViewPresenter.swift
```diff
@@ -49,13 +43,11 @@ import Foundation
         self.busyView.fadeOut()
     }
 
-    @objc func showBusyView(withLoadingText text: String) {
-        let topMostViewController = UIApplication.shared.keyWindow?.rootViewController?.topMostViewController
+    @objc func currentLoadingText() -> String? {
```
bchrisarriola: This can be a computed property since it doesn't take any params:

```
@objc var loadingtext: String? {
    return self.busyView.labelBusy.text
}
```

Blockchain/Wallet.m
```diff
@@ -3203,7 +3203,12 @@ - (void)getPrivateKeyPassword:(NSString *)canDiscard success:(JSValue *)success
 
 - (void)getSecondPassword:(NSString *)canDiscard success:(JSValue *)success error:(void(^)(id))_error helperText:(NSString *)helperText
 {
+    LoadingViewPresenter *loadingViewPresenter = [LoadingViewPresenter sharedInstance];
```
bchrisarriola: Since you are changing this behavior—I think we should follow the pattern where this method will call into a wallet delegate method.

How about adding a delegate protocol, `WalletDelegate` with a method `onPasswordConfirmationRequired`, and have `AuthenticationCoordinator` conform to that (which then displays the password confirm view)?

Blockchain/Wallet/WalletSecondPasswordDelegate.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Created by kevinwu on 6/4/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+protocol WalletSecondPasswordDelegate: class {
+    /// Method invoked when second password is required for JS function to complete.
+    func getSecondPassword(success: JSValue)
```
bchrisarriola: `JSValue` should not be passed around to layers outside the context of `Wallet.h/.m`—it's an implementation detail. Instead create an abstraction around it:

Something like:

```
protocol WalletSecondPasswordDelegate: class {
    /// Method invoked when second password is required for JS function to complete.
    func getSecondPassword(successCallback: WalletSuccessCallback)
}

protocol WalletSuccessCallback {
    func success(string: String)
}

extension JSValue: WalletSuccessCallback {
    func success(string: String) {
        self.call(withArguments: [string])
    }
}
```

Blockchain/Authentication/Views/PasswordConfirmView.swift
```diff
@@ -42,8 +42,10 @@ class PasswordConfirmView: UIView {
         guard let passwordText = textFieldPassword.text else {
             return
         }
-        confirmHandler?(passwordText)
-        textFieldPassword.text = nil
+        DispatchQueue.main.async {
```
bchrisarriola: Isn't `didTapContinue` already executed on the main thread? Dispatching to the main queue shouldn't be necessary

Blockchain/Authentication/AuthenticationCoordinator.swift
```diff
@@ -429,9 +434,11 @@ import Foundation
                 return
             }
 
-            confirmHandler(password)
-
             ModalPresenter.shared.closeModal(withTransition: kCATransitionFade)
+
+            if isLoadingShown { loadingViewPresenter.showBusyView(withLoadingText: loadingViewPresenter.currentLoadingText!) }
```
bchrisarriola: Should this check `if !isLoadingShown`? Either that, or should the busyView be hidden inside of the if statement?

Blockchain/Authentication/AuthenticationCoordinator.swift
```diff
@@ -415,6 +416,10 @@ import Foundation
         validateSecondPassword: Bool,
         confirmHandler: @escaping PasswordConfirmView.OnPasswordConfirmHandler
     ) {
+        let loadingViewPresenter = LoadingViewPresenter.shared
+        let isLoadingShown = loadingViewPresenter.isLoadingShown
+        if isLoadingShown { loadingViewPresenter.hideBusyView() }
```
bchrisarriola: Nitpick: I think adding a new line between opening/closing parens helps with readability:

```
if isLoadingShown {
    loadingViewPresenter.hideBusyView()
}
```

Blockchain/Wallet/WalletManager.swift
```diff
@@ -644,4 +645,13 @@ extension WalletManager: WalletDelegate {
             self.keyImportDelegate?.scanPrivateKeyForWatchOnlyAddress(address)
         }
     }
+
+    // MARK: - Second Password
+    @objc func getSecondPassword(success: JSValue) {
+        secondPasswordDelegate?.getSecondPassword(success: success)
+    }
+
+    @objc func getPrivateKeyPassword(success: JSValue) {
```
bchrisarriola: `success` needs to be of type `WalletSuccessCallback`.

Blockchain/Wallet/WalletManager.swift
```diff
@@ -644,4 +645,13 @@ extension WalletManager: WalletDelegate {
             self.keyImportDelegate?.scanPrivateKeyForWatchOnlyAddress(address)
         }
     }
+
+    // MARK: - Second Password
+    @objc func getSecondPassword(success: JSValue) {
+        secondPasswordDelegate?.getSecondPassword(success: success)
```
bchrisarriola: success needs to be of type WalletSuccessCallback.

Blockchain/Authentication/AuthenticationCoordinator.swift
bchrisarriola: Everything else looks good, just need to address this feedback @woudini

woudini: It should show the busy view if `isLoadingShown` is true because it was hidden before in line 422. The reason for this is to ensure the busy view doesn't show over the second password view (otherwise all gestures are blocked). A better name for this could be `shouldShowLoading` but then line 421 wouldn't make sense. It's hard to find a good name for this boolean.

bchrisarriola: Ah, that's what confused me.

Actually let's move the loading presentation show/hide logic out of the `showPasswordConfirm` method and instead leave it up to the caller (i.e. `getSecondPassword` and `getPrivateKeyPassword`). This way, `showPasswordConfirm` only does one thing and doesn't need to know about the busy view presentation state.

woudini: The only way I see this working is by adding a wrapper function around `showPasswordConfirm`, but I can't think of a good name, any suggestions?

bchrisarriola: Actually looks like `showPasswordConfirm` is called in a bunch of places and would be tedious to handle loading presentation there. Let's keep it as you have it implemented

PR #220, opened 2018-06-06
feat(pin): Creating presenter and interactor for pin.

bchrisarriola:
Changes:
* Created `PinPresenter`
* Create `PinInteractor`
* Created tests for `PinPresenter`

Next step:
* Remove old pin validation from `Wallet.h/.m` and `wallet-ios.js` and use these new classes. That should be the last change before `wallet_rearch` can be merged into `dev`

Note:
For more context on these design patterns: https://medium.com/@smalam119/viper-design-pattern-for-ios-application-development-7a9703902af6


Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
```diff
@@ -77,6 +77,7 @@ extension AuthenticationCoordinator: PEPinEntryControllerDelegate {
 
         showVerifyingBusyView(withTimeout: 30)
 
+        // TODO: Only save to keychain when putting pin, not getting
```
woudini: Not sure I understand this, can you elaborate?

bchrisarriola: Should've elaborated more but what I meant here is that we should only be updating the stored pin that's in the keychain when you:
* first set up your pin
* change your pin
* or when your pin gets deleted (after 4 incorrect tries)
 
Currently we are updating the stored pin even on validation (i.e. what I meant by "getting") which is not necessary.

bchrisarriola:
> Missing change where Alamofire was added
That was added in PR #215 

Similar to the iPhone X feature, since this change is quite a lot, I'm breaking it up into smaller reviewable PRs against `wallet_rearch`.

Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
woudini: I remember the reason for storing the pin on validation being that there wasn't an easy way to store it whenever wallet decryption succeeded, but hopefully the rearch will make this todo feasible.

bchrisarriola: yeah I think that would be easy to do with the rearch

PR #222, opened 2018-06-06
feat(BuySell): add buy sell coordinator and set as delegate

woudini:
Mainly to make sure the buy sell webview is initialized (previously was dependent on `awakeFromNib` being called in `BCCreateWalletView`, which now no longer happens).


Blockchain/AppDelegate.swift
```diff
@@ -101,7 +101,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         wallet.isFetchingTransactions = false
         wallet.isFilteringTransactions = false
         wallet.didReceiveMessageForLastTransaction = false
-        wallet.setupBuySellWebview()
+        BuySellCoordinator.shared.start()
```
bchrisarriola: This was something that wasn't too clear to me. Why does buy/sell need to be set up when the app goes to the background (seems like it's something that should be set up when the app comes to the foreground)?

Blockchain/Coordinators/AppCoordinator.swift
```diff
@@ -142,53 +141,7 @@ import Foundation
     }
 
     @objc func showBuyBitcoinView() {
```
bchrisarriola: Let's remove this method (all references to it should just use `BuySellCoordinator.shared.showBuyBitcoinView()`)

Blockchain/Network/NetworkManager.swift
```diff
@@ -53,39 +53,51 @@ final class NetworkManager: NSObject, URLSessionDelegate {
         persistServerSessionIDForNewUIWebViews()
     }
 
-    func checkForMaintenance(withCompletion handler: @escaping (_ response: String?) -> Void) {
+    func getWalletOptions(withCompletion success: @escaping (_ response: Dictionary<String, Any>?) -> Void, error: @escaping(_ error: String?) -> Void) {
```
bchrisarriola: Instead of returning a `Dictionary<String, Any>`, let's return a typed struct/class called `WalletOptions` (similar to what we do with `Trade`). It's much easier for the caller to work with a typed object where the contract is strict and so the caller doesn't need to know specific keys.

Blockchain/Constants.swift
```diff
@@ -80,6 +80,10 @@ struct Constants {
         static let userInfoId = "id"
         static let typePayment = "payment"
     }
+    struct WalletOptionsKeys {
```
bchrisarriola: See my comment on creating a struct called `WalletOptions`. If we do that, we don't need this struct since the keys can be encapsulated within the `WalletOptions` struct and we don't need to expose them.

Blockchain/AppDelegate.swift
woudini: Can't remember why. Moved

bchrisarriola: Actually maybe it's better to put it in `AppCoordinator.start()`? `applicationWillEnterForeground` won't be called on first launch

Blockchain/Network/Models/WalletOptions.swift
```diff
@@ -0,0 +1,60 @@
...
+
+struct WalletOptions {
+    private struct Keys {
+        static let mobile = "mobile"
+        static let walletRoot = "walletRoot"
+        static let maintenance = "maintenance"
+        static let mobileInfo = "mobileInfo"
+    }
+
+    let downForMaintenance: Bool?
```
bchrisarriola: This can be set as non-optional and if not set default to false. If possible, non-optional properties should always be preferred as you don't need to do conditional unwrapping.

So:

```
let downForMaintenance: Bool
```

and:

```
extension WalletOptions {
    init?(response: [String: Any]) {
        self.downForMaintenance = response[Keys.maintenance] as? Bool ?? false
        self.mobile = WalletOptions.Mobile(response)
        self.mobileInfo = WalletOptions.MobileInfo(response)
    }
}
```

Blockchain/Network/NetworkManager.swift
```diff
@@ -53,39 +53,51 @@ final class NetworkManager: NSObject, URLSessionDelegate {
         persistServerSessionIDForNewUIWebViews()
     }
 
-    func checkForMaintenance(withCompletion handler: @escaping (_ response: String?) -> Void) {
+    func getWalletOptions(withCompletion success: @escaping (_ response: WalletOptions?) -> Void, error: @escaping(_ error: String?) -> Void) {
```
bchrisarriola: `response` doesn't need to be optional.

```
withCompletion success: @escaping (_ response: WalletOptions) -> Void
```

Blockchain/Network/Models/WalletOptions.swift
```diff
@@ -0,0 +1,60 @@
...
+                message = nil
+            }
+        }
+    }
+
+    let mobileInfo: MobileInfo?
+}
+
+extension WalletOptions {
+    init?(response: [String: Any]) {
```
bchrisarriola: This doesn't need to be a failable initializer since it doesn't return nil at all.

```
init(response: [String: Any]) {
    // ...
}
```

PR #226, opened 2018-06-08
feat(AssetURLPayload): add support for amounts without double slashes…

woudini:
… and plain address strings


Blockchain/StringQueryArgs.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Created by kevinwu on 6/8/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+extension String {
+
+    /// Returns query arguments from a string in URL format
+    var queryArgs: [String: String] {
```
bchrisarriola: Instead of creating a new file, let's put this in `Strings.swift`

Blockchain/URL/AssetURLPayloadFactory.swift
```diff
@@ -9,6 +9,30 @@
...
 
 @objc class AssetURLPayloadFactory: NSObject {
+
+    @objc static func create(fromString string: String, legacyAssetType: LegacyAssetType) -> AssetURLPayload? {
+        if string.contains(":") {
+            guard let url = URL(string: string) else {
+                print("Could not create payload from URL \(string)")
+                return nil
+            }
+            guard let payload = create(from: url) else {
```
bchrisarriola: This `guard` statement is not necessary since `create(from: url)` returns an optional `AssetURLPayload`. You can just do:

```
return create(from: url)
```

Blockchain/URL/AssetURLPayloadFactory.swift
```diff
@@ -9,6 +9,30 @@
 import Foundation
 
 @objc class AssetURLPayloadFactory: NSObject {
+
+    @objc static func create(fromString string: String, legacyAssetType: LegacyAssetType) -> AssetURLPayload? {
```
bchrisarriola: Let's use `AssetType` instead of `LegacyAssetType`. Ideally all Swift code should use this newer enum as we gradually deprecate `LegacyAssetType`

BlockchainTests/URL/BitcoinURLPayloadTests.swift
```diff
@@ -25,6 +25,22 @@ class BitcoinURLPayloadTests: XCTestCase {
         XCTAssertNil(payload)
     }
 
+    func testBitcoinNoFormat() {
```
bchrisarriola: This test should be moved to `AssetURLPayloadFactory`.

BlockchainTests/URL/BitcoinURLPayloadTests.swift
```diff
@@ -25,6 +25,22 @@ class BitcoinURLPayloadTests: XCTestCase {
...
+        let address = "1Amu4uPJnYbUXX2HhDFMNq7tSneDwWYDyv"
+        let payload = BitcoinURLPayload(address: address, amount: nil)
+        XCTAssertNotNil(payload)
+        XCTAssertEqual(address, payload.address)
+    }
+
+    func testBitcoinNoFormatWithAmount() {
```
bchrisarriola: Not sure this test is quite useful since all it does is set the `address` and `amount` properties directly. Maybe move to `AssetURLPayloadFactory` and use the `create(fromString string: String, assetType: AssetType)` method to verify that it's working as intended?

Blockchain/URL/AssetURLPayloadFactory.swift
woudini: Is there a better way to do this other than by copying `assetTypeFromLegacyAssetType` in `PEPinEntryController.m`?

bchrisarriola: I guess we can create an Objective-C helper class (maybe something like `LegacyHelper`) and expose `assetTypeFromLegacyAssetType` as a static function and call that. I just declared `assetTypeFromLegacyAssetType` directly in `PEPinEntryController.m` since we're working towards deprecating obj-c code anyway.

PR #229, opened 2018-06-09
fix(bch): showing BCH swipe address

bchrisarriola:
BCH swipe address was not correctly showing because we were not transforming the BCH address to a BTC address before passing it to the `https://api.blockchain.info/bch/multiaddr` endpoint.


bchrisarriola:
@woudini I think we can just try to load the JSContext sooner (i.e. in `AppCoordinator.shared.start()`) to handle that case—let me test that.

The functions under `helpers.js` is also something we can implement natively (let's look into that in the next sprint/milestone).

bchrisarriola:
With the fix in ec5acbe, when doing a fresh launch the BCH address now displays and not "Please login to load more addresses"

PR #231, opened 2018-06-11
iPhone X Support for Dashboard

achtenhagen:
Highlights in this PR:
- Adds iPhone X support to the **Dashboard Screen**.
- Adds iPhone X support to the **Tab View Controller**.

### IMPORTANT
Please make sure the commit message when merging is:
```
iPhone X Support for Dashboard (#231)
```

### DIFF
--- a/Blockchain/AppDelegate.swift
+++ b/Blockchain/AppDelegate.swift
@@ -40,8 +40,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         Fabric.with([Crashlytics.self])
 
-        //: Global appearance customizations
+        // MARK: - Global Appearance
+
+        //: Status Bar
         UIApplication.shared.statusBarStyle = .default
+
+        //: Navigation Bar
+        let navigationBarAppearance = UINavigationBar.appearance()
+        navigationBarAppearance.shadowImage = UIImage()
+        navigationBarAppearance.isTranslucent = false
+        if #available(iOS 11.0, *) {
+            navigationBarAppearance.barTintColor = UIColor(named: "ColorBrandPrimary")
+        } else {
+            navigationBarAppearance.barTintColor = Constants.Colors.ColorBrandPrimary
+        }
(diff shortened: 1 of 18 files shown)


Blockchain/AppDelegate.swift
```diff
@@ -40,9 +40,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+        //: Status Bar
         UIApplication.shared.statusBarStyle = .default
 
+        //: Navigation Bar
+        let navigationBarAppearance = UINavigationBar.appearance()
+        navigationBarAppearance.shadowImage = UIImage()
+        navigationBarAppearance.isTranslucent = false
+        if #available(iOS 11.0, *) {
+            navigationBarAppearance.barTintColor = UIColor(named: "ColorBrandPrimary")
```
bchrisarriola: Reminder to create some sort of color helper component so we can hide the `#available` checks from areas that require a brand color.

Blockchain/TabViewController.m
```diff
@@ -10,41 +10,53 @@
...
-    selectedIndex = TAB_TRANSACTIONS;
+
+    selectedIndex = TAB_DASHBOARD;
     
     [self setupTabButtons];
 }
 
+- (void)viewDidLayoutSubviews
+{
+    UIWindow *window = [UIApplication sharedApplication].keyWindow;
```
bchrisarriola: Call `[super viewDidLayoutSubviews]` here

Blockchain/TabViewController.m
```diff
@@ -139,7 +176,7 @@ - (void)setSelectedIndex:(int)nindex
     });
     
     NSArray *titles = @[BC_STRING_SEND, BC_STRING_DASHBOARD, BC_STRING_TRANSACTIONS, BC_STRING_REQUEST];
-    
+    if (nindex == 2) { return; }
```
bchrisarriola: Can you add a comment why an early return is necessary here?

achtenhagen: Sure, the reason is because the title view of the navigation bar is used for both the navigation item's title as well as the balance label. This was split into multiple views previously. And since the balance always shows for the transactions tab (as opposed to the word **Transactions**), the code returns early.

PR #233, opened 2018-06-12
fix(2fa): fix 2FA flow

bchrisarriola:
Fixes related to 2FA:
* SMS & Google Auth | fix issue where the 2FA verification dialog (sms & google auth) would not be presented if the email authentication alert is still presented
* Email Auth | update email authorization alert message to instruct the user to return to the app after they have authorized the log in attempt

### DIFF
--- a/Blockchain/Authentication/AuthenticationManager.swift
+++ b/Blockchain/Authentication/AuthenticationManager.swift
@@ -323,7 +323,7 @@ extension AuthenticationManager: WalletAuthDelegate {
     }
 
     func emailAuthorizationRequired() {
-        failAuth(withError: AuthenticationError(
+        authHandler?(false, nil, AuthenticationError(
             code: AuthenticationError.ErrorCode.emailAuthorizationRequired.rawValue
         ))
     }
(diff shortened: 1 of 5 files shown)


Blockchain/Authentication/AuthenticationManager.swift
```diff
@@ -323,7 +323,7 @@ extension AuthenticationManager: WalletAuthDelegate {
     }
 
     func emailAuthorizationRequired() {
-        failAuth(withError: AuthenticationError(
+        authHandler?(false, nil, AuthenticationError(
```
woudini: It seems like this change is causing the PIN prompt to not show up in a certain case. If the user leaves the alert up while authorizing, the user will go straight to the dashboard without any PIN prompt. `showPinEntryView` is being called but not presented for some reason. On `dev` it doesn't automatically log you in if you leave the alert up, so you have to login again and the PIN prompt is shown.

bchrisarriola: Ah, good catch. I was mostly testing the scenario where 2FA is enabled.

PR #238, opened 2018-06-12
feat(pin): use PinPresenter in PEPinEntryController.

bchrisarriola:
Using the new components (PinPresenter, PinInteractor, & WalletService) to validate the user's pin. Also removed the old pin validation code. Once this is merged, #215 can be merged, too.

**Edit:**
Hold off on reviewing this for now, please prioritize iPhone X issues & stability first.


PinEntry/Classes/PEPinEntryController.m
```diff
@@ -314,6 +300,73 @@ - (void)pinEntryControllerDidEnteredPin:(PEViewController *)controller
...
+    pinStage = PS_ENTER2;
+}
+
+- (void)errorAndResetWithMessage:(NSString *)errorMessage
+{
+    __weak PEPinEntryController *weakSelf = self;
+    [AlertViewPresenter.sharedInstance standardErrorWithMessage:errorMessage
+                                                          title:[LocalizationConstantsObjcBridge error]
+                                                        handler:^(UIAlertAction * _Nonnull action) {
+                                                            [weakSelf reset];
```
woudini: This isn't resetting the PIN for some reason

bchrisarriola:
Thanks for the thorough review.

For the first two points, the reason both of those are happening (which also happen on `dev`) is because `PEPinEntryController` is not actually part of the app's view controller navigation stack. Instead, its view is added as a view to the `rootViewController` (i.e. it is presented "modally"). The correct fix for this is to never add it as a subview and always use the UIViewController `present` method (was planning on addressing that in a separate PR).

Will update this PR with your third point 👍

woudini:
After enabling Touch ID, disabling it, then enabling it again, I'm getting a Settings View Controller on top of another Settings View Controller.

bchrisarriola:
Addressed all of the feedback from @woudini and tested all the different flows for pin entry (changing pin, enabling/disabling touch ID, etc.). Would appreciate your review here @achtenhagen when you get a chance. Since this PR is against `wallet_rearch` and not `dev`, it should be in a reviewable/mergeable state.

achtenhagen:
I think it's important to note here that the reviewer needs to run:
```
cd Submodules/
git submodule init
git submodule update
```
to install RxSwift and Alamofire

PR #248, opened 2018-06-14
fix(Addresses): remove space between asset selector and container vie…

woudini:
…w (except for iPhone X)

@achtenhagen there's still ~24 pt of extra space left at the top of the Addresses screen for the iPhone X, and possibly some space missing at the bottom too. All other phones look good so I assume it has to do with the safe areas.


achtenhagen:
@woudini Yes, that is likely. I can can have a look and fix this issue.

PR #257, opened 2018-06-15
iPhone X Support for About Us Screen

bchrisarriola:
Redoing the about us screen using storyboard + autolayout to support iPhone X.

iPhone X Screenshot:
simulator screen shot - iphone x - 2018-06-15 at 15 18 51

iPhone 5S Screenshot:
simulator screen shot - iphone 5s - 2018-06-15 at 15 18 59


Blockchain/Settings/Views/AboutUsViewController.swift
```diff
@@ -0,0 +1,56 @@
...
+
+@objc class AboutUsViewController: UIViewController {
+    @IBOutlet weak var buttonClose: UIButton!
+    @IBOutlet weak var labelMain: UILabel!
+    @IBOutlet weak var buttonRateUs: UIButton!
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        UIApplication.shared.statusBarStyle = .default
```
achtenhagen: When the controller is dismissed, this value should change back to `lightContent`.

bchrisarriola: Ah, good catch. We really should be using `preferredStatusBarStyle` but that's another larger task/issue to address.

PR #262, opened 2018-06-20
fix(tx_detail): make sure share action for an ether transaction directs to etherscan

bchrisarriola:
Tested and make sure that BTC, ETH, and BCH share action in the TX detail direct to the correct URL.


achtenhagen:
Getting the following error when trying to compile:
```
TransactionDetailNavigationController.m:160:28: Use of undeclared identifier 'AssetTypeLegacyHelper'
```

achtenhagen:
Required a full restart of Xcode and the clean, false alarm.

PR #263, opened 2018-06-20
fix(send): correctly handle different permutations of QR code addresses for all assets.

bchrisarriola:
Tested all the following cases:

BITCOIN:
* Should accept BTC address with prefix bitcoin:1K43HTP8ayuJjfqAHG7azwVDDQaDwLtqtK
* Should accept BTC address without prefix 1K43HTP8ayuJjfqAHG7azwVDDQaDwLtqtK
* Should not accept addresses with other currency prefix or address

BITCOIN CASH:
* Should accept BCH address with prefix bitcoincash:qpheqxjud2p0yg85t679tnc7n05xffs5ks07zhg6y6
* Should accept BCH address without prefix qpheqxjud2p0yg85t679tnc7n05xffs5ks07zhg6y6
* Should accept BCH legacy address with prefix bitcoincash:1K43HTP8ayuJjfqAHG7azwVDDQaDwLtqtK
* Should accept BCH legacy address without prefix 1K43HTP8ayuJjfqAHG7azwVDDQaDwLtqtK
* Should not accept addresses with other currency prefix

ETHER:
* Should accept ETH address without prefix 0x6584EF6C986d82d9B48683312cFb66234896bF2c
* Should not accept addresses with other currency prefix or address

### DIFF
--- a/Blockchain/LocalizationConstants.swift
+++ b/Blockchain/LocalizationConstants.swift
@@ -170,7 +170,10 @@ struct LocalizationConstants {
     }
 
     struct SendAsset {
-        static let invalidXAddressY = NSLocalizedString("Invalid %@ address: %@", comment: "")
+        static let invalidXAddressY = NSLocalizedString(
+            "Invalid %@ address: %@",
+            comment: "String presented to the user when they try to scan a QR code with an invalid address."
+        )
     }
 
     struct SendEther {
--- a/Blockchain/LocalizationConstants.swift
+++ b/Blockchain/LocalizationConstants.swift
@@ -169,6 +169,10 @@ struct LocalizationConstants {
         static let addWatchOnlyAddressWarningPrompt = NSLocalizedString("These options are recommended for advanced users only. Continue?", comment: "")
     }
 
+    struct SendAsset {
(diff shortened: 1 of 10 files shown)


Blockchain/LocalizationConstants.swift
```diff
@@ -169,6 +169,10 @@ struct LocalizationConstants {
         static let addWatchOnlyAddressWarningPrompt = NSLocalizedString("These options are recommended for advanced users only. Continue?", comment: "")
     }
 
+    struct SendAsset {
+        static let invalidXAddressY = NSLocalizedString("Invalid %@ address: %@", comment: "")
```
achtenhagen: Should we add comments for `NSLocalizedString` retroactively?

bchrisarriola: Yeah, let's do that. How about this plan?
- update all open PRs that modify `LocalizationConstants` and add comments to added/updated strings
- create a task to retroactively add comments to all the other strings

achtenhagen: Sounds good, but we could ignore the constants that have not yet been migrated from `LocalizationConstants.h`.

PR #264, opened 2018-06-20
fix(RawValued): fixing crash on release builds when using RawValued int enum

bchrisarriola:
Fixes crash on release builds when trying to first access `AppFeature`. 

Unclear why this is happening only on release builds but the operation `AppFeature.rawValues` would eventually cause a buffer overflow and crash the app (the `AnyIterator` inside the `rawValues` implementation will continue to iterate even though no further enums with such values exist). Also, this only happens on enums that are Ints (does not happen to other enums that are Strings). I wanted to fix this on the `RawValued` implementation but could not come up with a solution.

As a fix, I've defined a static `allFeatures` property. Once Swift 4.2 is supported, we can updated that enum and conform to `CaseIterable`.

### DIFF
--- a/Blockchain/Extensions/Enum+RawValued.swift
+++ b/Blockchain/Extensions/Enum+RawValued.swift
@@ -8,6 +8,10 @@
 
 import Foundation
 
+/// Protocol definition for an enum that is also a `RawValue`.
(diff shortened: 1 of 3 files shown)


bchrisarriola:
Will update this PR to add comments to RawValued about the Int issue

PR #267, opened 2018-06-21
UIView Safe Area Frame Extension

achtenhagen:
Highlights in this PR:
- Adds convenient extension to get a view's frame based on the safe area of the root view.


Blockchain/UIView+SafeAreaFrame.swift
```diff
@@ -0,0 +1,67 @@
...
+
+import Foundation
+
+/**
+ UIView Safe Area Frame Extension
+ - Note: must only be called when safe area insets have been set (viewSafeAreaInsetsDidChange(), viewDidLayoutSubviews())
+ - Note: this extension only supports the default screen orientation.
+ - Parameter navigationBar: the height of the navigation bar should be considered when sizing the frame.
+ - Parameter tabBar: the height of the tab bar should be considered when sizing the frame.
+ - Parameter: assetSelector: the height of the asset selector should be considered when sizing the frame.
```
bchrisarriola: Seems odd to leave parameter name comments as part of the extension's documentation. Move this above the method definition `class func safeAreaFrame`?

Blockchain/UIView+SafeAreaFrame.swift
```diff
@@ -0,0 +1,67 @@
...
+ */
+@objc
+extension UIView {
+    class func safeAreaFrame(
+        navigationBar: Bool = false,
+        tabBar: Bool = false,
+        assetSelector: Bool = false) -> CGRect {
```
bchrisarriola: Nit:

I think adding a new line right before the closing parens really aids in readability. So:

```
class func safeAreaFrame(
        navigationBar: Bool = false,
        tabBar: Bool = false,
        assetSelector: Bool = false
) -> CGRect {
```

Blockchain/UIView+SafeAreaFrame.swift
```diff
@@ -0,0 +1,67 @@
...
+/**
+ UIView Safe Area Frame Extension
+ - Note: must only be called when safe area insets have been set (viewSafeAreaInsetsDidChange(), viewDidLayoutSubviews())
+ - Note: this extension only supports the default screen orientation.
+ - Parameter navigationBar: the height of the navigation bar should be considered when sizing the frame.
+ - Parameter tabBar: the height of the tab bar should be considered when sizing the frame.
+ - Parameter: assetSelector: the height of the asset selector should be considered when sizing the frame.
+ */
+@objc
+extension UIView {
```
bchrisarriola: I think we want this to be a `UIViewController` extension and not a `UIView` extension. Currently this assumes that the `UIView` requesting the `safeAreaFrame` or `safeAreaInsets` is inside the app's rootViewController which might not always be the case.

Blockchain/UIView+SafeAreaFrame.swift
achtenhagen: You're right about the readability, but when I press `CTRL + i`, I expect the code syntax and indentation to be adjusted to what Xcode prefers, so in that case your above suggestion would look like:
```
    class func safeAreaFrame(
        navigationBar: Bool = false,
        tabBar: Bool = false,
        assetSelector: Bool = false
        ) -> CGRect {
```

Will change the method signature to that.

bchrisarriola: Yeah I noticed that too which bugs the hell out of me. I'm sure that's something we can configure

Blockchain/UIView+SafeAreaFrame.swift
achtenhagen: I made it an extension of `UIView` because the `safeAreaLayoutGuide`, which gives us `layoutFrame`, is a property of `UIView`, not `UIViewController`.
In my opinion, `UIViewController` should only have methods that act upon `UIView` and not on properties of it's children. Thoughts?

bchrisarriola: Makes sense.

Sorry just thought of another adjustment—thoughts about renaming these methods to be prefixed with `rootView*` so that it's clear that these functions operate on `rootViewController.view`?

achtenhagen: Sure, sounds good.

PR #277, opened 2018-06-22
fix(Receive): prevent BTC payment received amount from showing twice.

bchrisarriola:
Fixes the issue wherein the alert shown when a payment is received displays duplicate payment amounts in BTC.

Ideally the fiat amount right after the BTC amount should be displayed (which happens most of the time). However, there is a race condition wherein we don't have the user's local currency's conversion value from/to BTC amount. When this happens, we'll just display the BTC amount and leave the fiat amount off. Also weird, when the local currency conversion is not available, the `NSNumberFormatter:formatMoney:localCurrency` method will fallback to _always_ just convert the amount to BTC and completely ignore the `localCurrency` flag. Fixing this method may have some unintended consequences since it's used in a lot of other places so I figured this patch will do for now.

Obviously this isn't ideal but the correct fix should be addressed in our wallet rearch efforts.

### DIFF
--- a/Blockchain/ReceiveBitcoinViewController.h
+++ b/Blockchain/ReceiveBitcoinViewController.h
@@ -46,8 +46,6 @@
 
 @property(nonatomic) UIView *bottomContainerView;
 @property(nonatomic) UILabel *receiveToLabel;
-@property(nonatomic) UIButton *selectFromButton;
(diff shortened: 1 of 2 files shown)


bchrisarriola:
See #278

PR #282, opened 2018-06-25
fix(Settings): show correct biometric type in settings

achtenhagen:
Highlights in this PR:
- Now shows the correct biometric authentication type in the **Settings** view.
- Now shows the correct biometric authentication type in the **Wallet Setup** view.
- Standardizes the naming of objects related to biometric authentication (to be more agnostic).
- Correctly sizes header labels for table view sections.

### IMPORTANT
Two keys in the UserDefaults have been renamed to reflect the object name standardization and thus will **disable** biometric authentication the **first time** the user returns to the app after updating the application.

Some localized constants were also affected and thus translations need to be updated before merging this PR.

As mentioned in the code docs, we should not have a toggle to opt-in to using biometric authentication, but rather prompt the user with the system dialog during the pin creation step. To be discussed further.

Biometric authentication-related configuration should **not** be stored by the application, but rather by the system itself.

For more info, please see:
https://developer.apple.com/design/human-interface-guidelines/ios/user-interaction/authentication

### DIFF
--- a/Blockchain/Extensions/UserDefaults.swift
+++ b/Blockchain/Extensions/UserDefaults.swift
@@ -21,7 +21,7 @@ extension UserDefaults {
 
     enum Keys: String {
         case assetType = "assetType"
-        case didFailTouchIDSetup = "didFailTouchIDSetup"
+        case didFailBiometrySetup = "didFailBiometrySetup"
         case encryptedPinPassword = "encryptedPINPassword"
         case environment = "environment"
         case firstRun = "firstRun"
@@ -36,10 +36,10 @@ extension UserDefaults {
         case reminderModalDate = "reminderModalDate"
         case shouldHideAllCards = "shouldHideAllCards"
         case shouldHideBuySellCard = "shouldHideBuySellNotificationCard"
-        case shouldShowTouchIDSetup = "shouldShowTouchIDSetup"
+        case shouldShowBiometrySetup = "shouldShowBiometrySetup"
         case swipeToReceiveEnabled = "swipeToReceive"
         case symbolLocal = "symbolLocal"
-        case touchIDEnabled = "touchIDEnabled"
+        case biometryEnabled = "biometryEnabled"
         case hideTransferAllFundsAlert = "hideTransferAllFundsAlert"
         case defaultAccountLabelledAddressesCount = "defaultAccountLabelledAddressesCount"
     }
--- a/Blockchain/Extensions/UserDefaults.swift
+++ b/Blockchain/Extensions/UserDefaults.swift
@@ -1,5 +1,5 @@
 //
-//  Userstandard.swift
+//  UserDefaults.swift
 //  Blockchain
 //
 //  Created by Maurice A. on 4/13/18.
--- a/Blockchain/Extensions/UserDefaults.swift
+++ b/Blockchain/Extensions/UserDefaults.swift
@@ -44,18 +44,15 @@ extension UserDefaults {
         case defaultAccountLabelledAddressesCount = "defaultAccountLabelledAddressesCount"
     }
 
-    static func migrateLegacyKeysIfNeeded() {
-        if let didFailTouchIDSetup = standard.object(forKey: "didFailTouchIDSetup") as? Bool {
-            standard.set(didFailTouchIDSetup, forKey: Keys.didFailBiometrySetup.rawValue)
-            standard.removeObject(forKey: "didFailTouchIDSetup")
-        }
-        if let shouldShowTouchIDSetup = standard.object(forKey: "shouldShowTouchIDSetup") as? Bool {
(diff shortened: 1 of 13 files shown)


Blockchain/LocalizationConstants.swift
```diff
@@ -115,6 +115,8 @@ struct LocalizationConstants {
 
         //: Deprecated Authentication Errors (remove once we stop supporting iOS >= 8.0 and iOS <= 11)
         static let touchIDLockout = NSLocalizedString("Unable to Authenticate because there were too many failed Touch ID attempts. Passcode is required to unlock Touch ID", comment: "")
+        static let biometryWarning = NSLocalizedString("Enabling this feature will allow all users with a registered %@ fingerprint on this device to access to your wallet.", comment: "%@ represents either Face ID or Touch ID to avoid additional translation")
+        static let enableBiometrics = NSLocalizedString("Enable %@", comment: "")
```
bchrisarriola: Let's start adding comments to `NSLocalizedString` to provide more context to the translator where the string is used as that helps provide a more accurate translation.

e.g.

```
static let enableBiometrics = NSLocalizedString(
    "Enable %@", 
    comment: "Question asked to user if they want to enable biometrics for authenticating into the app."
)
```

bchrisarriola: Also for strings that take in an argument, it would be helpful to know that by indicating in the variable name. So, `enableBiometrics` might be better renamed as `enableX` or `enableArgument` (I personally like the former naming convention).

Blockchain/Extensions/UserDefaults.swift
```diff
@@ -21,7 +21,7 @@ extension UserDefaults {
 
     enum Keys: String {
         case assetType = "assetType"
-        case didFailTouchIDSetup = "didFailTouchIDSetup"
+        case didFailBiometrySetup = "didFailBiometrySetup"
```
bchrisarriola: Since these keys are renamed, migration needs to be handled correctly since it is a bad experience to have to re-enable biometrics if you had enabled it in the past. One way to handle this is:
1. override the BlockchainSettings.App initializer and create a method `handleMigrationIfNeeded`
2. in `handleMigrationIfNeeded`, for each renamed key:
   - check if the key exists
   - if so, set the value for the new key
   - delete the key
(I think 2 can be an extension function of `UserDefault`)

bchrisarriola:
Just tested this on iPhone X and 8 and verified that touch ID & face ID work as expected—good work on this 👍

Blockchain/Extensions/UserDefaults.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  UserDefaults.swift
+//  Userstandard.swift
```
bchrisarriola: Revert?

Blockchain/Extensions/UserDefaults.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  UserDefaults.swift
```
achtenhagen: Mistake created by Find & replace. Will fix 😓

Blockchain/Extensions/UserDefaults.swift
```diff
@@ -36,11 +36,26 @@ extension UserDefaults {
...
+        case shouldShowBiometrySetup = "shouldShowBiometrySetup"
         case swipeToReceiveEnabled = "swipeToReceive"
         case symbolLocal = "symbolLocal"
-        case touchIDEnabled = "touchIDEnabled"
+        case biometryEnabled = "biometryEnabled"
         case hideTransferAllFundsAlert = "hideTransferAllFundsAlert"
         case defaultAccountLabelledAddressesCount = "defaultAccountLabelledAddressesCount"
     }
+
+    static func migrateLegacyKeysIfNeeded() {
```
bchrisarriola: This shouldn't be a static function—should just be an instance method and all `standard` references can be replaced by `self`.

Blockchain/Extensions/UserDefaults.swift
```diff
@@ -36,11 +36,26 @@ extension UserDefaults {
...
+        if let didFailTouchIDSetup = standard.object(forKey: "didFailTouchIDSetup") as? Bool {
+            standard.set(didFailTouchIDSetup, forKey: Keys.didFailBiometrySetup.rawValue)
+            standard.removeObject(forKey: "didFailTouchIDSetup")
+        }
```
bchrisarriola: You can create a private instance method helper to remove duplication here.

e.g.

```
private func migrateBool(fromKey: String, toKey: String) {
    guard let value = self.object(forKey: fromKey) as? Bool else { return }
    self.set(value, forKey: toKey)
    self.removeObject(forKey: fromKey)
}
```

achtenhagen: Nice, that makes things much cleaner!

PR #286, opened 2018-06-27
fix(Send): clear amount fields when view disappears

achtenhagen:
Highlights in this PR:
- In the **Send** and **Request** views, amount fields are now cleared when navigating away from them.
- The **Send** view also resets the **To** field and the **Use total available minus fee** label.
- Removed more references to the deprecated **Contacts** feature.

### DIFF
--- a/Blockchain/ReceiveBitcoinViewController.m
+++ b/Blockchain/ReceiveBitcoinViewController.m
@@ -115,6 +115,11 @@
     [self updateUI];
 }
 
+- (void)viewDidDisappear:(BOOL)animated
+{
+    [self clearAmounts];
+}
+
 - (void)viewWillDisappear:(BOOL)animated
 {
     [super viewWillDisappear:animated];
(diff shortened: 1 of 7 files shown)


Blockchain/ReceiveBitcoinViewController.m
```diff
@@ -115,6 +115,11 @@ - (void)viewDidLoad
     [self updateUI];
 }
 
+- (void)viewDidDisappear:(BOOL)animated
+{
+    [self clearAmounts];
```
bchrisarriola: Need to call `[super viewDidDisappear:animated]`

Blockchain/SendEtherViewController.m
```diff
@@ -174,6 +174,11 @@ - (void)viewWillAppear:(BOOL)animated
     [self getHistory];
 }
 
+- (void)viewDidDisappear:(BOOL)animated
+{
+    [self reload];
```
bchrisarriola: Same here

Blockchain/ReceiveBitcoinViewController.m
achtenhagen: I wonder why the autocompletion of this method does not insert it for you if it should always be called. Will fix.

PR #293, opened 2018-06-29
refactor(App Review): extract app review logic into separate component

achtenhagen:
Highlights in this PR:
- Extracts the app review logic into a separate component.

### DIFF
--- a/Blockchain/AppDelegate.swift
+++ b/Blockchain/AppDelegate.swift
@@ -40,8 +40,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         Fabric.with([Crashlytics.self])
 
-        BlockchainSettings.App.shared.appOpenedCount += 1
-
         // MARK: - Global Appearance
 
         //: Status Bar
@@ -157,6 +155,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func applicationWillEnterForeground(_ application: UIApplication) {
         print("applicationWillEnterForeground")
+
+        BlockchainSettings.App.shared.appOpenedCount += 1
 
         BuySellCoordinator.shared.start()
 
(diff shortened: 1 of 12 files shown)


Blockchain/AppDelegate.swift
```diff
@@ -40,6 +40,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         Fabric.with([Crashlytics.self])
 
+        BlockchainSettings.App.shared.appOpenedCount += 1
```
bchrisarriola: This will only count fresh launches. Seems like we may want to increment this in `applicationWillEnterForeground` instead.

Blockchain/AppReviewPrompt/AppReviewPrompt.swift
```diff
@@ -0,0 +1,87 @@
...
+            print("App review prompt will not show because the application open count is too low (\(count)).")
+            #endif
+            return
+        }
+    }
+
+    private func requestReview() {
+        if #available(iOS 10.3, *) {
+            SKStoreReviewController.requestReview()
+        } else {
```
bchrisarriola: Nit: for readability, this can be improved by early returning inside the if statement and remove the else branch:

i.e.

```
if #available(iOS 10.3, *) {
    SKStoreReviewController.requestReview()
    return
}

// ... pre iOS 10.3 handling goes here
```

...alternatively, you can wrap the legacy handling in another method

i.e.

```
if #available(iOS 10.3, *) {
    SKStoreReviewController.requestReview()
} else {
    requestReviewLegacy()
}
```

achtenhagen: Yeah I usually do that, but missed it because the snippet for the `#available` check inserts if/else. Will update.

Blockchain/AppDelegate.swift
achtenhagen: Good idea, makes sense. Will update.

PR #294, opened 2018-06-29
chore(WalletOptions): Check for site maintenance before validating pin

bchrisarriola:
Summary:
* Removed `getWalletOptions` and `checkForMaintenance` from `NetworkManager`
* Added `walletOptions` property to `WalletService` (the old checking for maintenance call is now replaced by inspecting `walletOptions` directly)
* Added test that an error is presented when the site is down for maintenance


Blockchain/Network/Models/WalletOptions.swift
```diff
@@ -8,53 +8,62 @@
...
-    let mobileInfo: MobileInfo?
 }
 
 extension WalletOptions {
-    init(response: [String: Any]) {
-        downForMaintenance = response[Keys.maintenance] as? Bool ?? false
-        self.mobile = WalletOptions.Mobile(response)
-        self.mobileInfo = WalletOptions.MobileInfo(response)
+    init(json: JSON) {
+        downForMaintenance = json[Keys.maintenance] as? Bool ?? false
```
woudini: I know I originally wrote this, but what do you think about making this `self.downForMaintenance` or removing `self` from these assignments?

PinEntry/Classes/PEPinEntryController+PinView.swift
```diff
@@ -20,8 +21,26 @@ extension PEPinEntryController: PinView {
...
+        // Check for maintenance first, followed by validating the user's pin
+        showLoadingView(withText: LocalizationConstants.verifying)
+
+        _ = WalletService.shared.walletOptions
+            .observeOn(MainScheduler.instance)
+            .subscribe(onSuccess: { [weak self] walletOptions in
+                guard let strongSelf = self else { return }
+                guard !walletOptions.downForMaintenance else {
+                    strongSelf.hideLoadingView()
+                    strongSelf.error(message: LocalizationConstants.Errors.siteMaintenanceError)
```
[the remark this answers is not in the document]
bchrisarriola: Ah, yeah good catch

Blockchain/Network/Models/WalletOptions.swift
bchrisarriola: I like directly using `self` in the initializer for instance members as it makes it super clear that you're initializing instance members. Will update this to use `self`

woudini:
Any idea why codebeat is showing changes introduced to `DebugTableViewController`?

PR #296, opened 2018-07-02
fix(Wallet Setup): do not show biometry setup view if the device is n…

achtenhagen:
Highlights in this PR:
- Fixes the issue where the biometrics setup view would be presented even if the device itself was not enrolled in biometric authentication.


achtenhagen:
The `UIDevice` extension sounds like a good place for that. I can add those changes into this PR. As a side note, we should not store an enabled value for biometrics and instead let the system tell the app if it is or not. I think we have a boolean in UserDefaults which violates this.

bchrisarriola:
👍 another thing we can refactor

Blockchain/Extensions/UIDevice+Biometrics.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    @available(iOS 11.0, *)
+    init?(type: LABiometryType) {
+        switch type {
+        case .faceID:
+            self.title = "Face ID"
+            self.asset = "IconFaceID"
+        case .touchID:
+            self.title = "Touch ID"
+            self.asset = "IconTouchID"
```
bchrisarriola: Call `self.init()` here to avoid duplication?

achtenhagen: This is unfortunately not possible and creating a helper function results in `Property not initialized at super.init call`. I would suggest keeping this duplication for now.

achtenhagen: This is a great solution and it works just as intended. Nice!

PR #302, opened 2018-07-03
fix(Pin): Do not clear pin in keychain when toggling biometric auth.

bchrisarriola:
Fixes issue wherein biometric authentication fails when you disable and re-enable it in the settings view. The reason that was failing was because we were clearing the pin that is stored in `BlockchainSettings.App` when that setting is toggled. Fix is to not clear the pin when toggling.

Verified that this works by creating a new wallet, enabling biometric auth in the new wallet set-up flow, toggling biometric auth in settings. Upon backgrounding and reforeground the app, biometric auth still works in the pin screen 👍

### DIFF
--- a/Blockchain/Pin/PinInteractor.swift
+++ b/Blockchain/Pin/PinInteractor.swift
@@ -31,12 +31,33 @@ import RxSwift
     /// - Returns: an Observable returning the response
     func validatePin(_ pinPayload: PinPayload) -> Single<GetPinResponse> {
         return self.walletService.validatePin(pinPayload)
-            .do(onSuccess: { response in
-                // Clear pin from keychain if the user exceeded the number of retries
-                // when entering the pin.
-                if response.statusCode == .deleted {
+            .do(onSuccess: { [unowned self] response in
+                guard let responseCode = response.statusCode else { return }
+
+                switch responseCode {
+                case .success:
+                    // Optionally save the pin to the keychain
+                    self.saveToKeychainIfNeeded(pinPayload: pinPayload)
+                    return
+                case .deleted:
+                    // Clear pin from keychain if the user exceeded the number of retries when entering the pin.
                     BlockchainSettings.App.shared.pin = nil
+                    return
+                default:
+                    return
                 }
             })
+    }
+
+    private func saveToKeychainIfNeeded(pinPayload: PinPayload) {
+        guard let config = AppFeatureConfigurator.shared.configuration(for: .biometry),
+            config.isEnabled,
+            BlockchainSettings.App.shared.biometryEnabled else {
+            // Biometric auth not enabled, not saving pin to keychain
+            return
+        }
+
+        let pin = Pin(string: pinPayload.pinCode)
+        pin?.saveToKeychain()
     }
 }
--- a/Blockchain/Pin/PinInteractor.swift
+++ b/Blockchain/Pin/PinInteractor.swift
@@ -32,9 +32,21 @@ import RxSwift
     func validatePin(_ pinPayload: PinPayload) -> Single<GetPinResponse> {
         return self.walletService.validatePin(pinPayload)
             .do(onSuccess: { response in
-                // Clear pin from keychain if the user exceeded the number of retries when entering the pin.
-                if response.statusCode == .deleted {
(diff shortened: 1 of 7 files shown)


bchrisarriola:
For some history on this issue, pretty sure this was introduced by #220 (see discussion there) wherein I added the change that we do not update the pin every single time the pin code is validated. Instead, the pin code in the keychain is only updated when the user first creates a pin, or when they update the pin.

bchrisarriola:
Hm yeah the scenario that you are mentioning unfortunately necessitates setting the pin in the keychain upon validating their pin. 

I think I'll update this PR with the following:
* set the pin in the keychain upon validating their pin to capture the use case that you mentioned
* add comments on why that needs to be done... eventually we should remove that though (perhaps after this upcoming version of the app has been released and many users are already on it)

Blockchain/Pin/PinInteractor.swift
```diff
@@ -31,12 +31,33 @@ import RxSwift
...
                     BlockchainSettings.App.shared.pin = nil
+                    return
+                default:
+                    return
                 }
             })
     }
+
+    private func saveToKeychainIfNeeded(pinPayload: PinPayload) {
+        guard let config = AppFeatureConfigurator.shared.configuration(for: .biometry),
```
woudini: Would it make sense to add an extension to `PinPayload` allowing it to take a `Bool` `shouldSavePinCode`, and pass `self?.verifyOptional` into it here? https://github.com/blockchain/My-Wallet-V3-iOS/blob/auth_fix/PinEntry/Classes/PEPinEntryController+PinView.swift#L40

woudini: That way you could remove this guard, rename this function to just `saveToKeychain` and call it if `pinPayload.shouldSavePinCode` is true (or remove  `saveToKeychainIfNeeded`)

Blockchain/SettingsTableViewController.m
```diff
@@ -632,7 +632,6 @@ - (void)toggleBiometry
         }]];
         [self presentViewController:alertForTogglingBiometry animated:YES completion:nil];
     } else {
-        BlockchainSettings.sharedAppInstance.pin = nil;
```
woudini: Can this be added back now that the plan is to save it when the user enables it?

bchrisarriola: Yeah, let me add it back

Blockchain/Pin/PinInteractor.swift
bchrisarriola: I'll move it as a method of `Pin`. I'd prefer to keep `PinPayload` as a "dumb" object.

bchrisarriola: But this code doesn't need to check for the value of `verifyOptional`, it would always save the pin to the keychain if biometric auth is enabled upon pin validation.

woudini: `BlockchainSettings.App.shared.biometryEnabled` isn't set to true until after the `guard` so the pin still isn't saved (https://github.com/blockchain/My-Wallet-V3-iOS/blob/auth_fix/Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift#L50). By using `verifyOptional` there isn't a need to remember any state because it will in every case be for enabling biometric auth.

bchrisarriola: I was trying to avoid adding logic like that in `AuthenticationCoordinator` but it seems like it unavoidable without refactoring a lot more. Committed fe819c4b w/ your feedback. I tested that on various cases (enabling/disabling touch ID, changing pin, etc.) and verified that all still works.

woudini: `lastEnteredPIN` is nil for me here https://github.com/blockchain/My-Wallet-V3-iOS/pull/302/files#diff-0e82b0260dd43a9782f10c719d2a0181R51

bchrisarriola: Sticking with the original design of caching the pin to the keychain in `PinInteractor`. Change is in 0e0ada6e

woudini:
Renaming merge commit

PR #305, opened 2018-07-03
refactor(OnboardingSettings): Moving onboarding-related preferences to OnboardingSettings

bchrisarriola:
`BlockchainSettings.App` is turning into a blob where all `NSUserDefaults` key values are stored. To prevent/avoid this, in this PR I moved all onboarding-specific key/values that are stored in `NSUserDefaults` into a more specific component called `OnboardingSettings`.


bchrisarriola:
Good point. In the spirit of being consistent, let me update this and move `OnboardingSettings` to a struct with `BlockchainSettings` i.e. BlockchainSettings.Onboarding`. This is also consistent with how we handle LocalizedStrings and Constants.

PR #307, opened 2018-07-05
 fix(Send): show missing fields on confirmation screen

achtenhagen:
Highlights in this PR:
- Shows the missing fields on the ethereum payment confirmation screen.

Tested with different local currency symbols to ensure it works.


Blockchain/BCConfirmPaymentViewModel.m
```diff
@@ -46,10 +46,12 @@ - (id)initWithTo:(NSString *)to
        fiatTotal:(NSString *)fiatTotal
 {
     if (self == [super init]) {
+        self.from = BC_STRING_MY_ETHER_WALLET;
```
bchrisarriola: Move this string to `LocalizatonConstants`?

PR #308, opened 2018-07-05
feat(WatchOnly): add non-spendable views and labels

woudini:
- Format watch only label and change label text in Transaction Cell to "Non-Spendable"
- When watch-only addresses are present, expand Balance Chart View, scroll view content size, and shift price views down
- Increase width and change label text in Addresses and Address Selection View to "Non-Spendable"


Blockchain/TransactionTableCell.m
```diff
@@ -56,29 +57,46 @@ - (void)reload
...
+        infoLabel.layer.borderColor = [exchangeBorderColor CGColor];
     } else {
         infoLabel.hidden = YES;
         actionLabel.frame = CGRectMake(actionLabel.frame.origin.x, 29, actionLabel.frame.size.width, actionLabel.frame.size.height);
         dateLabel.frame = CGRectMake(dateLabel.frame.origin.x, 11, dateLabel.frame.size.width, dateLabel.frame.size.height);
     }
     
     [infoLabel sizeToFit];
+    // Having issues vertically centering the text - I'm sure there is a better way to do this but it might take too long to do
+    [infoLabel changeHeight:infoLabel.frame.size.height + 3];
```
woudini: For reviewers, I am planning to fix/change this

Blockchain/WatchOnlyBalanceView.swift
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    @objc func updateText(balance: String) {
+        balanceLabel.text = balance + " " + LocalizationConstants.AddressAndKeyImport.nonSpendable
+        balanceLabel.sizeToFit()
+        let horizontalOffset = self.balanceLabel.customEdgeInsets.left + self.balanceLabel.customEdgeInsets.right
```
woudini: Will add comments here explaining need for offsets

Blockchain/BalanceDisplayModel.swift
```diff
@@ -0,0 +1,26 @@
...
+//  BalanceDisplayModel.swift
+//  Blockchain
+//
+//  Created by kevinwu on 7/4/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+@objc class BalanceModel: NSObject {
```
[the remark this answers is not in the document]
bchrisarriola: Actually this might not just be dashboard-specific. Could be something we can reuse elsewhere where balance needs to be displayed and can be toggled between fiat/crypto value.

Blockchain/BCBalancesChartView.m
```diff
@@ -92,6 +96,50 @@ - (void)setupLegendWithFrame:(CGRect)frame
...
+{
+    [self changeHeight:self.defaultHeight + [self watchOnlyViewHeight]];
+
+    if (!self.watchOnlyBalanceView) {
+        CGFloat containerViewHorizonalPadding = CONTAINER_VIEW_HORIZONTAL_PADDING;
+        self.watchOnlyBalanceView = [[WatchOnlyBalanceView alloc] initWithFrame:CGRectMake(containerViewHorizonalPadding, self.legendKeyContainerView.frame.origin.y + self.legendKeyContainerView.frame.size.height + 8, self.legendKeyContainerView.frame.size.width, 40)];
+        [self addSubview:self.watchOnlyBalanceView];
+    }
+
+    [self.watchOnlyBalanceView updateTextWithBalance:@"test"];
```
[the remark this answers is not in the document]
woudini: This part is WIP

Blockchain/BalanceDisplayModel.swift
```diff
@@ -0,0 +1,26 @@
...
+    @objc var balance: String?
+    @objc var fiatBalance: Double = 0
+}
+
+@objc class BalanceDisplayModel: BalanceModel {
+
+    @objc var watchOnly: WatchOnlyDisplayModel
+
+    override init() {
+        self.watchOnly = WatchOnlyDisplayModel()
```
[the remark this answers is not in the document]
bchrisarriola: Also, it might not be necessary for `BalanceDisplayModel` to inherit from `BalanceModel`. Instead, it can have two properties:

```
@objc class BalanceDisplayViewModel: NSObject {
    @objc let nonSpendableBalance: BalanceModel = BalanceModel()

    @objc let spendableBalance: BalanceModel = BalanceModel()
}
```

Blockchain/TransactionTableCell.m
woudini: https://github.com/blockchain/My-Wallet-V3-iOS/pull/308/files#diff-423e542c23374f9617b03c4fc7ddf6f5R22

Blockchain/WatchOnlyBalanceView.swift
woudini: See https://github.com/blockchain/My-Wallet-V3-iOS/pull/308#discussion_r200744770

Blockchain/BalanceDisplayModel.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Blockchain
+//
+//  Created by kevinwu on 7/4/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+@objc class BalanceModel: NSObject {
+    @objc var balance: String?
```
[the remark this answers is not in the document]
woudini: I wanted to minimize the string formatting required to use this model in the balance chart view. The pie chart requires `double` fiat values so I wanted to keep `fiatBalance` as a `double` for that and display them in places outside of the pie chart with `fiatStringFromDouble:`. As for `balance`, I wanted it to already be formatted for displaying since the numeric data structure can differ from BTC/BCH (which uses integers in My-Wallet-V3) and ETH (which uses decimal numbers in My-Wallet-V3).

Blockchain/BalanceDisplayModel.swift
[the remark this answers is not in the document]
woudini: It's a bit too specific for being reused because of the `double` requirement for the pie chart (https://github.com/blockchain/My-Wallet-V3-iOS/pull/308#discussion_r200762458). Will rename to `BalancesChartViewModel`

Blockchain/BalanceChartModel.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+//  Created by kevinwu on 7/4/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+@objc class BalanceChartModel: NSObject {
+    private var _balance: String?
+    @objc var balance: String? {
```
bchrisarriola: Looks like `balance` will never be nil when retrieving it based on the `get` logic below so it would be better if this is a non-optional which makes it nice from the caller's perspective so that you don't need to unwrap `balance`.

What about creating another property named `balanceDisplayText`? So:

```
@objc class BalanceChartModel: NSObject {
    @objc var balance: String?

    @objc var fiatBalance: Double = 0

    @objc var balanceDisplayText: String {
        return balance ?? "0"
    }
}
```

...so whatever UIView needs to display the balance, it will use `balanceDisplayText` instead

Blockchain/BCBalancesChartView.m
```diff
@@ -127,31 +228,31 @@ - (void)updateTotalFiatBalance:(NSString *)fiatBalance
 - (void)updateBitcoinBalance:(NSString *)balance
 {
     if (!balance) {
```
bchrisarriola: This whole method can be simplified since `BalanceChartModel`'s `balanaceDisplayText` will default to "0" if the balance is nil.

```
self.bitcoin.balance = balance
```

Blockchain/BCBalancesChartView.m
```diff
@@ -127,31 +228,31 @@ - (void)updateTotalFiatBalance:(NSString *)fiatBalance
...
 
 - (void)updateEtherBalance:(NSString *)balance
 {
-    self.etherBalance = balance;
+    self.ether.balance = balance;
 }
 
 - (void)updateBitcoinCashBalance:(NSString *)balance
 {
     if (!balance) {
```
bchrisarriola: This can be simplified as well

Blockchain/BalanceChartModel.swift
woudini: The tricky thing about this (and what I wanted to open another PR for) is that in the watch-only section you want `nil` to show 0, and in the regular balances you don't (like when the wallet is loading, in the image below). What I ended up trying was simplifying the model and handling the special case in the watch-only section.
simulator screen shot - iphone 6 - 2018-07-06 at 18 00 33

woudini:
Pending QA tomorrow - merging.

PR #315, opened 2018-07-09
fix(Payment Received): fix payment received alert not shown on pin screen

bchrisarriola:
Looks like it wasn't being presented because of the change wherein the alert was trying to be presented from `TabControllerManager` which is not in the foreground. Setting the "in:" parameter to `nil` will use the `topMostViewController` so that the alert is presented correctly.

These are the sort of side-effects that having a single delegate on the `Wallet` create. Once we get to refactoring websockets out of the `Wallet`, we'll have a more robust solution.

### DIFF
--- a/Blockchain/TabControllerManager.m
+++ b/Blockchain/TabControllerManager.m
@@ -74,7 +74,7 @@
 
 - (void)onPaymentReceivedWithAmount:(NSString * _Nonnull)amount assetType:(enum AssetType)assetType
 {
-    [AlertViewPresenter.sharedInstance standardNotifyWithMessage:amount title:BC_STRING_PAYMENT_RECEIVED in:self handler:nil];
+    [AlertViewPresenter.sharedInstance standardNotifyWithMessage:amount title:BC_STRING_PAYMENT_RECEIVED in:nil handler:nil];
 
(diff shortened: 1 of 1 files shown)


bchrisarriola:
> utACK, websockets currently not receiving messages.

Is that a back-end issue?

woudini:
I originally created `TabControllerManager` as an `NSObject`. I see now that it's a `UINavigationController` that sets its `viewControllers` to `@[self.tabViewController]`. @achtenhagen any thoughts on why this `TabControllerManager` isn't able to present an alert? Also is there a particular reason `TabControllerManager` could not stay as an `NSObject`?

woudini:
Yes, most likely a backend issue since logs show that the `SRWebSocket`s are being opened.

bchrisarriola:
>  any thoughts on why this TabControllerManager isn't able to present an alert

For this specific issue, I added support in `AlertViewPresenter` (ba8b29ec) wherein the caller can be explicit as to where it wants to present an alert (i.e. you can pass in a non-nill value for the "in" parameter). The reason this change in this PR is necessary is because `TabControllerManager` was trying to present an alert even though it is not in the foreground (ie the pin screen is in the foreground).

PR #316, opened 2018-07-09
fix(Pin): Make sure pin input is properly reseted when an alert is presented

bchrisarriola:
Fixes the issue wherein the circles in the pin entry view sometimes don't get cleared when an alert is presented.


woudini:
Any examples of where this was happening?

bchrisarriola:
Yep. This was happening when you either set a new pin or change your pin, enter a common pin (e.g. "1111"), and tap "try again" on the alert that gets presented.

PR #317, opened 2018-07-09
fix(Pin): make sure user is prompted to create new pin upon entering password

bchrisarriola:
Fixes the issue wherein when the user is prompted to enter their password after incorrectly entering their pin 4 times, they aren't asked to create a new pin.

### DIFF
--- a/Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
+++ b/Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
@@ -119,9 +119,9 @@ extension AuthenticationCoordinator: WalletPinEntryDelegate {
     func errorDidFailPutPin(errorMessage: String) {
         LoadingViewPresenter.shared.hideBusyView()
 
-        AlertViewPresenter.shared.standardError(message: errorMessage)
-
-        reopenChangePin()
+        AlertViewPresenter.shared.standardError(message: errorMessage) { [unowned self] _ in
+            self.reopenChangePin()
+        }
     }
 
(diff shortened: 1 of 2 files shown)


Blockchain/Authentication/Models/AuthenticationCoordinator+Pin.swift
```diff
@@ -119,9 +119,9 @@ extension AuthenticationCoordinator: WalletPinEntryDelegate {
     func errorDidFailPutPin(errorMessage: String) {
         LoadingViewPresenter.shared.hideBusyView()
 
-        AlertViewPresenter.shared.standardError(message: errorMessage)
-
-        reopenChangePin()
+        AlertViewPresenter.shared.standardError(message: errorMessage) { [unowned self] _ in
+            self.reopenChangePin()
```
woudini: Does this change anything? It doesn't look like `WalletPinEntryDelegate` is used anymore. For the sake of the build it doesn't have to be deprecated in this PR.

bchrisarriola: `WalletPinEntryDelegate` is actually used and this delegate method is still invoked when the user changes their pin (i.e. that network call still goes through `Wallet`). This should be refactored out in favor of using `PinInteractor` and other associated components.

This change is necessary though because you won't ever see the presented alert because the alert tries to be presented in `PEViewController` but fails to be presented because that view controller gets dismissed shortly after. Fix is to wait for the alert to be dismissed before calling `reopenChangePin`.

PR #319, opened 2018-07-10
fix(Project): allow building without Fabric keys

woudini:
Also updates README build instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,37 @@ Prepare OpenSSL:
     cd ../OpenSSL-for-iPhone
     ./build-libssl.sh
 
+Add production Config file
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,6 +11,13 @@ Prepare SocketRocket:
 
 Prepare the MyWallet Javascript:
 
+    # use node v7.9.0 by using a node version manager such as NVM or n
```
bchrisarriola: Didn't realize that "n" was actually a tool here. Maybe add a link for both NVM and n?

i.e.
> \# use node v7.9.0 by using a node version manager such as [nvm](https://github.com/creationix/nvm) or [n](https://github.com/tj/n)

PR #321, opened 2018-07-10
fix(Send, PrivateKeyReader): remove reload and show import error alerts

woudini:
Issue:
unable to send from watch-only address - after scanning a private key, nothing happens, and infinite spinner results.

Stack:
`KeyImportCoordinator` presents view controller to scan private key -> `SendBitcoinViewController`'s `viewDidDisappear` -> `reload` -> `self.fromAddress = @""`

Notes:
This will reopen the bug that the `reload` in `SendBitcoinViewController`'s `viewDidDisappear` was intended to fix, which was clearing the fields when navigating away from them (reminder: do not link the tickets on github).

My suggestion is to avoid trying to clear the fields until further wallet refactoring since it could create mismatching states between the UI and payment object.

Requesting feedback on moving the import error alert presentation to the `PrivateKeyReader` - I feel like it does belong in the `KeyImportCoordinator`, but it would seem a bit much to add another delegate to `PrivateKeyReader`.

Other fixes:
- Fixed strings to match translations
- Changed string used in alert.

### DIFF
--- a/Blockchain/Coordinators/KeyImportCoordinator.swift
+++ b/Blockchain/Coordinators/KeyImportCoordinator.swift
@@ -131,7 +131,7 @@ extension KeyImportCoordinator: WalletKeyImportDelegate {
         let importedKeyButForIncorrectAddress = LocalizationConstants.AddressAndKeyImport.importedKeyButForIncorrectAddress
         let importedKeyDoesNotCorrespondToAddress = LocalizationConstants.AddressAndKeyImport.importedKeyDoesNotCorrespondToAddress
         let message = String(format: "%@\n\n%@", importedKeyButForIncorrectAddress, importedKeyDoesNotCorrespondToAddress)
-        AlertViewPresenter.shared.standardNotify(message: message, title: LocalizationConstants.okString, handler: nil)
+        AlertViewPresenter.shared.standardNotify(message: message, title: LocalizationConstants.success, handler: nil)
     }
 
     @objc func alertUserOfImportedKey() {
--- a/Blockchain/Coordinators/KeyImportCoordinator.swift
+++ b/Blockchain/Coordinators/KeyImportCoordinator.swift
@@ -265,8 +265,4 @@ extension KeyImportCoordinator: PrivateKeyReaderDelegate {
     func didFinishScanning(_ privateKey: String, for address: AssetAddress?) {
         walletManager.wallet.addKey(privateKey, toWatchOnlyAddress: address?.address)
     }
-
-    func didFinishScanningWithError(_ error: PrivateKeyReaderError) {
-        // No custom implementation - error presentation handled by PrivateKeyReader
-    }
 }
--- a/Blockchain/Coordinators/KeyImportCoordinator.swift
+++ b/Blockchain/Coordinators/KeyImportCoordinator.swift
@@ -267,13 +267,6 @@ extension KeyImportCoordinator: PrivateKeyReaderDelegate {
     }
 
     func didFinishScanningWithError(_ error: PrivateKeyReaderError) {
-        switch error {
-        case .badMetadataObject:
-            AlertViewPresenter.shared.standardError(message: LocalizationConstants.Errors.error)
-        case .unknownKeyFormat:
-            AlertViewPresenter.shared.standardError(message: LocalizationConstants.AddressAndKeyImport.unknownKeyFormat)
-        case .unsupportedPrivateKey:
-            AlertViewPresenter.shared.standardError(message: LocalizationConstants.AddressAndKeyImport.unsupportedPrivateKey)
-        }
+        // No custom implementation - error presentation handled by PrivateKeyReader
     }
 }
--- a/Blockchain/PrivateKeyReader/PrivateKeyReader.swift
+++ b/Blockchain/PrivateKeyReader/PrivateKeyReader.swift
@@ -8,9 +8,9 @@
 
 import Foundation
 
-protocol PrivateKeyReaderDelegate: class {
+@objc protocol PrivateKeyReaderDelegate: class {
     func didFinishScanning(_ privateKey: String, for address: AssetAddress?)
-    func didFinishScanningWithError(_ error: PrivateKeyReaderError)
+    @objc optional func didFinishScanningWithError(_ error: PrivateKeyReaderError)
 }
 
 // TODO: remove once AccountsAndAddresses and SendBitcoinViewController are migrated to Swift
@@ -181,7 +181,7 @@ final class PrivateKeyReader: UIViewController & AVCaptureMetadataOutputObjectsD
             metadataObject.type == .qr,
             let codeObject = metadataObject as? AVMetadataMachineReadableCodeObject,
             let stringValue = codeObject.stringValue else {
-                delegate?.didFinishScanningWithError(.badMetadataObject)
+                delegate?.didFinishScanningWithError?(.badMetadataObject)
                 // TODO: remove once LegacyPrivateKeyDelegate is deprecated
                 legacyDelegate?.didFinishScanningWithError(.badMetadataObject)
                 return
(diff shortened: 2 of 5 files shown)


Blockchain/Coordinators/KeyImportCoordinator.swift
```diff
@@ -267,13 +267,6 @@ extension KeyImportCoordinator: PrivateKeyReaderDelegate {
...
     func didFinishScanningWithError(_ error: PrivateKeyReaderError) {
-        switch error {
-        case .badMetadataObject:
-            AlertViewPresenter.shared.standardError(message: LocalizationConstants.Errors.error)
-        case .unknownKeyFormat:
-            AlertViewPresenter.shared.standardError(message: LocalizationConstants.AddressAndKeyImport.unknownKeyFormat)
-        case .unsupportedPrivateKey:
-            AlertViewPresenter.shared.standardError(message: LocalizationConstants.AddressAndKeyImport.unsupportedPrivateKey)
-        }
+        // No custom implementation - error presentation handled by PrivateKeyReader
```
bchrisarriola: Might as well remove this method from `PrivateKeyReaderDelegate`. Alternatively, mark this method as `optional` in the protocol definition.

achtenhagen:
I'm not exactly clear on
> moving the import error alert presentation to the PrivateKeyReader

Why does this need to happen?

Blockchain/Coordinators/KeyImportCoordinator.swift
bchrisarriola: Also looks like alerts may be presented twice since the implementation in `PrivateKeyReader` is still calling into the delegate methods.

woudini: Alerts aren't presented twice, I set a breakpoint in https://github.com/blockchain/My-Wallet-V3-iOS/blob/fix-send-watchonly/Blockchain/Alerts/AlertViewPresenter.swift#L142 and it's only hit one time when trying to send from watch-only but scanning an incorrect key format.

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -8,9 +8,9 @@
 
 import Foundation
 
-protocol PrivateKeyReaderDelegate: class {
+@objc protocol PrivateKeyReaderDelegate: class {
```
bchrisarriola: Why does this need the `@objc` annotation? Doesn't look like this protocol is used in obj-c at all.

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -211,28 +211,41 @@ final class PrivateKeyReader: UIViewController & AVCaptureMetadataOutputObjectsD
...
+    func didFinishScanning(_ privateKey: String, for address: AssetAddress?) {
+        self.delegate?.didFinishScanning(privateKey, for: address)
+        // TODO: remove once LegacyPrivateKeyDelegate is deprecated
+        self.legacyDelegate?.didFinishScanning(privateKey)
+    }
+
+    func didFinishScanningWithError(_ error: PrivateKeyReaderError) {
+        self.delegate?.didFinishScanningWithError?(error)
+        // TODO: remove once LegacyPrivateKeyDelegate is deprecated
+        self.legacyDelegate?.didFinishScanningWithError(error)
```
bchrisarriola: `didFinishScanningWithError` in `AccountsAndAddressesViewController` has the same handling and so alerts would be presented twice. To fix this, `didFinishScanningWithError` should also be optional in `LegacyPrivateKeyDelegate` and remove the `didFinishScanningWithError` method in `AccountsAndAddressesViewController`.

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -8,9 +8,9 @@
 
 import Foundation
 
-protocol PrivateKeyReaderDelegate: class {
+@objc protocol PrivateKeyReaderDelegate: class {
     func didFinishScanning(_ privateKey: String, for address: AssetAddress?)
-    func didFinishScanningWithError(_ error: PrivateKeyReaderError)
+    @objc optional func didFinishScanningWithError(_ error: PrivateKeyReaderError)
```
bchrisarriola: Doesn't need to be annotated with @objc

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -181,7 +181,7 @@ final class PrivateKeyReader: UIViewController & AVCaptureMetadataOutputObjectsD
             metadataObject.type == .qr,
             let codeObject = metadataObject as? AVMetadataMachineReadableCodeObject,
             let stringValue = codeObject.stringValue else {
-                delegate?.didFinishScanningWithError(.badMetadataObject)
+                delegate?.didFinishScanningWithError?(.badMetadataObject)
                 // TODO: remove once LegacyPrivateKeyDelegate is deprecated
                 legacyDelegate?.didFinishScanningWithError(.badMetadataObject)
```
bchrisarriola: Replace this to call `self.didFinishScanningWithError(.badMetadataObject)`

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -211,28 +211,41 @@ final class PrivateKeyReader: UIViewController & AVCaptureMetadataOutputObjectsD
...
                 //: Pass valid private key back via success handler
-                self.delegate?.didFinishScanning(scannedKey, for: self.assetAddress)
-                // TODO: remove once LegacyPrivateKeyDelegate is deprecated
-                self.legacyDelegate?.didFinishScanning(scannedKey)
+                self.didFinishScanning(scannedKey, for: self.assetAddress)
             }
         }
     }
+
+    func didFinishScanning(_ privateKey: String, for address: AssetAddress?) {
```
bchrisarriola: private

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
```diff
@@ -211,28 +211,41 @@ final class PrivateKeyReader: UIViewController & AVCaptureMetadataOutputObjectsD
...
+        self.delegate?.didFinishScanning(privateKey, for: address)
+        // TODO: remove once LegacyPrivateKeyDelegate is deprecated
+        self.legacyDelegate?.didFinishScanning(privateKey)
+    }
+
+    func didFinishScanningWithError(_ error: PrivateKeyReaderError) {
```
bchrisarriola: private

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
woudini: The compiler errors with `'optional' can only be applied to members of an @objc protoco`

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
woudini: See https://github.com/blockchain/My-Wallet-V3-iOS/pull/321/files/259af320c8d160cc0a24318116d093602473e8e7#r201503941

Blockchain/PrivateKeyReader/PrivateKeyReader.swift
bchrisarriola: Ah. That's right. :/ 

What about this approach? https://stackoverflow.com/a/30744501

woudini:
>>moving the import error alert presentation to the PrivateKeyReader

>Why does this need to happen?

Because the relevant error alert presentation doesn't make sense where it is now. It's implemented twice, once here https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/AccountsAndAddressesViewController.m#L238 (which is called) and also here https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/Coordinators/KeyImportCoordinator.swift#L269 (which is never called because the `PrivateKeyReader`'s `delegate` is never set). By moving the implementation to `PrivateKeyReader` both Send and Addresses can use the same alert presentation code without having to add another delegate to the `PrivateKeyReader` interface.

PR #324, opened 2018-07-10
fix(Exchange): pass in enum instead of string

woudini:
Will create a ticket to add a `throw` somewhere to catch these kinds of oversights.

### DIFF
--- a/Blockchain/ExchangeCreateViewController.m
+++ b/Blockchain/ExchangeCreateViewController.m
@@ -1216,7 +1216,7 @@
 
 - (NSString *)bitcoinLabelText
 {
-    return [WalletManager.sharedInstance.wallet getActiveAccountsCount:LegacyAssetTypeBitcoin] > 1 ? [WalletManager.sharedInstance.wallet getLabelForAccount:self.btcAccount assetType:[AssetTypeLegacyHelper descriptionFor:AssetTypeBitcoin]] : [AssetTypeLegacyHelper descriptionFor:AssetTypeBitcoin];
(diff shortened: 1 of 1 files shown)


Blockchain/ExchangeCreateViewController.m
```diff
@@ -1216,7 +1216,7 @@ - (NSString *)assetPlaceholder
 
 - (NSString *)bitcoinLabelText
 {
-    return [WalletManager.sharedInstance.wallet getActiveAccountsCount:LegacyAssetTypeBitcoin] > 1 ? [WalletManager.sharedInstance.wallet getLabelForAccount:self.btcAccount assetType:[AssetTypeLegacyHelper descriptionFor:AssetTypeBitcoin]] : [AssetTypeLegacyHelper descriptionFor:AssetTypeBitcoin];
+    return [WalletManager.sharedInstance.wallet getActiveAccountsCount:LegacyAssetTypeBitcoin] > 1 ? [WalletManager.sharedInstance.wallet getLabelForAccount:self.btcAccount assetType:LegacyAssetTypeBitcoin] : [AssetTypeLegacyHelper descriptionFor:AssetTypeBitcoin];
```
bchrisarriola: I'm surprised this didn't even show a warning—pretty sure in Swift this wouldn't even compile. 

Ternary expression is nice but something we can also do to improve readability here is to break this up into multiple lines:

```
if ([WalletManager.sharedInstance.wallet getActiveAccountsCount:LegacyAssetTypeBitcoin] > 1) {
    return [WalletManager.sharedInstance.wallet getLabelForAccount:self.btcAccount assetType:AssetTypeBitcoin];
}
return [AssetTypeLegacyHelper descriptionFor:AssetTypeBitcoin];
```

PR #325, opened 2018-07-10
fix(Exchange): show continue button when selecting 'use minimum/maximum'

bchrisarriola:
Fixes the issue wherein tapping on 'Use minimum' or 'Use maximum' does not allow the user to continue until they tap on a UITextField.

With this change, if the user taps on those buttons and the keyboard isn't presented, the topLeftField will become the first responder so that the keyboard is presented.


bchrisarriola:
Didn't realize there was supposed to be another continue button. I think showing that secondary continue button is much prefer. Let me dig into why that's not showing and update this PR

bchrisarriola:
Some screenshots:

iPhone X:
simulator screen shot - iphone x - 2018-07-11 at 11 54 12

iPhone 6S:
simulator screen shot - iphone 6s - 2018-07-11 at 11 58 04

woudini:
Since that is back now, do you think it makes sense to revert the first commit https://github.com/blockchain/My-Wallet-V3-iOS/pull/325/commits/87d78f9a7bb11e4e580efe0e6019c5c53e7b5fb5 ? It's a bit overwhelming with so many changes on the screen.

PR #326, opened 2018-07-11
fix(App Review): show prompt on third completed transaction regardless of how often the application has become active

achtenhagen:
Highlights in this PR:
- Fixes an issue where the app review prompt would not show during the user's first session if the user has not had at least 3 transactions in total. The prompt will now show after the third total payment has been sent, regardless of how many times the application has been opened.

### DIFF
--- a/Blockchain/AppDelegate.swift
+++ b/Blockchain/AppDelegate.swift
@@ -40,6 +40,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         Fabric.with([Crashlytics.self])
 
+        BlockchainSettings.App.shared.appBecameActiveCount += 1
+
         // MARK: - Global Appearance
 
         //: Status Bar
@@ -153,7 +155,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func applicationWillEnterForeground(_ application: UIApplication) {
         print("applicationWillEnterForeground")
 
-        BlockchainSettings.App.shared.appOpenedCount += 1
+        BlockchainSettings.App.shared.appBecameActiveCount += 1
 
         BuySellCoordinator.shared.start()
 
--- a/Blockchain/AppReviewPrompt/AppReviewPrompt.swift
+++ b/Blockchain/AppReviewPrompt/AppReviewPrompt.swift
@@ -31,19 +31,20 @@ final class AppReviewPrompt: NSObject {
     }
 
     /// Ask to show the prompt, else handle failure silently
-    @objc func askToShow() {
-        if WalletManager.shared.wallet.getAllTransactionsCount() < numberOfTransactionsBeforePrompt {
+    @objc func showIfNeeded() {
+        let transactionsCount = WalletManager.shared.wallet.getAllTransactionsCount()
+        if transactionsCount < numberOfTransactionsBeforePrompt {
             #if DEBUG
             print("App review prompt will not show because the user needs at least \(numberOfTransactionsBeforePrompt) transactions.")
             #endif
             return
         }
-
-        // TODO: support overriding appOpenedCount for debugging
-        let count = BlockchainSettings.App.shared.appOpenedCount
+        // TODO: support overriding appBecameActiveCount for debugging
+        let count = BlockchainSettings.App.shared.appBecameActiveCount
         switch count {
-        case 10, 50: requestReview()
-        case _ where (count >= 100) && (count % 100 == 0): requestReview()
+        case 10, 50,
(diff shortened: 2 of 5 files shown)


Blockchain/AppDelegate.swift
```diff
@@ -40,6 +40,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         Fabric.with([Crashlytics.self])
 
+        BlockchainSettings.App.shared.appOpenedCount += 1
```
woudini: I'd rename this to something like `appLaunchedCount` since opening the app is different from invoking `didFinishLaunchingWithOptions`

Blockchain/AppReviewPrompt/AppReviewPrompt.swift
```diff
@@ -32,18 +32,19 @@ final class AppReviewPrompt: NSObject {
...
         }
-
         // TODO: support overriding appOpenedCount for debugging
         let count = BlockchainSettings.App.shared.appOpenedCount
         switch count {
-        case 10, 50: requestReview()
-        case _ where (count >= 100) && (count % 100 == 0): requestReview()
+        case 10, 50,
+             _ where (count >= 100) && (count % 100 == 0),
+             _ where transactionsCount == numberOfTransactionsBeforePrompt: requestReview()
```
woudini: FYI `WalletManager.shared.wallet.getAllTransactionsCount()` doesn't count transactions made during a single session - it counts the total BTC transactions fetched from multiaddress (which for now maxes out at 50).

Blockchain/AppDelegate.swift
achtenhagen: Naming it `appLaunchedCount` would not make sense here because this line is also called from `applicationWillEnterForeground`. If it must be renamed, a better alternative would be `appBecameActiveCount`.

Blockchain/AppReviewPrompt/AppReviewPrompt.swift
achtenhagen: That's fine as the user does not need to make three transactions in a single session, but rather just needs to have a total of 3 transactions made.

woudini: Also PR title implies `transactionsCount >= numberOfTransactionsBeforePrompt` but here it's `transactionsCount == numberOfTransactionsBeforePrompt`

achtenhagen: This is done intentionally, because it covers the case of the user having the third transaction during an applicationBecameActive count of < 10. Otherwise you would need to complete three transactions AND have the application become active 10 times.

achtenhagen: I will squash these commits into one and update the commit message with a more clear message.

woudini: >This is done intentionally, because it covers the case of the user having the third transaction during an applicationBecameActive count of < 10. Otherwise you would need to complete three transactions AND have the application become active 10 times.

If by `This` you mean using `==` instead of `>=`, that doesn't really make sense. Running the following in the playground prints "request", with `applicationBecameActiveCount` < 10 and a `transactionsCount` of 3:
```
let applicationBecameActiveCount = 0
let transactionsCount = 3
switch applicationBecameActiveCount {
    case 10, 50,
     _ where (applicationBecameActiveCount >= 100) && (applicationBecameActiveCount % 100 == 0),
     _ where transactionsCount >= 3: requestReview()
    default: print("e")
}

func requestReview() {
    print("request")
}
```
The only thing that using `==` instead of `>=` is doing is preventing users that already have more than 3 transactions from ever seeing the review prompt. AFAIK there aren't strict requirements around reviews so the results don't matter too much here, but it's important for everyone to be on the same page as far as what the code does (and/or communicating what `This` is if I misunderstood).

achtenhagen: > The only thing that using == instead of >= is doing is preventing users that already have more than 3 transactions from ever seeing the review prompt.

... This is false because, if the user has had more than 3 transactions, the switch case will **still** call `requestReview` whenever the `appBecameActiveCount` is either 10, 50, or every 100th consecutive time after that.

Your script is incorrect because changing `==` to `>=` will cause the switch case to call `requestReview` in every case where `transactionsCount` is greater than or equal to three (regardless of what the value of `appBecameActiveCount` is), which should **not be the case**.

I am aware that the requirements around reviews do not matter a great deal as the system itself will decide in the end whether to show the prompt or not. However, I have pointed this out on previous occasions and this is how the logic for showing the app review prompt has been decided.

woudini: Thanks for the clarification.

PR #327, opened 2018-07-12
fix(Upgrade): handle wallet upgrade for v2 wallets with a 2nd password.

bchrisarriola:
Fix the issue wherein v2 wallets aren't forced to be upgraded if they have a 2nd password set.

### Summary
This fix includes the following changes:
* The logic to present `UpgradeViewController` was moved to `TabControllerManager` (we had that in multiple places previously but now it's just in one place)
* Created `WalletUpgradeDelegate` so that upgrade success notification can be propagated up from the `Wallet`
* Not dismissing `UpgradeViewController` when the user taps on "Continue". This is necessary since previously the 2nd password view was being presented from `UpgradeViewController` which didn't show since that view controller was being dismissed

...with this change the "Upgrade" side menu item is not even necessary since this will always force the user to upgrade.

### Testing
Tested this change on both V2 wallets w/ and w/o 2nd password set.

### DIFF
--- a/Blockchain/TabControllerManager.m
+++ b/Blockchain/TabControllerManager.m
@@ -38,6 +38,12 @@
     walletManager.fiatAtTimeDelegate = self;
 }
 
+- (void)viewDidAppear:(BOOL)animated
+{
+    [super viewDidAppear:animated];
+    [AppCoordinator.sharedInstance showHdUpgradeViewIfNeeded];
+}
+
 - (void)didSetAssetType:(LegacyAssetType)assetType
 {
     self.assetType = assetType;
--- a/Blockchain/UpgradeViewController.m
+++ b/Blockchain/UpgradeViewController.m
@@ -221,12 +221,12 @@
 - (void)onWalletUpgraded
 {
     __weak UpgradeViewController *weakSelf = self;
-    [AlertViewPresenter.sharedInstance standardErrorWithMessage:LocalizationConstantsObjcBridge.upgradeSuccess
-                                                          title:LocalizationConstantsObjcBridge.upgradeSuccessTitle
-                                                             in:self
-                                                        handler:^(UIAlertAction * _Nonnull action) {
-                                                            [weakSelf dismissViewControllerAnimated:YES completion:nil];
-                                                        }];
+    [AlertViewPresenter.sharedInstance standardNotifyWithMessage:LocalizationConstantsObjcBridge.upgradeSuccess
+                                                           title:LocalizationConstantsObjcBridge.upgradeSuccessTitle
+                                                              in:self
+                                                         handler:^(UIAlertAction * _Nonnull action) {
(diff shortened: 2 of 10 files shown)


Blockchain/UpgradeViewController.m
```diff
@@ -219,4 +216,17 @@ - (void)scrollViewDidEndDecelerating:(UIScrollView *)scrollView
     [self setTextForCaptionLabel];
 }
 
+#pragma mark - WalletUpgradeDelegate
+
+- (void)onWalletUpgraded
+{
+    __weak UpgradeViewController *weakSelf = self;
+    [AlertViewPresenter.sharedInstance standardErrorWithMessage:LocalizationConstantsObjcBridge.upgradeSuccess
```
woudini: A bit trivial but `standardErrorWithMessage:` with success strings is a bit confusing, any reason to not use `standardNotifyWithMessage:`?

Blockchain/TabControllerManager.m
```diff
@@ -38,6 +38,12 @@ - (void)viewDidLoad
     walletManager.fiatAtTimeDelegate = self;
 }
 
+- (void)viewDidAppear:(BOOL)animated
+{
+    [super viewDidAppear:animated];
+    [AppCoordinator.sharedInstance showHdUpgradeViewIfNeeded];
```
woudini: Upgrade prompt is showing when launching the app without a wallet paired - probably needs to be
```
if ([WalletManager.sharedInstance.wallet isInitialized]) {
    [AppCoordinator.sharedInstance showHdUpgradeViewIfNeeded];
}

Comment (bchrisarriola) [reply Blockchain/TabControllerManager.m]: Yikes. Good catch

File: Blockchain/UpgradeViewController.m (hunk 1 above)
Comment (bchrisarriola) [reply Blockchain/UpgradeViewController.m]: Yeah that's better

File: Blockchain/Wallet/WalletManager.swift [hunk 3]
```diff
@@ -625,4 +626,12 @@ extension WalletManager: WalletDelegate {
     @objc func getPrivateKeyPassword(withSuccess success: WalletSuccessCallback) {
         secondPasswordDelegate?.getPrivateKeyPassword(success: success)
     }
+
+    // MARK: - Upgrade
+
+    func walletUpgraded(_ wallet: Wallet!) {
```
woudini: Can the `wallet` arg be removed?

bchrisarriola: This is a pretty standard pattern (e.g. similar to how `UITextFieldDelegate` all return the specific `textField` invoking the delegate method) that I'd like to keep which is useful to have should our requirements change when a wallet is upgraded. 

This way, if we did need to do another wallet operation as a result of upgrading a wallet, the delegate method doesn't need to make an assumption that `WalletManager.shared.wallet` was the component invoking that method. Instead, it can use the specific instance provided in the delegate method.

PR #328, opened 2018-07-12
fix(WalletManager): add back legacy code for reloading send screens

woudini:
To fix the issue where logging out/forgetting the wallet and logging back in shows the same `from` label as before, even if a different wallet is being used.

For reference, originally added here: https://github.com/blockchain/My-Wallet-V3-iOS/commit/5a4ed083faaeae128413129fd53b15dbe001ba68

### DIFF
--- a/Blockchain/ExchangeCreateViewController.m
+++ b/Blockchain/ExchangeCreateViewController.m
@@ -1218,17 +1218,26 @@
 
 - (NSString *)bitcoinLabelText
 {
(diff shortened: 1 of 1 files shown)


Blockchain/Wallet/WalletManager.swift
```diff
@@ -187,6 +187,9 @@ extension WalletManager: WalletDelegate {
         wallet.bchSwipeAddressToSubscribe = nil
         wallet.twoFactorInput = nil
 
+        AppCoordinator.shared.tabControllerManager.sendBitcoinViewController?.reload()
```
bchrisarriola: This should already be done in [AuthenticationCoordinator.authHandler](https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/Authentication/AuthenticationCoordinator.swift#L87...L89) which should be invoked by `authenticationCompleted`

Side-note: UI-related code shouldn't be here

woudini:
I don't remember closing this...weird

PR #331, opened 2018-07-13
feat(Buy and Sell): Create coordinators for native buy/sell

bchrisarriola:
Creating coordinators for the native buy/sell flow. Stubbed in properties/methods that are to be implemented. 

The native buy/sell flow: https://cloudup.com/ctuALWx1VtM


bchrisarriola:
Closing this. Will reopen when actual work starts on native buy/sell

PR #322, opened 2018-07-10
refactor(Colors): migrate colors and use UIColor extension

achtenhagen:
Highlights in this PR:
- Adds new `UIColor` extension for simpler color management.
- Migrates remaining colors from `Blockchain-Prefix.pch` file.

PR #333, opened 2018-07-13
fix(Recover Wallet): Update email validation

jstxx:
Fixes an issue creating and recovering wallets when using an invalid email address. Will need to incorporate changes from https://github.com/blockchain/My-Wallet-V3-iOS/pull/322 before merging.

### DIFF
--- a/Blockchain/Extensions/Strings.swift
+++ b/Blockchain/Extensions/Strings.swift
@@ -40,3 +40,16 @@ extension String {
         return queryArgs
     }
 }
+
+extension NSString {
+    @objc func isEmail() -> Bool {
+        let detector = try? NSDataDetector(types: NSTextCheckingResult.CheckingType.link.rawValue)
+        var validated = false
+        let range = NSRange(location: 0, length: (self as NSString).length)
+        
+        detector?.enumerateMatches(in: self as String, options: [], range: range) { result, _, _ in
+            validated = result?.url?.scheme == "mailto"
+        }
+        return validated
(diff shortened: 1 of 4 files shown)


Blockchain/Extensions/Strings.swift
```diff
@@ -40,3 +40,16 @@ extension String {
         return queryArgs
     }
 }
+
+extension NSString {
+    @objc func isEmail() -> Bool {
```
bchrisarriola: 🔥`BCCreateWalletView` also needs to be updated so that it uses this email validation. Specifically, this part of the code: https://github.com/blockchain/My-Wallet-V3-iOS/blob/dc35f92ff6e54adeecae80c2851eb3a2ea37da76/Blockchain/BCCreateWalletView.m#L369...L371

jstxx: ✅ Added to `BCCreateWalletView`

bchrisarriola: Lgtm!

Only change is that the My-Wallet-V3 submodule needs to be updated to use commit `045b78b` and should be good to go 👍

jstxx: ✅ Updated Submodules

bchrisarriola:
I think this can be merged now 💯 

Just need to rebase this against `dev` to fix conflicts.

jstxx:
Going to update this to use the new color scheme

PR #335, opened 2018-07-16
fix(TabBar): Fix iPhoneX TabBar constraints

jstxx:
Fixes an issue that could potentially cause the tab bar to be improperly formatted on iPhone X


jstxx:
Previously the tab bar was in the main view of the UIViewController, pinned to the bottom of the superview. Now the tab bar is embedded in a UIView, removing the superview margins, updating the vertical compression resistance and using anchor constraints.

PR #332, opened 2018-07-13
refactor(Settings): Use Autolayout and Swift for Settings

jstxx:
Migrates Settings and associated controllers to Swift, converts programmatic UITableView creation to storyboards. Also makes some abstractions and structured data.

PR #337, opened 2018-07-19
Fix warnings

jstxx:
Removes nullability warnings from displaying entirely, TODOs, and addresses other misc warnings.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -42,6 +42,7 @@ required_enum_case: true
 sorted_imports: true
 strict_fileprivate: true
 switch_case_on_newline: true
+todo: false
 trailing_closure: true
 type_body_length:
   warning: 500
--- a/Blockchain/BCCreateWalletView.m
+++ b/Blockchain/BCCreateWalletView.m
@@ -110,7 +110,7 @@
             self.frame = frame;
         }
         
-        [emailTextField becomeFirstResponder];
+        [self->emailTextField becomeFirstResponder];
     });
 }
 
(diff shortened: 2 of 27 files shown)


.swiftlint.yml
```diff
@@ -42,6 +42,7 @@ required_enum_case: true
 sorted_imports: true
 strict_fileprivate: true
 switch_case_on_newline: true
+todo: false
```
bchrisarriola: 👍

Blockchain/BCCreateWalletView.m
```diff
@@ -347,11 +347,11 @@ - (void)checkPasswordStrength
     self.passwordStrength = passwordStrength;
     
     [UIView animateWithDuration:ANIMATION_DURATION animations:^{
-        passwordFeedbackLabel.text = description;
-        passwordFeedbackLabel.textColor = color;
-        passwordStrengthMeter.progress = passwordStrength/100;
-        passwordStrengthMeter.progressTintColor = color;
-        passwordTextField.layer.borderColor = color.CGColor;
+        self->passwordFeedbackLabel.text = description;
```
bchrisarriola: I know this isn't warning-related, but should we change this to use a weak reference to self? Same comment applies to other changes in this PR where self is referenced within a closure.

jstxx: I don't see any problem with that, @woudini any objection?

thisisalexmcgregor: FWIW, I don't believe you need `weakSelf` in a UIView animation block. You can just do `self.`

woudini: Agree with @thisisalexmcgregor

Blockchain/Network/Models/WalletServiceError.swift
```diff
@@ -14,5 +14,6 @@ internal enum WalletServiceError: Error {
     case jsonParseError
     
     /// A generic error with an optional error message
+    // swiftlint:disable next identifier_name
```
woudini: Same as https://github.com/blockchain/My-Wallet-V3-iOS/pull/332#discussion_r204860255, showing as a compile error for me

PR #338, opened 2018-07-19
fix(Dashboard): hide non-spendable balance if watch only total balanc…

woudini:
…e is zero


Blockchain/js/wallet-ios.js
```diff
@@ -286,11 +286,6 @@ MyWalletPhone.watchOnlyBalance = function() {
     .reduce(Helpers.add, 0);
 }
 
-MyWalletPhone.hasWatchOnlyAddresses = function() {
```
bchrisarriola: This seems like this might still be useful. Thoughts on keeping it instead?

PR #341, opened 2018-07-20
fi(Exchange): enforce BCH upper limit from wallet options

woudini:
Goal is to merge this today into `release` for a build.

**Requirement:**
If the user is in the US, enforce the upper limit from wallet options for the `from` asset type in USD. Otherwise, enforce it in EUR.

**Issue:**
BCH upper limit from wallet options was not being fetched and returned when the `from` asset type was BCH. As a result, an incorrect conversion rate was used for this limit.

The current implementation for enforcing these limits were quite messy and nonsensical:
- Regardless of the `from` asset type, both BTC and ETH hard limits were being returned to the `ExchangeCreateViewController`
- The country code check for using USD or EUR was implemented incorrectly.
- For BTC, it was getting the last fetched rate, and for ETH it was fetching the latest rate.

**Fixes**
- Fetch and return the rate for only the `from` asset type
- Return the country code instead of a boolean

**Feedback Requests**
- https://github.com/blockchain/My-Wallet-V3-iOS/blob/kevin/IOS-1095/fix_bch_limit/Blockchain/js/wallet-ios.js#L2596-L2602 as I'm sure it can rewritten better. 
- What level of refactoring are we comfortable when making these kinds of fixes to put in a build quickly? We should use `Constants.swift` instead of the `Blockchain-Prefix.pch`. To take it even further we should create an object instead of using all of these dictionary keys here https://github.com/blockchain/My-Wallet-V3-iOS/blob/kevin/IOS-1095/fix_bch_limit/Blockchain/Wallet.m#L823

### DIFF
--- a/Blockchain/Blockchain-Prefix.pch
+++ b/Blockchain/Blockchain-Prefix.pch
@@ -475,9 +475,8 @@ green:((float)((rgbValue & 0xFF00) >> 8))/255.0 blue:((float)(rgbValue & 0xFF))/
 #define DICTIONARY_KEY_PAIR @"pair"
 #define DICTIONARY_KEY_RATE @"rate"
 #define DICTIONARY_KEY_MINER_FEE @"minerFee"
-#define DICTIONARY_KEY_BTC_HARD_LIMIT @"btcHardLimit"
-#define DICTIONARY_KEY_ETH_HARD_LIMIT @"ethHardLimit"
-#define DICTIONARY_KEY_ETH_HARD_LIMIT_RATE @"ethHardLimitRate"
+#define DICTIONARY_KEY_HARD_LIMIT @"hardLimit"
+#define DICTIONARY_KEY_HARD_LIMIT_RATE @"hardLimitRate"
 
 #define DICTIONARY_KEY_BITCOIN @"bitcoin"
 #define DICTIONARY_KEY_BITCOIN_CASH @"bitcoinCash"
--- a/Blockchain/Blockchain-Prefix.pch
+++ b/Blockchain/Blockchain-Prefix.pch
@@ -469,14 +469,9 @@ green:((float)((rgbValue & 0xFF00) >> 8))/255.0 blue:((float)(rgbValue & 0xFF))/
 #define DICTIONARY_KEY_MULTIADDRESS_SYMBOL_LOCAL @"symbol_local"
 #define DICTIONARY_KEY_MULTIADDRESS_SYMBOL_BTC @"symbol_btc"
 
-#define DICTIONARY_KEY_LIMIT @"limit"
-#define DICTIONARY_KEY_MINIMUM @"minimum"
-#define DICTIONARY_KEY_MAX_LIMIT @"maxLimit"
 #define DICTIONARY_KEY_PAIR @"pair"
 #define DICTIONARY_KEY_RATE @"rate"
 #define DICTIONARY_KEY_MINER_FEE @"minerFee"
-#define DICTIONARY_KEY_HARD_LIMIT @"hardLimit"
-#define DICTIONARY_KEY_HARD_LIMIT_RATE @"hardLimitRate"
 
 #define DICTIONARY_KEY_BITCOIN @"bitcoin"
 #define DICTIONARY_KEY_BITCOIN_CASH @"bitcoinCash"
--- a/Blockchain/ExchangeRate.swift
+++ b/Blockchain/ExchangeRate.swift
@@ -1,0 +1,70 @@
+//
+//  ExchangeRate.swift
+//  Blockchain
+//
+//  Created by kevinwu on 7/20/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+/// Used to return exchange information from the ShapeShift API and wallet-options to the ExchangeCreateViewController.
+@objc class ExchangeRate: NSObject {
+
+    private struct Keys {
+        static let limit = "limit"
+        static let minimum = "minimum"
+        static let minerFee = "minerFee"
+        static let maxLimit = "maxLimit"
+        static let rate = "rate"
+        static let hardLimit = "hardLimit"
+        static let hardLimitRate = "hardLimitRate"
+    }
+
+    @objc let limit: String? // Not currently used on web, so not used for iOS, but added for documentation
+    @objc let minimum: String? // Minimum amount required to exchange
+    @objc let minerFee: String? // Fee for exchange
+    @objc let maxLimit: String? // Maximum amount allowed to exchange, defined by ShapeShift
+    @objc let rate: String? // Exchange rate between the 'from' and 'to' asset types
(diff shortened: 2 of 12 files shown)


Blockchain/Blockchain-Prefix.pch
```diff
@@ -475,9 +475,8 @@ green:((float)((rgbValue & 0xFF00) >> 8))/255.0 blue:((float)(rgbValue & 0xFF))/
 #define DICTIONARY_KEY_PAIR @"pair"
 #define DICTIONARY_KEY_RATE @"rate"
 #define DICTIONARY_KEY_MINER_FEE @"minerFee"
-#define DICTIONARY_KEY_BTC_HARD_LIMIT @"btcHardLimit"
-#define DICTIONARY_KEY_ETH_HARD_LIMIT @"ethHardLimit"
-#define DICTIONARY_KEY_ETH_HARD_LIMIT_RATE @"ethHardLimitRate"
+#define DICTIONARY_KEY_HARD_LIMIT @"hardLimit"
+#define DICTIONARY_KEY_HARD_LIMIT_RATE @"hardLimitRate"
```
bchrisarriola: Move these to `Constants`?

Blockchain/Wallet.m
```diff
@@ -819,8 +819,8 @@ - (void)loadJS
         [weakSelf on_get_exchange_trades_success:trades];
     };
 
-    self.context[@"objc_on_get_exchange_rate_success"] = ^(JSValue *limit, JSValue *minimum, JSValue *minerFee, JSValue *maxLimit, JSValue *pair, JSValue *rate, JSValue *ethHardLimit) {
-        [weakSelf on_get_exchange_rate_success:@{DICTIONARY_KEY_LIMIT : [limit toString], DICTIONARY_KEY_MINIMUM : [minimum toString], DICTIONARY_KEY_MINER_FEE : [minerFee toString], DICTIONARY_KEY_MAX_LIMIT : [maxLimit toString], DICTIONARY_KEY_RATE : [rate toString], DICTIONARY_KEY_ETH_HARD_LIMIT_RATE : [ethHardLimit toString]}];
+    self.context[@"objc_on_get_exchange_rate_success"] = ^(JSValue *limit, JSValue *minimum, JSValue *minerFee, JSValue *maxLimit, JSValue *pair, JSValue *rate, JSValue *hardLimit) {
+        [weakSelf on_get_exchange_rate_success:@{DICTIONARY_KEY_LIMIT : [limit toString], DICTIONARY_KEY_MINIMUM : [minimum toString], DICTIONARY_KEY_MINER_FEE : [minerFee toString], DICTIONARY_KEY_MAX_LIMIT : [maxLimit toString], DICTIONARY_KEY_RATE : [rate toString], DICTIONARY_KEY_HARD_LIMIT_RATE : [hardLimit toString]}];
```
bchrisarriola: How about creating a class for this? This way we're passing in very clear types also all the keys can be removed and hidden within that class.

...ideally this would be a struct but since it needs obj-c compat we'll have to stick with using a class

e.g.

```
@objc class ExchangeRate: NSObject {
    let limit: NSDecimalNumber
    // ...
}

extension ExchangeRate {
     convenience init(json: JSON) {
        // ...
     }
}
```

Blockchain/js/wallet-ios.js
```diff
@@ -2581,10 +2581,25 @@ MyWalletPhone.getExchangeTrades = function() {
...
+
+        var coins = coinPair.split('_');
+        var from = coins[0];
+
+        var fetchRateSuccess = function(hardLimit) {
+            var currencyCode = MyWalletPhone.currencyCodeForHardLimit();
+            objc_on_get_exchange_rate_success(result.limit, result.minimum, result.minerFee, result.maxLimit, result.pair, result.rate, hardLimit[currencyCode].last);
+        };
+
+        if (from.indexOf(btcTicker) !== -1) {
```
bchrisarriola: Any reason why checking for the index is used instead of a direct string comparison?

e.g. `from === btcTicker`

bchrisarriola: Seems like this can also be simplified (also prefer to use immutable variables when possible):

```
const supportedTickers = ['BTC', 'ETH', 'BCH'];
const rateForCoin = coinPair.split('_')[0];
const fetchRateSuccess = function(hardLimit) {
            var currencyCode = MyWalletPhone.currencyCodeForHardLimit();
            objc_on_get_exchange_rate_success(result.limit, result.minimum, result.minerFee, result.maxLimit, result.pair, result.rate, hardLimit[currencyCode].last);
};

if (rateForCoin in supportedTickers) {
    MyWalletPhone.getExchangeRateForHardLimit(rateForCoin).then(fetchRateSuccess);
}
```

Blockchain/Blockchain-Prefix.pch
bchrisarriola: Ignore this. See my comment below about creating a class.

Blockchain/js/wallet-ios.js
woudini: The iteration here is getting all the rates but I only need it for the `from` so it turns out that all of that `if` branching I was doing was unnecessary

Blockchain/Wallet.m
woudini: I'll make this class in this case since it seems fairly small in scope, but I think it's also important to gauge the amount of changes we're making to a release build with small fixes (refer to second point in "Feedback Requests" in the PR description)

bchrisarriola: I think in this case we should implement this the "right" way, since as you mentioned, doing so is fairly small in scope. 

We're in the process of improving the wallet codebase as a whole, and while it surely will take more time to implement it the "right" way, doing so will help us in that effort.

Blockchain/ExchangeRate.swift
```diff
@@ -0,0 +1,74 @@
...
+        static let minimum = "minimum"
+        static let minerFee = "minerFee"
+        static let maxLimit = "maxLimit"
+        static let rate = "rate"
+        static let hardLimit = "hardLimit"
+        static let hardLimitRate = "hardLimitRate"
+    }
+
+    @objc let limit: String? // Not currently used on web, so not used for iOS, but added for documentation
+    @objc let minimum: String? // Minimum amount required to exchange
```
bchrisarriola: Could've sworn I left a comment about this—seems odd for all of these values to be String. Why not NSDecimalNumber?

bchrisarriola: Or just use NSNumber? 

Storing these values (which are all numbers) as strings seems inherently wrong esp. since calculations are done using them. The only point these should be converted to a string is when they are being displayed.

woudini: Agreed and changed to `NSDecimalNumber`. Looking back using strings seemed like a workaround without proper unit tests. We should work towards establishing a standard for data types and their corresponding representations.
>The only point these should be converted to a string is when they are being displayed.

 This is a good rule of thumb to start with.

PR #342, opened 2018-07-20
refactor(Settings): Migrate defaults, namespaced component architecture

jstxx:
Namespaces UserDefaults and converts KeychainItemWrapper + Credentials to Swift.


Blockchain/AppDelegate.swift
```diff
@@ -40,11 +40,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
         Fabric.with([Crashlytics.self])
 
-        BlockchainSettings.App.shared.appBecameActiveCount += 1
-
+        let activeCount = UserDefaults.System.integer(forKey: .appBecameActiveCount) // 3
```
bchrisarriola: Nit: remove `// 3` comment?

bchrisarriola: Also should this value be accessed through `BlockchainSettings`? Would be good to abstract from AppDelegate (and the application-level as a whole) that `UserDefaults` is the underlying caching mechanism for this property.

Blockchain/KeychainItemWrapper+Credentials.swift
```diff
@@ -0,0 +1,137 @@
...
+//  Blockchain
+//
+//  Created by Justin on 7/19/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+protocol KeychainFlow {
+    static func guid() -> String?
```
bchrisarriola: How about computed static properties instead of functions for some of these fields?

Also seems like we can remove `set*` and `remove*` functions by implementing a set (remove would just be setting the value to nil).

```
protocol KeychainFlow {
    static func guid: String? { get }
    static func hashedGuid: String? { get }
    static func sharedKey: String? { get }
}
```

Blockchain/LocalizationConstants.swift
```diff
@@ -33,6 +33,9 @@ struct LocalizationConstants {
         "My Ether Wallet",
         comment: "The default name of the ether wallet."
     )
+    
+    static let marketPriceTitle = NSLocalizedString("Current Price", comment: "")
```
bchrisarriola: Remove?

Blockchain/Settings/BlockchainSettings.swift
```diff
@@ -8,6 +8,144 @@
...
 import Foundation
 
+protocol KeyNamespaceable { }
+
+protocol BCSettings: KeyNamespaceable {
+    associatedtype SettingsKey: RawRepresentable
+}
+
+extension KeyNamespaceable {
+    private static func namespace(_ key: String) -> String {
```
bchrisarriola: Not sure I understand why this method is necessary—can the reference to this method just return key.rawValue?

Blockchain/Settings/BlockchainSettings.swift
```diff
@@ -8,6 +8,144 @@
...
+
+    static func bool(forKey key: SettingsKey) -> Bool {
+        let key = namespace(key)
+        return UserDefaults.standard.bool(forKey: key)
+    }
+
+    // Strings
+    static func set(_ string: String, forKey key: SettingsKey) {
+        let key = namespace(key)
+        UserDefaults.standard.set(NSString(string: string), forKey: key)
```
bchrisarriola: Is it necessary to cast to NSString?

Blockchain/Settings/BlockchainSettings.swift
```diff
@@ -44,47 +182,60 @@ final class BlockchainSettings: NSObject {
...
         }
 
         /// The encrypted wallet password (encryption key was generated by the user's pin)
         @objc var encryptedPinPassword: String? {
             get {
-                return defaults.string(forKey: UserDefaults.Keys.encryptedPinPassword.rawValue)
+                return UserDefaults.User.string(forKey: .encryptedPINPassword)
+            }
+            set {
+                if let hasPin = newValue {
```
bchrisarriola: Should this check be removed?

Blockchain/Settings/BlockchainSettings.swift
```diff
@@ -8,6 +8,144 @@
...
+        UserDefaults.standard.set(bool, forKey: key)
+    }
+
+    static func bool(forKey key: SettingsKey) -> Bool {
+        let key = namespace(key)
+        return UserDefaults.standard.bool(forKey: key)
+    }
+
+    // Strings
+    static func set(_ string: String, forKey key: SettingsKey) {
```
bchrisarriola: Allow optional?

Blockchain/Settings/BlockchainSettings.swift
```diff
@@ -107,26 +258,26 @@ final class BlockchainSettings: NSObject {
...
             set {
-                defaults.set(newValue, forKey: UserDefaults.Keys.pinKey.rawValue)
+                if let pinKey = newValue {
+                    UserDefaults.User.set(pinKey, forKey: .pinKey)
+                }
             }
         }
 
-        var onSymbolLocalChanged: ((Bool) -> Void)?
+        var onSymbolLocalChanged: ((Bool) -> ())?
```
bchrisarriola: Revert? Imo a return value of `Void` is clearer than `()`.

bchrisarriola:
My-Wallet-V3 submodule needs to be reverted

Blockchain/Settings/BlockchainSettings.swift
jstxx: So, that is true! It was in case in the future if we wanted to namespace future components

Blockchain/Settings/BlockchainSettings.swift
jstxx: Sadly

Blockchain/Settings/BlockchainSettings.swift
jstxx: Ehhh, if something is calling set I would rather the responsibility of unwrapping it fall on the responsibility of the routine.

Blockchain/Settings/BlockchainSettings.swift
bchrisarriola: Ah. What was the warning cc @achtenhagen

achtenhagen: This appears to only be an issue with Swift 4.2 + Xcode 10, where the warning appears.

jstxx:
Closing this while I investigate some regression issues.

PR #344, opened 2018-07-20
 refactor(Pin): IOS-1072 move pin change flow to PinPresenter.

bchrisarriola:
### Summary of Changes:
* Added creating/changing a pin into the new pin architecture (PinPresenter + PinInteractor + PinView)
* Added unit tests for different business flows around changing/creating a new pin (See PinPresenterTests)
* Removed change pin related code from Wallet.m/.h and wallet-ios.js

Since there are a lot of changes in here, would be good to get approval from 2 reviewers. 

#311 needs to be merged first before this can be reviewed/merged.


Blockchain/Pin/PinInteractor.swift
```diff
@@ -18,36 +18,98 @@ import RxSwift
...
+    /// - Returns: a Single returning the response
+    func createPin(_ pinPayload: PinPayload) -> Single<PinStoreResponse> {
+        return self.walletService.createPin(pinPayload)
+            .do(onSuccess: { [weak self] response in
+                try self?.handleCreatePinResponse(response: response, pinPayload: pinPayload)
+            })
+    }
+
+    private func handleCreatePinResponse(response: PinStoreResponse, pinPayload: PinPayload) throws {
+        guard let password = wallet.password else {
```
bchrisarriola: I think a `nil` password shouldn't ever happen because by the time you set your pin, a password should already have been set. I didn't remove this check so that it mimics the previous implementation. Requesting for input here to see if we can remove this check @woudini @mpfluger

mpfluger: The only case I could think of is if the user backgrounds and the response comes back afterwards. If you're sure this can't happen in this instance then I don't think that check should be there.

woudini: The scenario that Mark pointed out seems pretty unlikely since the user is already engaged when the call happens (right after entering all the digits to confirm the PIN). Also tried it a couple of times myself and couldn't get the `throw` to happen. I think it can be removed

EDIT: this is being thrown during this test https://github.com/blockchain/My-Wallet-V3-iOS/pull/344/files#diff-6e5c7e5a9e6a9b75d63aae9beca1ee71R90

BlockchainTests/Pin/PinPresenterTests.swift
```diff
@@ -14,19 +14,98 @@ import XCTest
...
+        )
+        pinView.didCallShowLoadingViewExpectation = expectation(description: "Show loading view called.")
+        pinView.didCallHideLoadingViewExpectation = expectation(description: "Hide load view called.")
+        pinView.didCallErrorExpectation = expectation(description: "Did call error.")
+        _ = presenter.validateConfirmPinForChangePin(pin: Pin(code: 2222), firstPin: Pin(code: 2222))
+        waitForExpectations(timeout: 1)
+    }
+
+    /// Tests that a valid response from the pin store flow will succeed
+    func testCreatePinSuccess() {
```
woudini: This is failing for me because the `password` is `nil`. I tried setting it to `"password"` here but am now getting this error when running this test https://github.com/blockchain/My-Wallet-V3-iOS/pull/344/files#diff-c4cb12026c070437bb698524e5739337R72 . Looking further into it
EDIT: `self.context` is `nil` at this point, so there is no encryption happening. This test requires the `Wallet` object to be instantiated.

Third Party/PinEntry/Classes/PEPinEntryController.h
```diff
@@ -70,6 +71,7 @@
 - (void)setupQRCode;
 - (void)reset;
 - (void)goToEnter1Pin;
```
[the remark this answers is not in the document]
bchrisarriola: I think you meant `goToEnter2Pin`? Will remove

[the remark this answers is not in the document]
bchrisarriola: Actually goToEnter1Pin and goToEnter2Pin are both used. In Swift, calling the latter looks like:

```
self.go(toEnter2Pin: pin)
```

BlockchainTests/Pin/PinPresenterTests.swift
bchrisarriola: Ah, will need to mock the response of `wallet.encrypt()`. Will fix.

Third Party/PinEntry/Classes/PEPinEntryController.h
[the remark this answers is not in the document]
woudini: Sorry I must have confused it with `didEnter1Pin`

woudini:
Saw a really strange case while creating a PIN after creating a wallet and making errors with it (PINs do not match + common PIN) where the PIN showed again when the email reminder was dismissed. The PIN did go away after though. This bug have have existed for a while.

bchrisarriola:
> Saw a really strange case while creating a PIN after creating a wallet and making errors with it (PINs do not match + common PIN) where the PIN showed again when the email reminder was dismissed. The PIN did go away after though, and I'm having trouble reproducing it. This bug may have existed for a while.

Strange. Let me see if I can repro

PR #347, opened 2018-07-24
refactor(KYC): update UI skeleton

achtenhagen:
Highlights in this PR:

- Updates the UI skeleton to use internal access levels and adds `KYC` prefix to class names.


Blockchain/KYC/KYCCountrySelectionController.swift
```diff
@@ -0,0 +1,65 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.navigationItem.title = "Select Your Country"
+    }
+
+    // MARK: - Actions
+
+    @IBAction func primaryButtonTapped(_ sender: Any) {
+        self.performSegue(withIdentifier: segueIdentifier!, sender: self)
```
thisisalexmcgregor: Do we need `self.` here?

Blockchain/KYC/KYCAccountStatusController.swift
```diff
@@ -8,34 +8,34 @@
...
-        setUpInterfaceFor(accountStatus)
+        accountStatus = .inReview
+        setUpInterfaceFor(accountStatus!)
     }
 
-    override public func didReceiveMemoryWarning() {
-        super.didReceiveMemoryWarning()
+    // MARK: - Methods
+
+    func setAccountStatus(_ accountStatus: AccountStatus) {
```
bchrisarriola: This seem unnecessary since the field `accountStatus` is already publicly exposed.

Blockchain/KYC/KYCAccountStatusController.swift
```diff
@@ -8,34 +8,34 @@
...
 
     // MARK: - Properties
 
     enum AccountStatus {
-        case inReview, approved, failed
+        case approved, failed, inReview
     }
 
-    var accountStatus: AccountStatus = .inReview
+    var accountStatus: AccountStatus?
```
bchrisarriola: I would add a didSet here which calls into setUpInterfaceFor()

```
var accountStatus: AccountStatus? {
    didSet {
        setUpInterfaceFor(accountStatus)
    }
}
```

Blockchain/KYC/KYCAccountStatusController.swift
```diff
@@ -48,18 +48,21 @@ public class AccountStatusController: OnboardingController {
             We had some trouble verifying your account with the documents provided.
             Our support team will contact you shortly to help you with this issue.
             """
+        case .inReview:
+            titleLabel.text = "Account In Review"
+            imageView.image = UIImage(named: "AccountInReview")
+            descriptionLabel.text = "Great job - you are now done. Your account should be approved in minutes."
+            primaryButton.setTitle("Notify Me", for: .normal)
```
bchrisarriola: What about adding computed properties to the enum `AccountStatus`?

e.g.

```
enum AccountStatus {
    // ...
   var titleLabel: String {
       switch self {
           case .failed: return "verification failed" // etc.
       }
   }
   // ...other props
}
```

Blockchain/KYC/KYCAccountStatusController.swift
```diff
@@ -8,34 +8,34 @@
...
+    var accountStatus: AccountStatus?
 
     // MARK: - View Lifecycle
 
-    override public func viewDidLoad() {
+    override func viewDidLoad() {
         super.viewDidLoad()
-        setUpInterfaceFor(accountStatus)
+        accountStatus = .inReview
+        setUpInterfaceFor(accountStatus!)
```
thisisalexmcgregor: Force unwrap? If it's required why make it optional? Maybe make it `.failed` as the default?

Blockchain/KYC/KYCAccountStatusController.swift
```diff
@@ -8,34 +8,34 @@
...
 
-    override public func didReceiveMemoryWarning() {
-        super.didReceiveMemoryWarning()
+    // MARK: - Methods
+
+    func setAccountStatus(_ accountStatus: AccountStatus) {
+        self.accountStatus = accountStatus
     }
 
+    // MARK: - Private Methods
```
thisisalexmcgregor: 👍 Thanks for the `// MARK`'s. Helps a lot with readability.

Blockchain/KYC/KYCOnboardingController.swift
```diff
@@ -24,31 +24,31 @@ open class OnboardingController: UIViewController & OnboardingNavigation {
     @IBOutlet var imageView: UIImageView!
     @IBOutlet var titleLabel: UILabel!
     @IBOutlet var descriptionLabel: UILabel!
-    @IBOutlet public var primaryButton: PrimaryButton!
+    @IBOutlet var primaryButton: PrimaryButton!
 
     // MARK: - View Lifecycle
 
-    override open func viewDidLoad() {
```
thisisalexmcgregor: We can nix these lifecycle functions since they're not implementing anything.

Blockchain/KYC/KYCCountrySelectionController.swift
achtenhagen: Nope, will remove.

Blockchain/KYC/KYCAccountStatusController.swift
achtenhagen: Good idea, will implement.

Blockchain/KYC/KYCAccountStatusController.swift
achtenhagen: Agreed that this is a better way to encapsulate the data. Now `setUpInterfaceFor` is solely responsible for assigning the values to the UI without implementing any logic of its own.
Thanks for the suggestion!

Blockchain/KYC/KYCOnboardingController.swift
achtenhagen: Yes I mostly left these there because the designs are still changing. This PR was by no means a final draft of the code.

Blockchain/KYC/KYCVerifyAccountController.swift
```diff
@@ -57,30 +50,23 @@ open class VerifyAccountController: OnboardingController {
         documentDialog.addAction(identityCardAction)
         documentDialog.addAction(residencePermitCardAction)
         documentDialog.addAction(cancelAction)
+
+        self.present(documentDialog, animated: true)
     }
 
     private func didSelect(_ document: DocumentType) {
```
thisisalexmcgregor: Doesn't look like we're using `document` here. Do we plan to?

Blockchain/KYC/KYCVerifyAccountController.swift
```diff
@@ -57,30 +50,23 @@ open class VerifyAccountController: OnboardingController {
         documentDialog.addAction(identityCardAction)
         documentDialog.addAction(residencePermitCardAction)
         documentDialog.addAction(cancelAction)
+
+        self.present(documentDialog, animated: true)
```
thisisalexmcgregor: No need for `self.`

Blockchain/KYC/KYCVerifyAccountController.swift
achtenhagen: Yes, the designs currently depict that this information is passed to the Onfido flow. How it works exactly I don't know until we sort out the SDK integration.

bchrisarriola:
Code looks good but why is commit 87c7c94 asl in this PR?

PR #349, opened 2018-07-25
 feat(Logger): IOS-1116 Creating Logger class.

bchrisarriola:
### Summary

Created a `Logger` class for consolidating all logging into one class. Logging can now be done through using any of the following methods depending on severity: `debug()`, `info()`, `warning()`, `error()`.

Example usage:
```
Logger.shared.debug("A debug statement.")
```

The logged statement will then be displayed on the console which displays a timestamp, the file, method, and line number that invoked the log statement:

screen shot 2018-07-24 at 5 59 10 pm

The `Logger` is also configured differently depending if the build is configured as DEBUG (i.e. no such print statements will be done on release builds).

### Context
A performance problem we have right now is that we are explicitly calling the Swift method `print()` whenever we want to print out information for debugging/error. Although the Swift compiler on occasion can optimize this and remove these calls, that's not always the case. With this new `Logger` class, we can hide ugly `#if DEBUG` calls inside this class.

### Extensibility
In the future, if we also want to add different log destinations (e.g. to file, to a remote server, etc.) we can easily do so by adding another destination in the `Logger` lazy initializer.

...still need to replace NSLog calls but wanted to do a PR to get input before I get too far.

Inspired by: https://github.com/SwiftyBeaver/SwiftyBeaver


bchrisarriola:
Ah, missed a file. Will update

woudini:
Sorry one more thing - `Logging.swift` seems to be in Rx.xcodeproj, and looks like it should be in the `Logging` directory with the correct Target Memberships (Blockchain and BlockchainTests)

bchrisarriola:
> Logging.swift seems to be in Rx.xcodeproj, and looks like it should be in the Logging directory with the correct Target Memberships (Blockchain and BlockchainTests)

That is a different Logging file not created by us and instead provided by the Rx project

woudini:
Oh I see, I got it confused with the directory.

PR #359, opened 2018-07-30
refactor(Settings): Use Autolayout and Swift for Settings

jstxx:
Migrates Settings and associated controllers to Swift, converts programmatic UITableView creation to storyboards. Also makes some abstractions and structured data. Addresses feedback from https://github.com/blockchain/My-Wallet-V3-iOS/pull/332


woudini:
For some reason the table view controller is empty and the navigation controller isn't dismissible with the top right 'X' button

jstxx:
There must have been a bad merge/rebase, looking into it.

PR #360, opened 2018-07-30
feat(KYC): IOS-1052 1st pass for KYC SMS verification step

bchrisarriola:
#### Summary

Adding business logic for the SMS verification step. Still a lot of TODOs but this [incremental] progress can be merged/reviewed.

#### TODOs 

There are still a lot of TODOs since there are several parts of this flow that are still TBD. Namely, the following:
* the `PUT /users/{userId}/mobile` endpoint for updating a user's information is still in development
* the `userId` needs to be passed into `KYCVerifyPhoneNumberController` from the previous step (previous step is in development... maybe pass that information in `prepare(for segue: UIStoryboardSegue, sender: Any?)`)
* the presentation of how the verification code should be accepted is still being decided (i.e. should an alert be presented? should another view controller be presented?)


Blockchain/KYC/KYCCountrySelectionDataSource.swift
```diff
@@ -25,13 +25,14 @@ final class KYCCountrySelectionDataSource: NSObject, UIPickerViewDataSource {
     }
 
     func fetchListOfCountries() {
-        _ = KYCNetworkRequest(get: .listOfCountries, success: { responseData in
```
jstxx: FWIW I think having a discardable here is Swiftier and doesn't require you to think about the lifecycle of the request

bchrisarriola: Good point about not thinking about the lifecycle. Much of the changes I made here were to remove some code duplication in the initializers, let me revert this but still try to remove the duplication as much as possible.

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -15,8 +15,9 @@ final class KYCNetworkRequest {
     typealias TaskFailure = (HTTPRequestError) -> Void
 
     // TODO: read from .xcconfig
-    fileprivate let rootUrl = "https://api.dev.blockchain.info/nabu-app"
+    fileprivate static let rootUrl = "https://api.dev.blockchain.info/nabu-app"
```
achtenhagen: Could you explain why this needs to be `static`?

thisisalexmcgregor: It's a constant. I usually do the same for things like constraint values, e.g. `fileprivate static let verticalPadding: CGFloat = 8.0`

achtenhagen: Well it's a constant even without `static` 😛, but I can see the benefits of making it a type property since we want **all instances** of `KYCNetworkRequest` to have the same value for `rootUrl`

achtenhagen: But looking at how the class is declared as `final` and the scope of `rootUrl` is `fileprivate`, there is no way for an instance of `KYCNetworkRequest` to change this value anyways, so it is deemed unnecessary and should be reverted.

bchrisarriola: The reason I changed this to `static` is because these initializers are convenience initializers and need to call into `super.init`. At the point `rootUrl` is accessed, the instance hasn't been created yet.

achtenhagen: Okay there is nothing that can be done there if we're using `convenience init`. Disregard request to change it back.

bchrisarriola: Cool 👍 

I added `convenience init` for code sharing since these requests have very similar initializations.

PR #361, opened 2018-07-31
refactor(KYC): make network calls reusable

achtenhagen:
Highlights in this PR:

- Addresses reusable networking code issue from #358 by adding `KYCNetworkRequest`.
- Simplifies how data is fetched in the KYC flow.
- Deprecates the use of the *delegate* pattern for making network requests.
- Restructures error handling for `KYCNetworkRequest`.
- Updates networking code for country selection screen.

### DIFF
--- a/Blockchain/KYC/KYCNetworkRequest.swift
+++ b/Blockchain/KYC/KYCNetworkRequest.swift
@@ -1,0 +1,101 @@
+//
+//  KYCNetworkRequest.swift
+//  Blockchain
+//
+//  Created by Maurice A. on 7/30/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+/// Handles network requests for the KYC flow
+final class KYCNetworkRequest {
+
+    typealias TaskSuccess = (Data) -> Void
+    typealias TaskFailure = (HTTPRequestError) -> Void
+
+    fileprivate let rootUrl = "https://api.dev.blockchain.info/nabu-app"
+    private let timeoutInterval = TimeInterval(exactly: 30)!
+
+    // swiftlint:disable nesting
+    struct KYCEndpoints {
+        enum GET: String {
+            case credentials = "/kyc/credentials"
+            case credentialsForProvider = "/kyc/credentials/provider"
+            case healthCheck = "/healthz"
+            case listOfCountries = "/countries?filter=eea"
+            case nextKYCMethod = "/kyc/next-method"
+            case users, userDetails = "/users"
+        }
+
+        enum POST: String {
+            case registerUser = "/users"
+            case verifications = "/verifications"
+            case submitVerification = "/kyc/verifications"
+        }
+
+        enum PUT: String {
+            case updateUserDetails = "/users"
+        }
+    }
+    // swiftlint:enable nesting
+
+    // MARK: - Initialization
+
+    /// HTTP GET Request
+    init(get url: KYCEndpoints.GET, success: @escaping TaskSuccess, error: @escaping TaskFailure) {
+        var request = URLRequest(url: URL(string: rootUrl + url.rawValue)!)
+        request.httpMethod = "GET"
+        request.addValue("application/json", forHTTPHeaderField: "Accept")
+        request.timeoutInterval = timeoutInterval
+        doTask(with: request, success, error)
+    }
+
+    /// HTTP POST Request
+    init(post url: KYCEndpoints.POST, parameters: [String: String], success: @escaping TaskSuccess, error: @escaping TaskFailure) {
+        let postBody = parameters.reduce("", { initialResult, nextPartialResult in
+            let delimeter = initialResult.count > 0 ? "&" : ""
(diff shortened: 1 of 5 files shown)


Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -0,0 +1,101 @@
...
+    init(get url: KYCEndpoints.GET, success: @escaping TaskSuccess, error: @escaping TaskFailure) {
+        var request = URLRequest(url: URL(string: rootUrl + url.rawValue)!)
+        request.httpMethod = "GET"
+        request.addValue("application/json", forHTTPHeaderField: "Accept")
+        request.timeoutInterval = timeoutInterval
+        doTask(with: request, success, error)
+    }
+
+    /// HTTP POST Request
+    init(post url: KYCEndpoints.POST, parameters: [String: String], success: @escaping TaskSuccess, error: @escaping TaskFailure) {
```
bchrisarriola: Should parameters be of type `[String: Any]`? We also have the typealias `URLParameters`

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -0,0 +1,101 @@
...
+
+import Foundation
+
+/// Handles network requests for the KYC flow
+final class KYCNetworkRequest {
+
+    typealias TaskSuccess = (Data) -> Void
+    typealias TaskFailure = (HTTPRequestError) -> Void
+
+    fileprivate let rootUrl = "https://api.dev.blockchain.info/nabu-app"
```
bchrisarriola: Add a TODO that this should be read from .xcconfig?

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        enum POST: String {
+            case registerUser = "/users"
+            case verifications = "/verifications"
+            case submitVerification = "/kyc/verifications"
+        }
+
+        enum PUT: String {
+            case updateUserDetails = "/users"
```
bchrisarriola: This is fine for now but there might be some other URL path arguments:

e.g.
```
/users/{userId}/somepath
```

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    // MARK: - Initialization
+
+    /// HTTP GET Request
+    init(get url: KYCEndpoints.GET, success: @escaping TaskSuccess, error: @escaping TaskFailure) {
+        var request = URLRequest(url: URL(string: rootUrl + url.rawValue)!)
+        request.httpMethod = "GET"
+        request.addValue("application/json", forHTTPHeaderField: "Accept")
+        request.timeoutInterval = timeoutInterval
+        doTask(with: request, success, error)
```
bchrisarriola: Did you imagine that `KYCNetworkRequest` will also handle image uploading (or is that something that the Onfido SDK will handle)? If it can potentially have large payloads, it would be good to defer to the caller to invoke `resume()` on the underlying `URLSessionDataTask` in the scenario that it also may want to cancel/suspend the data task.

achtenhagen: Great question, according to their docs, the SDK handles uploading for us, we are just responsible for creating and managing the identity checks themselves.

Blockchain/KYC/KYCNetworkRequest.swift
achtenhagen: Since the raw value for enum cases must be a literal,
```
init(put url: KYCEndpoints.PUT, parameters: [String: String], success: @escaping TaskSuccess, error: @escaping TaskFailure)
```
could be modified to accept a userId as one option.

Blockchain/KYC/KYCNetworkRequest.swift
bchrisarriola: Good point about safety. `[String: String]` works.

Blockchain/KYC/KYCNetworkRequest.swift
bchrisarriola: I think the URL should probably be more configurable than that. Let's go with this for now and we can improve it once we merge PR #363 from @thisisalexmcgregor

achtenhagen: Sounds good.

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// HTTP POST Request
+    init(post url: KYCEndpoints.POST, parameters: [String: String], success: @escaping TaskSuccess, error: @escaping TaskFailure) {
+        let postBody = parameters.reduce("", { initialResult, nextPartialResult in
```
jstxx: Didn't see this before it was merged, however my $0.02 is that parameters should actually be a `URLComponent.queryItems` which handles the ?, & under the hood so we can get rid of the reduce fn entirely.

thisisalexmcgregor: Actually I submitted (and then merged) a PR that adds a helper function in `URLs.swift` for building endpoint URLs.

PR #363, opened 2018-07-31
refactor(KYC): URL convenience function

thisisalexmcgregor:
## Description

Adds a simple function to `URLs.swift` to make URL endpoint generation a bit easier.

The call site would look like this:

`let parameters = ["method": "do.a.thing", "api_key": Blockchain.key]`

`guard let url = URL.endpoint(Blockchain.baseURL, pathComponents: ["path", "component"], queryParameters: parameters) else { return }`

### DIFF
--- a/Blockchain/Extensions/URLs.swift
+++ b/Blockchain/Extensions/URLs.swift
@@ -49,9 +49,7 @@ extension URL {
 
         guard var components: URLComponents = URLComponents(url: mutableBaseURL as URL, resolvingAgainstBaseURL: false) else { return nil }
 
-        if queryItems.count > 0 {
-            components.queryItems = queryItems
-        }
+        components.queryItems = queryItems
 
         return components.url
     }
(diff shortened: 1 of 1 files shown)


Blockchain/Extensions/URLs.swift
```diff
@@ -18,4 +18,41 @@ extension URL {
...
+                if let value = queryParameters[keyValue] {
+                    let queryItem = URLQueryItem(name: keyValue, value: value)
+                    queryItems.append(queryItem)
+                }
+            }
+        }
+
+        guard var components: URLComponents = URLComponents(url: mutableBaseURL as URL, resolvingAgainstBaseURL: false) else { return nil }
+
+        if queryItems.count > 0 {
```
bchrisarriola: Can this check be removed?

thisisalexmcgregor: Derp. Yeah. 

- [ ] Remove unnecessary `if` check.

thisisalexmcgregor: @bchrisarriola Done: 4cf3273

PR #364, opened 2018-07-31
feat(KYC): improve personal details screen

achtenhagen:
Highlights in this PR:

- Adds the birthdate picker to the personal information screen.
- Removes deprecated storyboard files and classes.


Blockchain/KYC/KYCPersonalDetailsController.swift
```diff
@@ -13,21 +13,80 @@ final class KYCPersonalDetailsController: UIViewController, KYCOnboardingNavigat
...
 
-    var segueIdentifier: String?
+    private let birthdatePicker: UIDatePicker!
+    var segueIdentifier: String? = "verifyMobileNumber"
 
     // MARK: - IBOutlets
 
+    @IBOutlet var firstNameField: UITextField!
+    @IBOutlet var lastNameField: UITextField!
+    @IBOutlet var birthdateField: UITextField!
```
thisisalexmcgregor: `private`?

Blockchain/KYC/KYCPersonalDetailsController.swift
```diff
@@ -13,21 +13,80 @@ final class KYCPersonalDetailsController: UIViewController, KYCOnboardingNavigat
...
+        birthdatePicker.maximumDate = Date()
+        super.init(coder: aDecoder)
+    }
+
+    // MARK: - Private Methods
+
+    private func setUpBirthdatePicker() {
+        let toolBar = UIToolbar()
+        toolBar.barStyle = .default
+        toolBar.isTranslucent = true
```
thisisalexmcgregor: `default` and `true` are the default values here so, you don't need this.

Just curious, why `sizeToFit()` after?

Blockchain/KYC/KYCPersonalDetailsController.swift
```diff
@@ -13,21 +13,80 @@ final class KYCPersonalDetailsController: UIViewController, KYCOnboardingNavigat
...
+        toolBar.sizeToFit()
+
+        let doneButton = UIBarButtonItem(title: "Done", style: .plain, target: self, action: #selector(submitBirthdate))
+        let spaceButton = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
+        let cancelButton = UIBarButtonItem(title: "Cancel", style: .plain, target: self, action: #selector(hideBirthdatePicker))
+        toolBar.setItems([cancelButton, spaceButton, doneButton], animated: false)
+        toolBar.isUserInteractionEnabled = true
```
thisisalexmcgregor: This is `true` by default.

Blockchain/KYC/KYCPersonalDetailsController.swift
```diff
@@ -13,21 +13,80 @@ final class KYCPersonalDetailsController: UIViewController, KYCOnboardingNavigat
...
     // MARK: - IBOutlets
 
+    @IBOutlet private var firstNameField: UITextField!
+    @IBOutlet private var lastNameField: UITextField!
+    @IBOutlet private var birthdateField: UITextField!
     @IBOutlet var primaryButton: PrimaryButton!
 
+    override func viewDidLoad() {
+        setUpBirthdatePicker()
+        birthdateField.inputView = birthdatePicker
```
thisisalexmcgregor: `birthdateField.inputView = birthdatePicker` should probably be in `setUpBirthdatePicker()` no? 🤷‍♂️

Blockchain/KYC/KYCPersonalDetailsController.swift
achtenhagen: I was unsure whether the frame will always be set properly by the system, I'll remove the default values.

Blockchain/KYC/KYCPersonalDetailsController.swift
achtenhagen: Will remove.

achtenhagen:
That is correct. AFAIK there is no straightforward of moving between UITextfields using the keyboard in iOS.

Blockchain/KYC/KYCPersonalDetailsController.swift
achtenhagen: `birthdateField` won't be available during initialization, so it needs to be in `viewDidLoad`.

Blockchain/KYC/KYCPersonalDetailsController.swift
achtenhagen: Update: `sizeToFit()` is required otherwise the frame is empty.

thisisalexmcgregor:
RE: Jumping to next `UITextField`, I believe you need to hook into the `UITextFieldDelegate` in `textFieldShouldReturn`

achtenhagen:
Yes, I think the annoyance is that every textfield is associated with that method so people use tags to get around it. Will investigate best way to implement it.

thisisalexmcgregor:
I don't think you need tags actually, you can just do:
`if textField == firstNameField { // Do a thing }`

achtenhagen:
Confirmed, works. Thanks for the tip!

Blockchain/KYC/KYCPersonalDetailsController.swift
```diff
@@ -13,21 +13,80 @@ final class KYCPersonalDetailsController: UIViewController, KYCOnboardingNavigat
...
+                actions: [
+                    UIAlertAction(title: "OK, I understand", style: .default, handler: { _ in
+                        // TODO: exit KYC flow and send user back to dashboard
+                    })
+                ],
+                in: self
+            )
+            return
+        }
+        performSegue(withIdentifier: segueIdentifier!, sender: self)
```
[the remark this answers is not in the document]
achtenhagen: Is this necessary given that the string is defined in the file, thus guaranteeing it will never be nil?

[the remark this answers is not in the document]
thisisalexmcgregor: Force unwrapping is generally a code smell.

PR #365, opened 2018-07-31
refactor(Exchange): IOS-1129 Move views out of ExchangeCreateViewController

woudini:
Goal is to add modularity for Homebrew Shapeshift implementation

Moved the `setupViews` method, view properties, and view-related methods to a new file `ExchangeCreateView.swift`.


woudini:
Will consider replacing view code with a xib if there is time before designs are finalized.

PR #366, opened 2018-07-31
dep(CocoaPods): Add dependency management

jstxx:
Adds cocoapods for dependency management and adds the following Pods:

- Alamofire
- Charts
- RxSwift
- Onfido


jstxx:
✅

PR #372, opened 2018-08-03
feat(KYC): IOS-1133 Adding mobile confirmation code screen

bchrisarriola:
#### Summary:
* Added screen for entering the received confirmation code when going through the mobile verification process during KYC
* Added a few files in the `BlockchainTests` target related to addresses in KYC since unit tests were failing
* Added some more unit tests for this flow

#### Screenshots:
simulator screen shot - iphone x - 2018-08-02 at 16 35 22

#### Notes:
* Can't quite test the flow yet since `POST /users/{userId}/mobile` is still in development


bchrisarriola:
Please hold off on reviewing. I need to resync my branch with `kyc`

Blockchain/KYC/Mobile Number/KYCConfirmPhoneNumberController.swift
```diff
@@ -0,0 +1,108 @@
...
+    @IBOutlet private var layoutConstraintBottomButton: NSLayoutConstraint!
+
+    var phoneNumber: String = "" {
+        didSet {
+            guard isViewLoaded else { return }
+            labelPhoneNumber.text = phoneNumber
+        }
+    }
+
+    var userId: String = ""
```
thisisalexmcgregor: I'm guessing this is going to get set somewhere else (outside of `KYCConfirmPhoneNumberController`) since it's passed in as a parameter below. Should it be optional in that case?

thisisalexmcgregor:
For `startVerificationSuccess()`, well, up until it's hit, I'm guessing we want to disable the `resend` button until that comes back successfully. 

Actually does the `BCFadeView` prevent user interaction? Does `KYCEnterPhoneNumberController` get presented only after a code is sent successfully?

Blockchain/KYC/Mobile Number/KYCConfirmPhoneNumberController.swift
bchrisarriola: Yeah, probably better for it to be optional. Let me change that

bchrisarriola:
> For startVerificationSuccess(), well, up until it's hit, I'm guessing we want to disable the resend button until that comes back successfully.

`startVerificationSuccess()` should only be invoked after the call to `POST /nabu-app/users/{userId}/mobile` completes. So by the time they see the next screen (i.e. the `KYCConfirmPhoneNumberController`), the SMS message should already be on its way.

> Actually does the BCFadeView prevent user interaction? 

Good question! Yep, looks like user interaction is enabled on `BCFadeView`. But since it takes up the whole screen, users shouldn't be able to tap behind it (also tapping on it won't dismiss the view). However, we should probably explicitly disable user interaction on that view.

> Does KYCEnterPhoneNumberController get presented only after a code is sent successfully?

If you meant `KYCConfirmPhoneNumberController` (`KYCEnterPhoneNumberController` is the view controller where we collect the user's phone number, not the code), then yes, that only gets presented after the code is sent.

thisisalexmcgregor:
Gotcha, makes sense. Should we always be using that loading indicator vs. the native one?

Blockchain/KYC/Mobile Number/KYCConfirmPhoneNumberController.swift
bchrisarriola: Done 5fe9a686

thisisalexmcgregor: ✅

bchrisarriola:
> Should we always be using that loading indicator vs. the native one?
I'm using it mostly for consistency with the rest of the app, but I'm open to revisiting it. On the upside, since `LoadingViewPresenter` is our go to component for indicating a loading state, making that change globally shouldn't be too crazy.

PR #379, opened 2018-08-06
chore(fix):Fix unit tests

jstxx:
## Objective

*Briefly summarize the object of this pull request.*
Fixes unit tests
## Description

*Describe in detail how you achieved the objective*
Updated all target memberships and dependencies for BlockchainTests, updated linked libraries.
## How to Test

*Describe how a reviewer can verify your changes. This helps with those who are unfamiliar with the feature.*
Build the project, ensure all tests pass.
## Screenshot
image

*If the change is UI related, paste a screenshot here.*

## Related Information

* Ticket Number:

## Merge Checklist

- [x] The PR uses a title supported by [.changelogrc](https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/.changelogrc#L6...L69).
- [x] Areas of technical debt are marked with a `// TICKET:` comment that includes a ticket number.
- [x] All unit tests pass.
- [x] New files added are within the correct directory. (e.g. if a file is required for unit tests to compile, be sure it is added to the tests target.)

PR #381, opened 2018-08-07
feat(Networking): IOS-1137 add socket manager singleton

woudini:
## Objective

Create a singleton for Homebrew Exchange to initially use and later to support BTC/ETH/BCH when `SocketRocket` websockets are removed from the `Wallet` class.

## Description

Created singleton `SocketManager` in Networking directory.
Created `SocketMessage.swift` to contain `ExchangeSocketMessage` struct and other message types in the future.

## How to Test

N/A, not yet implemented by the Exchange due to unavailable endpoints. Also looking for early feedback. Unit tests won't pass until https://github.com/blockchain/My-Wallet-V3-iOS/pull/379 is merged

## Screenshot

N/A

## Related Information

* Ticket Number: IOS-1137

## Merge Checklist

- [ ] All unit tests pass.


Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+
+    private let exchangeSocket: WebSocket
+    // Add the following properties when removing the websocket from Wallet class
+    // private let btcSocket: Websocket
+    // private let ethSocket: Websocket
+    // private let bchSocket: Websocket
+
+    // MARK: - Initialization
+    init() {
+        self.exchangeSocket = WebSocket(url: URL(string: BlockchainAPI.Nabu.quotes)!)
```
jstxx: It would be really nice if we could init the websocket URL construct from the get go rather than forming it adhoc from a bare string.

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+    // MARK: - Public methods
+    func send(message: SocketMessage) {
+        /* Creating a SocketMessage to send
+         let message = SocketMessage(parameterOne: 0, parameterTwo: "parameterTwo", parameterThree: "parameterThree")
+         let encodedData = try JSONEncoder().encode(message)
+         let string = String(data: encodedData, encoding: .utf8)
+         */
+
+        if message is ExchangeSocketMessage {
+            guard exchangeSocket.isConnected else {
```
jstxx: Any reason not to combine this if/guard? Then you should be able to remove the `message as! ExchangeSocketMessage` force cast

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+                pendingExchangeMessages.append(message as! ExchangeSocketMessage)
+                return
+            }
+
+            let onError: () -> Void = {
+                Logger.shared.error("Could send websocket message as string")
+            }
+
+            do {
+                let encodedData = try JSONEncoder().encode(message as! ExchangeSocketMessage)
```
jstxx: Having an extension on Encodable makes this more readable. 

so from: `let encodedData = try JSONEncoder().encode(message as! ExchangeSocketMessage)`
to: `let encodedData = try ExchangeSocketMessage.encode()`


```
extension Encodable {
    func encode() throws -> Data {
        let encoder = JSONEncoder()
        encoder.outputFormatting = .prettyPrinted
        return try encoder.encode(self)
    }
}
```

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+            Logger.shared.error("Could not form SocketMessage object from string")
+        }
+
+        if socket == self.exchangeSocket {
+            do {
+                guard let data = text.data(using: .utf8) else {
+                    onError()
+                    return
+                }
+                let socketMessage = try JSONDecoder().decode(ExchangeSocketMessage.self, from: data)
```
jstxx: Same thing here for decoded

` let socketMessage = try JSONDecoder().decode(ExchangeSocketMessage.self, from: data)` becomes
`let socketMessage = try ExchangeSocketMessage.decode(data: data)`

```
extension Decodable {
    static func decode(data: Data) throws -> Self {
        let decoder = JSONDecoder()
        return try decoder.decode(Self.self, from: data)
    }
}
```

Podfile
```diff
@@ -12,6 +12,7 @@ target 'Blockchain' do
 	pod 'Charts'
 	pod 'RxSwift',    '~> 4.0'
 	pod 'RxCocoa',    '~> 4.0'
+        pod 'Starscream', '~> 3.0.2'
```
bchrisarriola: Nit: fix spacing

Blockchain/Models/SocketMessage.swift
```diff
@@ -0,0 +1,24 @@
...
+//  SocketMessage.swift
+//  Blockchain
+//
+//  Created by kevinwu on 8/3/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+typealias SocketMessage = Codable
```
bchrisarriola: This might be better as a protocol that conforms to `Codable` rather than a `typealias`. In the future we may want to add properties/functions that are shared across all `SocketMessage` classes/structs.

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+
+    // MARK: - Initialization
+    init() {
+        self.exchangeSocket = WebSocket(url: URL(string: BlockchainAPI.Nabu.quotes)!)
+        self.webSocketMessageSubject = PublishSubject<SocketMessage>()
+    }
+
+    /// Data providers should suscribe to this and filter (e.g., { $0 is ExchangeSocketMessage })
+    var webSocketMessageObservable: Observable<SocketMessage> {
+        return webSocketMessageSubject
```
bchrisarriola: Should call `.asObservable()`

i.e.
```
return webSocketMessageSubject.asObservable()
```

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private let webSocketMessageSubject: PublishSubject<SocketMessage>
+    private lazy var pendingExchangeMessages = [ExchangeSocketMessage]()
```
bchrisarriola: This doesn't need to be declared as `lazy`. That's only necessary when instantiation is expensive and should be done lazily or when a reference to `self` is needed. In this case, neither of those apply. 

Also, I think we can make this more generic (i.e. should be of type `SocketMessage` instead).

i.e.
```
private var pendingSocketMessages = [SocketMessage]()
```

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+         let message = SocketMessage(parameterOne: 0, parameterTwo: "parameterTwo", parameterThree: "parameterThree")
+         let encodedData = try JSONEncoder().encode(message)
+         let string = String(data: encodedData, encoding: .utf8)
+         */
+
+        if message is ExchangeSocketMessage {
+            guard exchangeSocket.isConnected else {
+                Logger.shared.info("Exchange socket is not connected - will append message to pending messages")
+                exchangeSocket.connect()
+                pendingExchangeMessages.append(message as! ExchangeSocketMessage)
```
bchrisarriola: This should go one line up before we try to connect to the `exchangeSocket`. This is a chance that `exchangeSocket` might get connected before this happens.

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+        /* Creating a SocketMessage to send
+         let message = SocketMessage(parameterOne: 0, parameterTwo: "parameterTwo", parameterThree: "parameterThree")
+         let encodedData = try JSONEncoder().encode(message)
+         let string = String(data: encodedData, encoding: .utf8)
+         */
+
+        if message is ExchangeSocketMessage {
+            guard exchangeSocket.isConnected else {
+                Logger.shared.info("Exchange socket is not connected - will append message to pending messages")
+                exchangeSocket.connect()
```
bchrisarriola: Call `connect(socketType: .exchange)`?

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+    }
+
+    func connect(socketType: SocketType) {
+        switch socketType {
+        case .exchange: {
+            self.exchangeSocket.advancedDelegate = self
+            self.exchangeSocket.connect()
+        }()
```
bchrisarriola: Remove parens, should just be terminated by `}`.

Blockchain/Network/SocketManager.swift
woudini: Sorry I'm not clear on what you mean here

Blockchain/Models/SocketMessage.swift
bchrisarriola: For example, we can add `SocketType`.

```
protocol SocketMessage: Codable {
    var type: SocketType { get }
}
```

Blockchain/Network/SocketManager.swift
woudini: Was planning to add onto this eventually with something like `else if message is BtcSocketMessage` and so on. Would a `guard` still be applicable in that future case?

Blockchain/Network/SocketManager.swift
jstxx: I'm saying `BlockchainAPI.Nabu.quotes` should already be a fully formed/qualified URL as opposed to a string by the time it gets to this init method

Blockchain/Network/SocketManager.swift
bchrisarriola: If the `type` property is added to `SocketMessage`, we can just do something like:

```
switch message.type {
    case .exchange:
        tryToSend(message: SocketMessage, socket: exchangeSocket)
    case .btc:
        tryToSend(message: SocketMessage, socket: exchangeSocket)
}
```

...and have a function:

```
private func tryToSend(message: SocketMessage, socket: WebSocket) {
    guard socket.isConnected else {
        // do stuff...
        return
    }
    // do stuff...
}
```

Blockchain/Network/SocketManager.swift
woudini: Moved to a new public method `setupSocket(socketType: SocketType, url: URL)`

Blockchain/Network/SocketManager.swift
```diff
@@ -0,0 +1,128 @@
...
+        let onSuccess: (SocketMessage) -> Void = { socketMessage in
+            self.webSocketMessageSubject.onNext(socketMessage)
+        }
+
+        guard let data = text.data(using: .utf8) else {
+            onError()
+            return
+        }
+
+        Quote.tryToDecode(data: data, onSuccess: onSuccess, onError: onError)
```
woudini: Tried to make the decoding here cleaner but couldn't figure out how to get past this error.
screen shot 2018-08-08 at 11 15 57 am

bchrisarriola: Trying to parse `data` into every single type of `SocketMessage` is pretty computationally expensive. It would be best to try to decode first to a JSON (i.e. [String: Any]) and then inspect the "channel" property to see which type it should be decoded to. Don't need to address now

woudini: Added `TODO` comment for when this is implemented with correct socket message parameters

PR #383, opened 2018-08-07
feat(KYC): IOS-1046 - Personal Information Screens & IOS-1141 - Progession

thisisalexmcgregor:
## Objective

**Note:** Marking DNM due to conflict with develop and some minor polish items I'm still addressing. 

This PR includes the new `Personal Information` screen as well as a protocol for handling the `UIProgressView` that should be in all the KYC related screens. 

## Description

In addition to the features described, there's a few areas of code cleanup. 
* Added `ValidationFormView` - This protocol should be used with all `UIViewControllers` that have a `UIScrollView` and a bunch of `ValidationTextFields`. It will handle keeping the selected `UITextField` in the user's field of view when the keyboard is presented as well as checking all the fields' validation state. 
* Added `ValidationDateField` - This is only for when you have a `UITextField` that only takes a date as input. It's not ideal as I'd rather we use `UIPickerView` and configure it with whatever dataSource is necessary but, due to time constraints I'm putting that off. The date is validated with the `validationBlock`. 
* Added`ProgressableView`. I'm assuming this will be tweaked as we get closer to shipping. Per creative, we may end up showing progression per field vs. per screen. 

## How to Test

1. Before building, in the `KYCWelcome` storyboard, set the segue to route to the `KYCPersonalDetails` storyboard. 
2. Build and run
3. Go to the KYC flow.
4. Tap `Apply Now`
5. View the `Personal Details` page. Mess with the fields. 

## Screenshot

**Note:** The `primaryButton` is a bit high. I'd like to offset it based on the keyboards position but haven't gotten to that yet. 

screen shot 2018-08-07 at 4 18 05 pm

## Related Information

* Ticket Number: IOS-1141 & IOS-1046

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/KYCPersonalDetailsController.swift
+++ b/Blockchain/KYC/KYCPersonalDetailsController.swift
@@ -9,74 +9,66 @@
 import UIKit
 
 /// Personal details entry screen in KYC flow
-final class KYCPersonalDetailsController: UIViewController {
+final class KYCPersonalDetailsController: UIViewController, ValidationFormView {
 
-    // MARK: - Properties
+    // MARK: - Public IBOutlets
 
-    private let birthdatePicker: UIDatePicker!
+    @IBOutlet var scrollView: UIScrollView!
 
     // MARK: - IBOutlets
 
-    @IBOutlet private var firstNameField: UITextField!
-    @IBOutlet private var lastNameField: UITextField!
-    @IBOutlet private var birthdateField: UITextField!
-    @IBOutlet var primaryButton: PrimaryButton!
+    @IBOutlet fileprivate var firstNameField: ValidationTextField!
+    @IBOutlet fileprivate var lastNameField: ValidationTextField!
+    @IBOutlet fileprivate var birthdayField: ValidationDateField!
+    @IBOutlet fileprivate var primaryButton: PrimaryButton!
 
-    override func viewDidLoad() {
-        setUpBirthdatePicker()
-        birthdateField.inputView = birthdatePicker
+    // MARK: Public Properties
+
+    var validationFields: [ValidationTextField] {
+        get {
+            return [firstNameField,
+                    lastNameField,
+                    birthdayField]
+        }
     }
 
-    required init?(coder aDecoder: NSCoder) {
-        birthdatePicker = UIDatePicker()
-        birthdatePicker.datePickerMode = .date
-        birthdatePicker.maximumDate = Date()
-        super.init(coder: aDecoder)
+    var keyboard: KeyboardPayload? = nil
+
+    // MARK: Lifecycle
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        setupTextFields()
+        handleKeyboardOffset()
+        setupNotifications()
     }
 
     // MARK: - Private Methods
 
-    private func setUpBirthdatePicker() {
-        let toolBar = UIToolbar()
-        toolBar.sizeToFit()
-        let doneButton = UIBarButtonItem(title: "Done", style: .plain, target: self, action: #selector(submitBirthdate))
-        let spaceButton = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
-        let cancelButton = UIBarButtonItem(title: "Cancel", style: .plain, target: self, action: #selector(hideBirthdatePicker))
-        toolBar.setItems([cancelButton, spaceButton, doneButton], animated: false)
-        birthdateField.inputAccessoryView = toolBar
+    fileprivate func setupTextFields() {
+        firstNameField.returnKeyType = .next
+        firstNameField.contentType = .name
+
+        lastNameField.returnKeyType = .next
+        lastNameField.contentType = .familyName
     }
 
-    @objc private func submitBirthdate() {
-        let dateFormatter = DateFormatter()
-        dateFormatter.dateStyle = .medium
-        dateFormatter.timeStyle = .none
-        birthdateField.text = dateFormatter.string(from: birthdatePicker.date)
-        birthdateField.resignFirstResponder()
-    }
-
-    @objc private func hideBirthdatePicker() {
-        birthdateField.resignFirstResponder()
+    fileprivate func setupNotifications() {
+        NotificationCenter.when(.UIKeyboardWillHide) { [weak self] _ in
+            self?.scrollView.contentInset = .zero
+            self?.scrollView.setContentOffset(.zero, animated: true)
+        }
+        NotificationCenter.when(.UIKeyboardWillShow) { [weak self] notification in
+            let keyboard = KeyboardPayload(notification: notification)
+            self?.keyboard = keyboard
+        }
     }
 
     // MARK: - Actions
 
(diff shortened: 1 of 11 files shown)


Blockchain/KYC/KYCPersonalDetailsController.swift
```diff
@@ -9,74 +9,101 @@
...
-        let toolBar = UIToolbar()
-        toolBar.sizeToFit()
-        let doneButton = UIBarButtonItem(title: "Done", style: .plain, target: self, action: #selector(submitBirthdate))
-        let spaceButton = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
-        let cancelButton = UIBarButtonItem(title: "Cancel", style: .plain, target: self, action: #selector(hideBirthdatePicker))
-        toolBar.setItems([cancelButton, spaceButton, doneButton], animated: false)
-        birthdateField.inputAccessoryView = toolBar
+    fileprivate func setupTextFields() {
+        firstNameField.returnKeyType = .next
+        firstNameField.contentType = .name
```
bchrisarriola: Set `autocapitalizationType` as well? Something that should be exposed in `ValidationTextField`, too.

Blockchain/KYC/KYCPersonalDetailsController.swift
```diff
@@ -85,13 +112,16 @@ final class KYCPersonalDetailsController: UIViewController {
...
-        return false
+extension KYCPersonalDetailsController: UIScrollViewDelegate {
+    func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        // TODO: May not be necessary. 
+        validationFields.forEach({$0.resignFocus()})
     }
 }
+
+extension Date {
+    static let eighteenYears: Date = Calendar.current.date(
```
bchrisarriola: Nice UX 💯

Blockchain/KYC/Views/ProgressableView.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Blockchain
+//
+//  Created by Alex McGregor on 8/7/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+protocol ProgressableView {
+    var progressView: UIProgressView! { get }
+    var barColor: UIColor { get set }
+    var startingValue: Float { get set }
```
bchrisarriola: What do you think about have a `Range<Float>` here instead? For example, if the KYC flow has 3 views that the user needs to go through, the ranges would be:

View 1: [0.0, 0.33],
View 2: [0.33, 0.66],
View 3: [0.66, 1.0]

The function `updateProgress(_ progress: Float)` would then be a value between [0.0, 1.0] and would be set by the view controller as the user fills out fields. Having a range would restrict the view from setting arbitrary values that are outside its specified range.

bchrisarriola: Another thought... it would be good for another component (e.g. `KYCCoordinator`) to set the value of these ranges. This way, view controllers don't "know" where in the KYC flow they are. Also, if the KYC sequence does change, we won't need to update all KYC view controllers, we'd only need to update the coordinator.

Doesn't need to be addressed in this PR since that would probably require a larger architectural change.

thisisalexmcgregor: Yeah agreed, not a fan of having the VC's know about their progression changes. I know creative mentioned the progression may change from screen to screen _or_ from field to field. 

And yeah I think the coordinator could be a good spot for some of this, if not all. Just spit balling but maybe each VC is created like so:

`make(with coordinator: KYCCoordinator) -> KYCViewController`

Then the VC can maybe pass validation events to the coordinator and the coordinator returns a model with progression? 🤷‍♂️

Blockchain/KYC/KYCPersonalDetailsController.swift
```diff
@@ -9,74 +9,101 @@
 import UIKit
 
 /// Personal details entry screen in KYC flow
-final class KYCPersonalDetailsController: UIViewController {
+final class KYCPersonalDetailsController: UIViewController, ValidationFormView, ProgressableView {
```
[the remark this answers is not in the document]
thisisalexmcgregor: Yep, I can break those up.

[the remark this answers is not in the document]
thisisalexmcgregor: Huh. `Extensions must not contain stored properties` Didn't think I'd run into that. Oh well. I'll break it up with better `// MARK` usage

Blockchain/KYC/KYCPersonalDetailsController.swift
thisisalexmcgregor: Done f0d536f

Blockchain/KYC/KYCPersonalDetailsController.swift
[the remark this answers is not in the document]
thisisalexmcgregor: Done f0d536f

thisisalexmcgregor:
PR should be good to go. I'd like to polish up the positioning of the `Next` button on keyboard presentation. Not going to get hung up on that for this PR though.

Blockchain/KYC/KYCPersonalDetailsController.swift
bchrisarriola: Set the `autocapitalizationType` for `firstNameField` and `lastNameField`?

Blockchain/KYC/Views/ProgressableView.swift
bchrisarriola: Yeah that can work, too. Let's mess around with it in another PR 👍

Blockchain/KYC/KYCPersonalDetailsController.swift
thisisalexmcgregor: Dumb question, it's the default if it's not set but does that mean I never have to set it? 

` var autocapitalizationType: UITextAutocapitalizationType = .words {
        didSet {
            textField.autocapitalizationType = autocapitalizationType
        }
    }`

bchrisarriola: Ah, I think initialization won't actually invoke the closure inside `didSet`. But if you set this value inside `ValidationTextField` init method, it should.

thisisalexmcgregor: kk, set it in the `nib` 57fb1fe

Blockchain/KYC/KYCPersonalDetailsController.swift
thisisalexmcgregor: Ah, you mean as the minimumDate on the picker? Not sure. May get feedback from creative about it when we push a build to QA. If we set it as the minimum date on the picker than users that are under 18 may just hit `Next` since they can't enter in their actual birthdate. Might be a legal issue in that case 🤷‍♂️

bchrisarriola: Sorry I meant the value of the `date` property should be initialized to 18 years back. So when you tap on the field, instead of showing Aug. 8 2018, it would default to Aug. 8, 2000. Wdyt?

PR #385, opened 2018-08-08
fix(KYC): Improve UX in KYC confirmation code view

bchrisarriola:
## Objective

Improving UX in the confirmation code view in the KYC flow.

## Description

Improving UX by:
* Using `ValidationTextField` instead of `UITextField`
* Have consistent handling of the CTA in the KYC flow (i.e. it is always enabled, however, if one of the fields are invalid, it will show an error on the `ValidationTextField` and focus)

## How to Test

Launch KYC and go through the different steps until entering your phone #. You won't be able to pass that step though since `userId` isn't passed around yet. To get around that, just replace line 100 in `KYCEnterPhoneNumberController`.

```
self.performSegue(withIdentifier: "verifyMobileNumber", sender: nil)
```

## Screenshot

simulator screen shot - iphone 5s - 2018-08-08 at 12 13 39

## Related Information

* Ticket Number:

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/Mobile Number/KYCConfirmPhoneNumberController.swift
+++ b/Blockchain/KYC/Mobile Number/KYCConfirmPhoneNumberController.swift
@@ -44,7 +44,6 @@ final class KYCConfirmPhoneNumberController: UIViewController, BottomButtonConta
     override func viewDidLoad() {
         super.viewDidLoad()
         // TICKET: IOS-1141 display correct % in the progress view
-        validationTextFieldConfirmationCode.keyboardType = .namePhonePad
         validationTextFieldConfirmationCode.autocapitalizationType = .allCharacters
         labelPhoneNumber.text = phoneNumber
         originalBottomButtonConstraint = layoutConstraintBottomButton.constant
--- a/Blockchain/KYC/Mobile Number/KYCConfirmPhoneNumberController.swift
+++ b/Blockchain/KYC/Mobile Number/KYCConfirmPhoneNumberController.swift
@@ -8,12 +8,9 @@
 
 import Foundation
 
-class KYCConfirmPhoneNumberController: UIViewController {
+final class KYCConfirmPhoneNumberController: UIViewController, BottomButtonContainerView {
 
-    @IBOutlet private var nextButton: PrimaryButton!
-    @IBOutlet private var labelPhoneNumber: UILabel!
-    @IBOutlet private var textFieldConfirmationCode: UITextField!
-    @IBOutlet private var layoutConstraintBottomButton: NSLayoutConstraint!
+    // MARK: Public Properties
 
     var phoneNumber: String = "" {
         didSet {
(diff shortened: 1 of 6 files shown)


Blockchain/KYC/Mobile Number/KYCEnterPhoneNumberController.swift
```diff
@@ -90,26 +88,6 @@ final class KYCEnterPhoneNumberController: UIViewController {
         confirmPhoneNumberViewController.userId = userId
         confirmPhoneNumberViewController.phoneNumber = validationTextFieldMobileNumber.text ?? ""
```
jstxx: Any reason this isn't unwrapped in the guard above?

Blockchain/KYC/Views/BottomButtonContainerView.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    var layoutConstraintBottomButton: NSLayoutConstraint! { get }
+}
+
+extension BottomButtonContainerView where Self: UIViewController {
+
+    /// Sets up this view so that it can respond to keyboard show/hide events.
+    /// This should be called in viewDidAppear()
+    func setUpBottomButtonContainerView() {
+        NotificationCenter.when(NSNotification.Name.UIKeyboardWillShow) {
```
jstxx: Shouldn't need the extraneous enum prefix here

Blockchain/KYC/Views/BottomButtonContainerView.swift
```diff
@@ -0,0 +1,55 @@
...
+            self.keyboardWillShow(with: KeyboardPayload(notification: $0))
+        }
+        NotificationCenter.when(NSNotification.Name.UIKeyboardWillHide) {
```
jstxx: same thing here

Blockchain/KYC/Mobile Number/KYCEnterPhoneNumberController.swift
bchrisarriola: Actually `KYCConfirmPhoneNumberController` has `userId` as optional. Will remove the `??`

Blockchain/KYC/Mobile Number/KYCConfirmPhoneNumberController.swift
```diff
@@ -24,34 +21,39 @@ class KYCConfirmPhoneNumberController: UIViewController {
...
+        cleanUp()
     }
 
     // MARK: View Controller Lifecycle
 
     override func viewDidLoad() {
         super.viewDidLoad()
         // TICKET: IOS-1141 display correct % in the progress view
+        validationTextFieldConfirmationCode.keyboardType = .namePhonePad
+        validationTextFieldConfirmationCode.autocapitalizationType = .allCharacters
```
woudini: After changing `supportsAutoCorrect` and `autocapitalizationType` in `ValidationTextField` I can't get autocapitalization to work. Might be better to use `shouldChangeCharactersIn` and return the string.`uppercased()`

bchrisarriola: Let me try to get that fixed—better for the keyboard to be shown as capitalized chars

bchrisarriola: Looks like the `keyboardType` was interfering with characters being autocapitalized. Removing the line above fixes that (I think having the default keyboard type is fine). Fixed in 4a74f760

Blockchain/KYC/Views/BottomButtonContainerView.swift
```diff
@@ -0,0 +1,55 @@
...
+/// Protocol definition for a view that contains a CTA button at the bottom of it's view.
+/// Conforming to this protocol will auto-adjust the CTA button whenever the keyboard
+/// is presented/hidden.
+protocol BottomButtonContainerView {
+    var originalBottomButtonConstraint: CGFloat! { get set }
+
+    var layoutConstraintBottomButton: NSLayoutConstraint! { get }
+}
+
+extension BottomButtonContainerView where Self: UIViewController {
```
thisisalexmcgregor: 💯 

I need to figure out how to apply this to the `Personal Details` and `Address` screens. More so the `Personal Details` screen since for most screen sizes there's a good amount of space between the `primaryButton` and the input fields. They're both wrapped in a `UIScrollView` which complicates this a bit.

bchrisarriola: Ah, yeah. We might need another inheriting protocol that handles the case where the containing view has a `UIScrollView`

bchrisarriola:
> Will the PIN always be digits? If so should the keyboard type be changed?
If it's anything like 2FA for the wallet, the confirmation code will be alphanumeric.

PR #387, opened 2018-08-09
fix(PIN entry): Fix pin entry screen

jstxx:
## Objective

The PIN entry screen was being called each time the app would resign being active, which meant that if control center or such was launched the user would have to re-enter their PIN.  This changes the lifecycle of the privacy screen to only be called when the app is moved to the background.  When the app is inactive, we apply a blur view and remove it when the app state becomes active again.

## Description

Updates lifecycle methods in the app delegate.

## How to Test
Video is attached.  Run the app, try launching control center and/or switching between apps and verify that the screen is blurred and unblurred when moving back to the app. Also adds haptic feedback during PIN entry.

## Screenshot
ezgif-4-914b717d15
## Related Information

* Ticket Number:
iOS-554
## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/AppDelegate.swift
+++ b/Blockchain/AppDelegate.swift
@@ -102,7 +102,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func applicationWillResignActive(_ application: UIApplication) {
         Logger.shared.debug("applicationWillResignActive")
-        DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
             if application.applicationState != .active {
                 self.showBlurCurtain()
             }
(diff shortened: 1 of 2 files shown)


Blockchain/AppDelegate.swift
```diff
@@ -88,9 +102,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func applicationWillResignActive(_ application: UIApplication) {
         Logger.shared.debug("applicationWillResignActive")
-        if !AuthenticationCoordinator.shared.isPromptingForBiometricAuthentication {
-            showPrivacyScreen()
+        DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
```
woudini: I feel like the 2 seconds is a little long - if this is changed to 1 second do you foresee any issues?

woudini: Let's change to 1 if you're cool with that

PR #388, opened 2018-08-09
feat(KYC): Adding in network events for Personal Details and Address screens

thisisalexmcgregor:
## Objective

This adds in the networking related events that occur when you tap the `primaryButton` or the `Next` button in the `Personal Details` screen or the `Address` screen. 

## Description

* `Personal Details` - This didn't have a coordinator to handle the back and forth between API events and the view's interface. This PR adds those classes. 
* `KYCNetworkRequest` - I noticed that `application/x-www-form-urlencoded` was added as the `Content-Type` but, I didn't see anywhere in the docs that said we needed that and not `application/json`. This PR fixes that. Once I made that change the errors I received were no longer `500`. They now return a `400` which I think is expected at this point in development.
* The `PUT` request now takes a generic type as the `parameters`. So long as the payload is `Encodable` it should work. I plan on applying this to the other network events. 
* Design showed a `Next` button in the top left of the navigation bar. This PR adds that. When a network request is in flight we swap this out with a `UIActivityIndicatorView` and disable the `primaryButton`. 
* `UserAddress` - Not ideal as I'd rather use `PostalAddress` but, all those properties on that are optional given that they are populated by `MapKit`. But, it'd require some slight refactoring to use `PostalAddress` in both the `updateAddress` event and in the autocompletion logic.  

## How to Test

*Note:* For the `Address` screen or the `Personal Details` screen, go to `KYCWelcome.storyboard` and change the destination storyboard to the `KYCAddress.storyboard` or the `KYCPersonalDetails.storyboard`. 

1. Build and run
2. Tap Debug
3. Tap the continue button
4. View the respective screen.
5. Enter your data and submit
6. For the `Personal Details` screen, the network request will not fire as you *cannot* submit without an email. The email of the user is pulled from the wallet and the wallet is not initialized. 
7. For the `Address` screen, the network request will fire but will fail with a `400`. 

## Screenshot

N/A

## Related Information

* Ticket Number: IOS-1144

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Extensions/DateFormatter+Conveniences.swift
```diff
@@ -15,4 +15,10 @@ extension DateFormatter {
         formatter.timeStyle = .none
         return formatter
     }()
+
+    static let kycFormatter: DateFormatter = {
```
bchrisarriola: Instead of attaching a feature-specific name to this formatter, maybe name it to `birthdayDigitsFormatter`?

Blockchain/KYC/Personal/KYCPersonalDetailsController.swift
```diff
@@ -39,16 +39,39 @@ final class KYCPersonalDetailsController: UIViewController, ValidationFormView,
...
     var keyboard: KeyboardPayload? = nil
 
+    // MARK: Public Properties
+
+    weak var delegate: PersonalDetailsDelegate?
+
+    // MARK: Private Properties
+
+    fileprivate var coordinator: PersonalDetailsCoordinator!
+    fileprivate var nextBarButton: UIBarButtonItem!
```
bchrisarriola: Not needed for this screen

Blockchain/KYC/Address/KYCAddressController.swift
```diff
@@ -72,6 +76,18 @@ class KYCAddressController: UIViewController, ValidationFormView {
         validationFieldsSetup()
         setupNotifications()
 
+        nextBarButton = UIBarButtonItem(
```
bchrisarriola: Spoke with @thianh about this and the reason that's there is because the CTA button is hidden because the UIButton in the design is a subview of the `UIScrollView`. Instead, we shouldn't obscure the CTA when the keyboard comes up (i.e. the CTA button and the `UIScrollView` should have the same parent view). This kind of makes implementation easier, too.

Blockchain/KYC/Personal/API/PersonalDetailsInterface.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Alex McGregor on 8/9/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+protocol PersonalDetailsInterface: class {
+    func nextPage()
+    func rightBarButton(_ visibility: Visibility)
+    func primaryButtonEnabled(_ enabled: Bool)
```
bchrisarriola: var instead?

Blockchain/KYC/Personal/PersonalDetailsCoordinator.swift
```diff
@@ -0,0 +1,50 @@
...
+
+        if let controller = interface as? KYCPersonalDetailsController {
+            controller.delegate = self
+        }
+    }
+}
+
+extension PersonalDetailsCoordinator: PersonalDetailsDelegate {
+    func onStart() {
+        // TODO: Pre-populate fields from already fetched data.
```
bchrisarriola: 💯

Blockchain/KYC/Search/LocationSuggestionCoordinator.swift
```diff
@@ -80,6 +82,26 @@ extension LocationSuggestionCoordinator: SearchControllerDelegate {
         }
     }
 
+    func onSubmission(_ address: UserAddress) {
+        interface?.rightBarButton(.hidden)
+        interface?.updateBarButtonActivityIndicator(.visible)
+        interface?.primaryButtonEnabled(false)
+        // TODO: Pass in correct userID
+        api.updateAddress(address: address, for: "userID") { [weak self] (error) in
```
bchrisarriola: Take `userId` as part of this method arguments?

Blockchain/KYC/Address/KYCAddressController.swift
thisisalexmcgregor: 🤔 
So do we want to remove the Next button now?

The Address screen is a bit tricky there. If you're on an iPhone 5S and you're mid-way through the Address form entry, moving the `primaryButton` up would cover the text fields, even if I move the in focus field up a bit. Since the form is in a scrollView if you scroll the fields at all, we dismiss the keyboard and the `primaryButton` is visible at the bottom. 

Maybe I'm confused though. I haven't had a lot of ☕️ yet...

Blockchain/KYC/Personal/PersonalDetailsCoordinator.swift
thisisalexmcgregor: Not looking forward to that `// TODO` lol.

Blockchain/KYC/Personal/API/PersonalDetailsInterface.swift
thisisalexmcgregor: I can, only kept it that way so it was consistent with the rest of the protocol 🤷‍♂️

Blockchain/KYC/Search/LocationSuggestionCoordinator.swift
thisisalexmcgregor: Yep, agreed. Though where exactly do we get this elusive `userID` from? lol.

Blockchain/Extensions/DateFormatter+Conveniences.swift
thisisalexmcgregor: Yep, I can do. 👍

thisisalexmcgregor:
Marked DNM until I implement some feedback and changes per design discussion.

PR #389, opened 2018-08-09
feat(KYC): IOS-1049 creating account status view.

bchrisarriola:
## Objective

Implementing the KYC account status view

## Description

Implemented different states for the KYC account status view. There is a function in KYCCoordinator that will handle the presentation of this view. It's currently not called anywhere but should be glued in the flow eventually (should be displayed when tapping on the item in the settings view).

## How to Test

Invoked `KYCCoordinator#presentAccountStatusView()` and test different account statuses.

## Screenshot

simulator screen shot - iphone 5s - 2018-08-09 at 15 31 34
simulator screen shot - iphone 5s - 2018-08-09 at 15 31 15
simulator screen shot - iphone 5s - 2018-08-09 at 15 31 08
simulator screen shot - iphone 5s - 2018-08-09 at 15 30 59

## Related Information

* Ticket Number: IOS-1049

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/KYCAccountStatusController.swift
+++ b/Blockchain/KYC/KYCAccountStatusController.swift
@@ -14,12 +14,10 @@ struct KYCAccountStatusViewConfig {
 }
 
 extension KYCAccountStatusViewConfig {
-    static func defaultConfig() -> KYCAccountStatusViewConfig {
-        return KYCAccountStatusViewConfig(
-            titleColor: UIColor.gray5,
-            isPrimaryButtonEnabled: false
-        )
-    }
+    static let defaultConfig: KYCAccountStatusViewConfig = KYCAccountStatusViewConfig(
+        titleColor: UIColor.gray5,
+        isPrimaryButtonEnabled: false
+    )
 
     static func create(for accountStatus: KYCAccountStatus) -> KYCAccountStatusViewConfig {
         let titleColor: UIColor
@@ -69,7 +67,7 @@ final class KYCAccountStatusController: UIViewController {
     }
 
     /// The view configuration for this view
-    var viewConfig: KYCAccountStatusViewConfig = KYCAccountStatusViewConfig.defaultConfig()
+    var viewConfig: KYCAccountStatusViewConfig = KYCAccountStatusViewConfig.defaultConfig
 
     // MARK: - Lifecycle Methods
 
--- a/Blockchain/KYC/KYCAccountStatusController.swift
+++ b/Blockchain/KYC/KYCAccountStatusController.swift
@@ -8,72 +8,90 @@
 
 import UIKit
 
+struct KYCAccountStatusViewConfig {
+    let titleColor: UIColor
+    let isPrimaryButtonEnabled: Bool
+}
+
+extension KYCAccountStatusViewConfig {
+    static func defaultConfig() -> KYCAccountStatusViewConfig {
+        return KYCAccountStatusViewConfig(
+            titleColor: UIColor.gray5,
+            isPrimaryButtonEnabled: false
+        )
+    }
+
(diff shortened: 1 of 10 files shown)


Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -19,15 +22,56 @@ import Foundation
...
+            switch viewController.accountStatus {
+            case .approved:
+                viewController.dismiss(animated: true) {
+                    ExchangeCoordinator.shared.start()
+                }
+            case .inProgress:
+                PushNotificationManager.shared.requestAuthorization()
+            case .failed:
+                // Confirm with design that this is how we should handle this
+                guard let url = URL(string: Constants.Url.blockchainSupport) else { return }
```
jstxx: How about an extension, I've seen this kind of code written a lot. 

```
typealias 🚀 = Constants.Url.blockchainSupport
extension Constants {
    func launch() {
        guard let url = URL(string: self) else { return }
        UIApplication.shared.open(url)
    }
}
```
`🚀.launch() `

Blockchain/KYC/KYCAccountStatusController.swift
```diff
@@ -8,72 +8,90 @@
...
 
-final class KYCAccountStatusController: UIViewController {
+struct KYCAccountStatusViewConfig {
+    let titleColor: UIColor
+    let isPrimaryButtonEnabled: Bool
+}
 
-    // MARK: - Properties
+extension KYCAccountStatusViewConfig {
+    static func defaultConfig() -> KYCAccountStatusViewConfig {
```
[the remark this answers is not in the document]
thisisalexmcgregor: Yeah that or:

`static let default: KYCAccountStatusViewConfig = KYCAccountStatusViewConfig(blah, blah)`

So call site would look like `KYCAccountStatusViewConfig.default`, semantics 🤷‍♂️

thisisalexmcgregor:
Is there anything in the KYC endpoints that indicates account status yet?

thisisalexmcgregor:
Derp, never mind, found `"kycState": "NONE"` on the user call.

Blockchain/KYC/KYCAccountStatusController.swift
```diff
@@ -8,72 +8,90 @@
...
+            isPrimaryButtonEnabled: false
+        )
+    }
 
-    enum AccountStatus {
-        case approved, failed, inReview
-        /// Graphic which visually represents the account status
-        var image: UIImage {
-            switch self {
-            case .approved: return UIImage(named: "AccountApproved")!
```
thisisalexmcgregor: TIL if you use an image literal it's not optional.

Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -19,15 +22,56 @@ import Foundation
...
+                UIApplication.shared.open(url)
+            default:
```
thisisalexmcgregor: I would put in _all_ the enum options. Reason being, if another status is added other than what we currently have, this wont compile and it'll force us to account for it wherever it's used.

Blockchain/KYC/KYCCoordinator.swift
bchrisarriola: I think an extension on `URL` makes sense for a function called `launch()`. Will add that 👍

Blockchain/KYC/KYCCoordinator.swift
bchrisarriola: Good point—will update 👍

Blockchain/KYC/KYCAccountStatusController.swift
bchrisarriola: Nice! Will udpate

PR #393, opened 2018-08-13
feat(KYC): User Drop Off for KYC

thisisalexmcgregor:
## Objective

This is for handling when a user exits and resumes the KYC flow. 

*Note: This PR is marked DNM as it's not complete. I would like to get 👀 on it prior to completion.*

## Description

`KYCPageType` - My thinking is that each screen should be associated with a `KYCPageType`. Some types have a model (optional). Should it have said model, the screen will be pre-populated with that data. 
`KYCUser` - This is the user model. It's from the user model that we determine what `KYCPageType` the user is at. 

## How to Test

N/A

## Screenshot

N/A

## Related Information

* Ticket Number: IOS-1142

## Merge Checklist

- [x] All unit tests pass.


Blockchain/KYC/Models/KYCPage.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+enum KYCPageType {
+    typealias PhoneNumber = String
+
+    case welcome
+    case profile
```
bchrisarriola: Add `countrySelection` here as well?

Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -62,15 +131,53 @@ import Foundation
...
-            Logger.shared.warning("Could not instantiated KYCOnboardingNavigationController")
-            return
+        ).instantiateInitialViewController() as? KYCOnboardingNavigationController
+
+        navController.pushViewController(viewController, animated: false)
+        navController.modalTransitionStyle = .coverVertical
+        presentingViewController.present(navController, animated: true)
+    }
+
+    private func screenFor(pageType: KYCPageType) -> KYCBaseViewController {
```
bchrisarriola: Thoughts on adding a computed property in `KYCPageType`? Something like:

```
enum KYCPageType {
    var viewController: KYCBaseViewController.Type {
        switch self {
              case .welcome: return KYCWelcomeController.self
              // etc.
        }
    }
} 
```

...and then this function can just be:

```
pageType.viewController.make(with: self)
```

thisisalexmcgregor: Thought about it and even typed it out earlier lol. I opted not to as I didn't want a `KYCBaseViewController` to have the ability to do:
`pageType.viewController` without going through the coordinator. 

🤷‍♂️

bchrisarriola: Ah, makes sense!

Blockchain/KYC/KYCBaseViewController.swift
```diff
@@ -0,0 +1,26 @@
...
+//  KYCBaseViewController.swift
+//  Blockchain
+//
+//  Created by Alex McGregor on 8/13/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+class KYCBaseViewController: UIViewController {
```
jstxx: Any particular reason you're subclassing this as opposed to conforming to a `KYCBaseView` protocol and then overriding later?

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -19,13 +19,30 @@ final class KYCNetworkRequest {
...
+            case credentialsForProvider
+            case healthCheck
+            case listOfCountries
+            case nextKYCMethod
+            case users(userID: String)
+
+            var path: String {
+                switch self {
+                case .credentials:
+                    return "/kyc/credentials"
```
jstxx: Ideally this should be constructed as [URLComponents](https://developer.apple.com/documentation/foundation/urlcomponents)

Blockchain/KYC/Personal/KYCPersonalDetailsController.swift
```diff
@@ -122,14 +130,18 @@ final class KYCPersonalDetailsController: UIViewController, ValidationFormView,
         guard let email = WalletManager.shared.wallet.getEmail() else { return }
         validationFields.forEach({$0.resignFocus()})
 
-        let details = PersonalDetails(
-            identifier: "",
-            firstName: firstNameField.text ?? "",
-            lastName: lastNameField.text ?? "",
+        guard let details = PersonalDetails(
+            id: "",
```
jstxx: 🤔 Will we be generating a NONCE for this?

Blockchain/KYC/Address/KYCAddressController.swift
```diff
@@ -30,6 +30,16 @@ class KYCAddressController: UIViewController, ValidationFormView {
 
     @IBOutlet var scrollView: UIScrollView!
 
+    // MARK: Factory
+
+    override class func make(with coordinator: KYCCoordinator) -> KYCAddressController {
```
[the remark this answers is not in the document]
thisisalexmcgregor: Sounds good, looking into this now.

Blockchain/KYC/KYCNetworkRequest.swift
thisisalexmcgregor: Agreed. There is this: 
`func endpoint(_ baseURL: URL, pathComponents: [String]?, queryParameters: [String : String]?) -> URL?`
I can clean up `KYCNetworkRequest` to use that.

jstxx: I'm totally fine with that provided we also change the function signature on `endpoint` so that it takes a baseURL and URLComponents.

Blockchain/KYC/Personal/KYCPersonalDetailsController.swift
bchrisarriola: I think this is just temporary, this view controller should be injected with a valid userID (maybe mark as a TODO?).

thisisalexmcgregor: ☝️ Yep, `userID` is up in the air and likely will be injected into the view controller.

Blockchain/KYC/Address/KYCAddressController.swift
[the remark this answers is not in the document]
bchrisarriola: That's a nice helper 👍 

I don't think the storyboard names are the same as the class names though (i.e. some leave out the `*Controller` suffix) so those will need to be updated as well.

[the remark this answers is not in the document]
thisisalexmcgregor: Yep, addeed [here.](https://github.com/blockchain/My-Wallet-V3-iOS/pull/393/commits/5ecd9ab069521fcdaf5c57efd733d338b3558a8f)

Blockchain/KYC/KYCBaseViewController.swift
thisisalexmcgregor: When the user goes to a new screen, in `viewWillAppear` the `KYCBaseViewController` calls `coordinator.handle(event: .pageWillAppear(pageType))`. If the coordinator has a `KYCUser` and that user has all the necessary data for that particular screen, the fields will be pre-populated with data the user already entered. Alternatively they'd be pushed to the next screen in the flow (sans animation).

Blockchain/KYC/Address/KYCAddressController.swift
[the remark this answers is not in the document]
thisisalexmcgregor: Good call on this btw. I like it.

PR #394, opened 2018-08-13
feat(KYC): IOS-1045 Updating country selection screen

bchrisarriola:
## Objective

Polishing country selection screen.

## Description

Summary:
* Polishing country selection screen by implementing the search field, adding right hand side jump to section view
* Handling logic for what should happen when the user taps on a country and the country either is: (1) supported by our native KYC, (2) supported by Shapeshift or (3) is not supported at all.
* Added unit tests for logic mentioned above and fixed unit tests (had to add Swift files to BlockchainTests)

Note: some of this will change (i.e. `KYCCountrySelectionController` needs to be updated to integrate w/ `KYCCoordinator`) once #393 gets merged

## How to Test

Pull this branch and "Launch KYC" from the debug view

## Screenshot

simulator screen shot - iphone 5s - 2018-08-13 at 16 56 25
simulator screen shot - iphone 5s - 2018-08-13 at 16 48 05

## Related Information

* Ticket Number: IOS-1045

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/CountrySelector/KYCCountrySelectionController.swift
+++ b/Blockchain/KYC/CountrySelector/KYCCountrySelectionController.swift
@@ -34,8 +34,8 @@ final class KYCCountrySelectionController: KYCBaseViewController, ProgressableVi
 
     private var selectedCountry: KYCCountry?
 
-    private lazy var presenter: KYCCountrySelectorPresenter = {
-        return KYCCountrySelectorPresenter(view: self)
+    private lazy var presenter: KYCCountrySelectionPresenter = {
+        return KYCCountrySelectionPresenter(view: self)
     }()
 
     // MARK: Factory
@@ -109,7 +109,7 @@ extension KYCCountrySelectionController: UITableViewDataSource, UITableViewDeleg
     }
 }
 
-extension KYCCountrySelectionController: KYCCountrySelectorView {
+extension KYCCountrySelectionController: KYCCountrySelectionView {
     func continueKycFlow(country: KYCCountry) {
         // TICKET: IOS-1142 - move to coordinator
         performSegue(withIdentifier: "promptForPersonalDetails", sender: self)
--- a/Blockchain/KYC/CountrySelector/KYCCountrySelectionController.swift
+++ b/Blockchain/KYC/CountrySelector/KYCCountrySelectionController.swift
@@ -6,16 +6,37 @@
 //  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
 //
 
+import RxSwift
 import UIKit
 
 /// Country selection screen in KYC flow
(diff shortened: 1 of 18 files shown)


thisisalexmcgregor:
Chris flexin' on us with his tests 🥇 
Nice!

Blockchain/Network/Models/WalletOptions.swift
```diff
@@ -60,10 +69,23 @@ extension WalletOptions.MobileInfo {
...
 }
 
+extension WalletOptions.Shapeshift {
+    init(json: JSON) {
+        guard let shapeshiftJson = json[Keys.shapeshift] as? JSON else {
+            self.countriesBlacklist = []
+            self.statesWhitelist = []
+            return
+        }
+        self.countriesBlacklist = shapeshiftJson["countriesBlacklist"] as? [String] ?? []
```
jstxx: Having this be optional seems to make more sense as opposed to an empty array

thisisalexmcgregor:
I think we should dismiss the keyboard if the user scrolls the tableview.

Blockchain/KYC/CountrySelector/KYCCountrySelectionController.swift
```diff
@@ -25,56 +108,107 @@ final class KYCCountrySelectionController: KYCBaseViewController, UITableViewDat
...
+            return
+        }
+        Logger.shared.info("User selected '\(selectedCountry.name)'")
+        presenter.selected(country: selectedCountry)
+    }
+}
+
+extension KYCCountrySelectionController: KYCCountrySelectionView {
+    func continueKycFlow(country: KYCCountry) {
+        // TICKET: IOS-1142 - move to coordinator
```
thisisalexmcgregor: The `KYCCoordinator` has a `nextPageFromPageType` event. That said, it doesn't do anything currently for forwarding this data to the next screen. Also, not all screens need the data from the prior screen. Thoughts on how we should handle this? I'm open to anything.

thisisalexmcgregor:
There's still a [UISearchDisplayCountroller](https://developer.apple.com/documentation/uikit/uisearchdisplaycontroller) in the storyboard. We should swap this out with a `UISearchBar`, but we can do it in a separate PR too.

bchrisarriola:
> There's still a UISearchDisplayCountroller in the storyboard. We should swap this out with a UISearchBar, but we can do it in a separate PR too.

Ah, thought I removed that. Will update.

Blockchain/KYC/CountrySelector/KYCCountrySelectionController.swift
bchrisarriola: I think the coordinator should probably store that state and in the `make` factory method, a view controller that requires state from a previous view controller (like how KYCAddressController might need the selected country from KYCCountrySelectionController) and retrieve it from the provided coordinator. Wdyt?

PR #395, opened 2018-08-14
fix(KYCUser): KYCUser data mapping

thisisalexmcgregor:
## Objective

This updates `KYCUser` to reflect what is documented in the API. 

**Note: It's possible the userID will be removed entirely per internal discussion.**

## Description

* `KYCUser` now has an `id`, `state`, and `status`.
* `KYCAccountStatus` has now been updated to reflect what is documented in the API

## How to Test

N/A - `users` endpoint unavailable. 

## Screenshot

N/A

## Related Information

* Ticket Number: N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/Models/KYCAccountStatus.swift
+++ b/Blockchain/KYC/Models/KYCAccountStatus.swift
@@ -8,16 +8,19 @@
 
 import Foundation
 
-enum KYCAccountStatus: Int {
-    case approved, failed, underReview, inProgress
+enum KYCAccountStatus: String {
+    case none = "NONE"
+    case expired = "EXPIRED"
+    case approved = "VERIFIED"
+    case failed = "REJECTED"
+    case pending = "PENDING"
 
     /// Graphic which visually represents the account status
     var image: UIImage {
         switch self {
         case .approved: return #imageLiteral(resourceName: "AccountApproved")
-        case .failed:   return #imageLiteral(resourceName: "AccountFailed")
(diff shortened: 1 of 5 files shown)


Blockchain/KYC/Models/KYCAccountStatus.swift
```diff
@@ -8,33 +8,38 @@
 
 import Foundation
 
-enum KYCAccountStatus: Int {
-    case approved, failed, underReview, inProgress
+enum KYCAccountStatus: String {
+    case none = "NONE"
```
jstxx: In my PR I have a struct that addresses this https://github.com/blockchain/My-Wallet-V3-iOS/pull/396/files.  It would be nice to combine the two so we can have the enum `KYCAccountStatus` conform to Codable and then instead of the string in `KYCUserResponse` we could reference the enum.  Thoughts?

bchrisarriola: ++ to @jstxx's idea

bchrisarriola: Can we consolidate `KYCUser` and `KYCUserResponse`?

thisisalexmcgregor: @jstxx with `KYCUser` already in `dev` do you want to pull in those changes and modify as you need it?

PR #396, opened 2018-08-14
feat(KYC | Update Settings Screen)

jstxx:
## Objective

Updates settings with a new visual spec and KYC status cell indicator.

## Description

Added static table cell to Settings.storyboard and added various methods to check the current verification status of the user using `KYCNetworkRequest`. 

## How to Test

Launch the app, go to settings and verify the new visual design matches the spec.  The KYC flow cell does not currently launch from the cell.

## Screenshot
simulator screen shot - iphone 8 plus - 2018-08-14 at 14 44 57

*If the change is UI related, paste a screenshot here.*

## Related Information

* Ticket Number:

## Merge Checklist

- [ ] All unit tests pass.


Blockchain/KYC/Models/KYCAccountStatus.swift
```diff
@@ -8,6 +8,25 @@
...
+
+struct KYCUserResponse: Codable {
+    let firstName: String?
+    let lastName: String?
+    let email: String?
+    let mobile: String?
+    let mobileVerified: Bool
+    let address: KYCUserAddress?
+    let kycState: String
+}
```
thisisalexmcgregor: [KYCUser](https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/KYC/Models/KYCUser.swift) is already available. [This PR](https://github.com/blockchain/My-Wallet-V3-iOS/pull/395) adds a few more things to it and fixes some mapping issues.

Blockchain/Coordinators/AppCoordinator.swift
```diff
@@ -110,8 +110,8 @@ import Foundation
     }
     
     @objc func showKYCView() {
-        let storyboard = UIStoryboard(name: "KYCVerifyIdentity", bundle: nil)
-       let upgradeViewController = storyboard.instantiateViewController(withIdentifier: "KYCVerifyIdentity")
+        let storyboard = UIStoryboard(name: "KYCVerifyIdentityController", bundle: nil)
+        let upgradeViewController = storyboard.instantiateViewController(withIdentifier: "KYCVerifyIdentity")
```
thisisalexmcgregor: Not sure if this is up to date with `dev` but, that `Storyboardable` protocol is available to you 🎉

Blockchain/KYC/Models/KYCAccountStatus.swift
```diff
@@ -8,6 +8,25 @@
 
 import Foundation
 
+struct KYCUserAddress: Codable {
+    let city: String?
+    let line1: String?
+    let line2: String?
+    let state: String?
+    let country: String?
+    let postCode: String?
```
thisisalexmcgregor: When we pull down the user's address it comes down as a `UserAddress`. Arguably it should be renamed to `KYCUserAddress`  but aside from that we have it [here](https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/KYC/Search/Models/PostalAddress.swift).

Blockchain/Settings/Settings+Table.swift
```diff
@@ -11,7 +11,7 @@ import Foundation
 extension SettingsTableViewController {
 
     enum SettingsCell {
-        case base, wallet, email, phoneNumber, currency, recovery, emailNotifications, twoFA, biometry, swipeReceive
+        case base, identity, wallet, email, phoneNumber, currency, recovery, emailNotifications, twoFA, biometry, swipeReceive
```
thisisalexmcgregor: _woah._ That's a lot. 

Dunno what everyone's thoughts on this is but maybe a `// TICKET` to track this as tech debt?

Blockchain/Settings/Settings+Table.swift
```diff
@@ -142,8 +147,40 @@ extension SettingsTableViewController {
...
+            }
+        }, taskFailure: { error in
+            // TODO: handle error
+            handler(nil, false)
+            Logger.shared.error(error.debugDescription)
+        })
+    }
+    
+    func prepareIdentityCell(_ cell: UITableViewCell) {
+        DispatchQueue.main.async {
```
thisisalexmcgregor: I think this should be moved to the completion handler so the creation of the cell is on the main thread.

Blockchain/Settings/Settings+Table.swift
```diff
@@ -142,8 +147,40 @@ extension SettingsTableViewController {
         cell.accessoryView = switchForEmailNotifications
     }
 
+    func getUserVerificationStatus(handler: @escaping (KYCUserResponse?, Bool) -> Void) {
+        KYCNetworkRequest(get: .users(userID: "userID"), taskSuccess: { responseData in
```
thisisalexmcgregor: Not a big deal since `KYCCoordinator` is a bit in flux, but we have this fetch in two places. Maybe it should expose a property with the user's status? Not sure tbh.

Blockchain/Settings/Settings+Table.swift
```diff
@@ -195,23 +234,35 @@ extension SettingsTableViewController {
...
+
     override func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
         let headerView = UIView()
         headerView.backgroundColor = .clear
-
-        let headerLabel = UILabel(frame: CGRect(x: 18, y: 12, width:
-            tableView.bounds.size.width, height: 45))
-        headerLabel.font = UIFont(name: Constants.FontNames.montserratRegular, size: Constants.FontSizes.SmallMedium)
-        headerLabel.textColor = .brandPrimary
+        let headerLabel = UILabel(frame: CGRect(x: 18, y: 3, width:
```
thisisalexmcgregor: Lot of magic numbers here. Does this translate okay to various device sizes?

Blockchain/Settings/Settings+Table.swift
```diff
@@ -195,23 +234,35 @@ extension SettingsTableViewController {
...
+        headerView.layoutIfNeeded()
         return headerView
     }
+
     override func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
-        return 45
+         if section == 0 {
+                return 50
+        }
+        return 32
```
thisisalexmcgregor: Same thing here.

Blockchain/View Controllers/Settings/SettingsViewController.swift
```diff
@@ -48,7 +53,8 @@ MobileNumberDelegate, WalletAccountInfoDelegate {
     weak var delegate: (UIViewController & EmailDelegate)!
     weak var numberDelegate: (UIViewController & MobileNumberDelegate)!
     let walletManager: WalletManager
-
+    var userIdentityStatus: KYCUserResponse?
```
thisisalexmcgregor: This brings me back to `KYCCoordinator`. Not sure if we want a user status on both the `SettingsViewController` and on the `KYCCoordinator`. I definitely think settings needs a way to access this meta data but, not sure if it should own that ability to fetch it.

Blockchain/Settings/Settings+Helpers.swift
```diff
@@ -129,14 +129,26 @@ extension AppSettingsController {
...
+            cell.detailTextLabel?.textColor = .white
         }
     }
     
+    func createBadge(_ cell: UITableViewCell,  color: UIColor? = nil, _ using: KYCUserResponse? = nil) {
+        cell.detailTextLabel?.layer.cornerRadius = 4
+        cell.detailTextLabel?.layer.masksToBounds = true
+        cell.detailTextLabel?.backgroundColor = .unverified
+        cell.detailTextLabel?.textColor = .white
+        cell.detailTextLabel?.frame.origin.x += 25
```
thisisalexmcgregor: Looks like Settings has lots of different types of cell types so, I can understand why we have this. 

Maybe we should track it in a ticket and make standard `ViewModels` for all these different types?

Blockchain/Settings/Settings+Table.swift
bchrisarriola: If the `KYCUser` is available in `KYCCoordinator`, then we'll be able to get the user's status from there. We would need to have a "if not nil" check though to make sure the `KYCUser` object is available though.

If we're open to using Rx here, we can do something similar to how getting `WalletOptions` is implemented. So that the first time we try to get the `KYCUser` object, it'll fetch it over the network, otherwise, every subsequent reference will use a cached value. From the caller's side, it doesn't need to know where that object is coming from, it'll just receive it.

https://github.com/blockchain/My-Wallet-V3-iOS/blob/f5d8a0a729ae2f5be4caf44de5f9ce499e503cc9/Blockchain/Network/API/WalletService.swift#L26...L56

thisisalexmcgregor: I'm down. Show me the `Rx` ways. ✨

Blockchain/Settings/Settings+Table.swift
jstxx: Have been testing on different sized devices.  So far okay.

Blockchain/KYC/Models/KYCAccountStatus.swift
jstxx: Yah I'll pull all that in, it wasn't merged when I was digging into this.

Blockchain/Coordinators/AppCoordinator.swift
jstxx: Nice, that will get implemented.

Blockchain/Settings/Settings+Table.swift
jstxx: There is bound to be a way to make this cleaner for sure.

Blockchain/Settings/Settings+Helpers.swift
jstxx: Maaaaybe, I like the idea of having IBDesignables and then mocking them in interface builder, if it didn't crash XCode so frequently.  Will think on it.

Blockchain/Settings/Settings+Table.swift
woudini: Would organizing by section work better? `ProfileCell`, `PreferencesCell`...

Blockchain/Settings/Settings+Helpers.swift
thisisalexmcgregor: `ValidationTextField` uses that. Works pretty well once its setup, but yeah IB can be finicky initially.

Blockchain/Settings/Settings+Table.swift
bchrisarriola: What that looks like in Rx! https://github.com/blockchain/My-Wallet-V3-iOS/pull/399/files#diff-d787741e6bfc1810819cc6c27e35ed8aR20

Blockchain/Settings/Settings+Table.swift
woudini: Would be ideal to use `BlockchainDataRepository` here too.

bchrisarriola: ++ to using `BlockchainDataRepository` to consolidate how we fetch the `KYCUser`.

Usage:

```
private var: Disposable?

deinit {
    disposable?.dispose()
}

func foo() {
    disposable = BlockchainDataRepository.shared.kycUser
                               .subscribeOn(MainScheduler.async) // network call will be performed off the main thread
                               .observeOn(MainScheduler.instance) // closures passed in subscribe will be on the main thread
                               .subscribe(onSuccess: { user in
                                   // do stuff
                               }, onError: {
                                   // handle error
                               })
}
```

Blockchain/KYC/Models/KYCAccountStatus.swift
```diff
@@ -65,4 +65,15 @@ enum KYCAccountStatus: String {
         case .none, .expired: return nil
         }
     }
+
+ var badge: String {
```
bchrisarriola: Spacing. Also looks like you need to merge/rebase your branch so you have the latest version of `KYCAccountStatus`

Blockchain/Settings/Settings+Helpers.swift
```diff
@@ -129,14 +129,26 @@ extension AppSettingsController {
...
     }
     
+    func createBadge(_ cell: UITableViewCell, color: UIColor? = nil, _ using: KYCUser? = nil) {
+        cell.detailTextLabel?.layer.cornerRadius = 4
+        cell.detailTextLabel?.layer.masksToBounds = true
+        cell.detailTextLabel?.backgroundColor = color
+        cell.detailTextLabel?.textColor = .white
+        cell.detailTextLabel?.font = UIFont(name: Constants.FontNames.montserratSemiBold, size: Constants.FontSizes.Tiny)
+        cell.detailTextLabel?.layoutIfNeeded()
+    //    cell.detailTextLabel?.sizeToFit()
```
bchrisarriola: Remove

Blockchain/Settings/Settings+Table.swift
bchrisarriola: To enable unit testing, view controllers shouldn't invoke network calls directly and instead should be the role of another component (like an interactor)

Blockchain/Settings/Settings+Table.swift
bchrisarriola: ++ to avoiding magic numbers especially in view controllers. Can this be moved to a .xib and have those numbers defined there?

Podfile
```diff
@@ -23,3 +23,9 @@ target 'Blockchain' do
   end
 
 end
+post_install do |installer|
+    installer.pods_project.build_configurations.each do |config|
+        config.build_settings.delete('CODE_SIGNING_ALLOWED')
+        config.build_settings.delete('CODE_SIGNING_REQUIRED')
```
bchrisarriola: Do we want to delete this for all build configurations? Curious why these lines were added (build optimization?)?

jstxx: It was an attempt to fix Interface Builder crashing when using IBInspectable

jstxx:
I'm okay with using a xib for a TableView header.  Working on making the `getUserVerificationStatus` use the `BlockchainDataRepository` now

Blockchain/Extensions/Strings.swift
```diff
@@ -9,7 +9,15 @@
 import Foundation
 
 extension String {
-
+    
+    func capitalizingFirstLetter() -> String {
```
jstxx: These aren't currently used but it's nice to have as a helper function imo.

Blockchain/Settings/Settings+Table.swift
```diff
@@ -142,8 +124,36 @@ extension SettingsTableViewController {
...
+            .subscribe(onSuccess: { user in
+                handler(user, true)
+            }, onError: {  error in
+                handler(nil, false)
+            })
+    }
+    
+    func prepareIdentityCell(_ cell: UITableViewCell) {
+        DispatchQueue.main.async {
+        self.getUserVerificationStatus { status, success in
```
thisisalexmcgregor: Does `getUserVerificationStatus` need to be forced onto the main thread or just updating the cell in the completion handler?

Blockchain/Settings/Settings+Table.swift
```diff
@@ -142,8 +124,36 @@ extension SettingsTableViewController {
         cell.accessoryView = switchForEmailNotifications
     }
 
+    func getUserVerificationStatus(handler: @escaping (KYCUser?, Bool) -> Void) {
+        disposable = BlockchainDataRepository.shared.kycUser
+            .subscribeOn(MainScheduler.asyncInstance) // network call will be performed off the main thread
```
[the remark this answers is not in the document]
jstxx: Ack I missed a subclass on the Wallet "Copy" label, thanks, fixed now.

Podfile
woudini: When exactly was it crashing? Settings is working fine for me without the added code above as of now.

woudini:
Wallet ID badge is not vertically centered and changing frame when scrolling:
id badge

woudini:
Still seeing https://github.com/blockchain/My-Wallet-V3-iOS/pull/396#issuecomment-413880640 (not sure exactly what https://github.com/blockchain/My-Wallet-V3-iOS/pull/396/commits/9067de6dd094218c97acfbe0f1d810cb82d6420b was intended to address)

Would also like clarification on https://github.com/blockchain/My-Wallet-V3-iOS/pull/396#discussion_r210922975, modifying things related to code signing doesn't seem like something that should be added.

Podfile
woudini: Discussed with @jstxx and this is just for debugging.

woudini:
Seeing this exception

    Failed to set (background) user defined inspected property on (Blockchain.EdgeInsetBadge): [<Blockchain.EdgeInsetBadge 0x7f875bd6b7f0> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key background.

Blockchain/Settings/Settings+Helpers.swift
```diff
@@ -129,14 +129,26 @@ extension AppSettingsController {
...
             cell.detailTextLabel?.textColor = .green
         } else {
+            createBadge(cell, color: .unverified)
             cell.detailTextLabel?.text = LocalizationConstants.unverified
-            cell.detailTextLabel?.textColor = .error
+            cell.detailTextLabel?.textColor = .white
         }
     }
     
+    func createBadge(_ cell: UITableViewCell, color: UIColor? = nil, _ using: KYCUser? = nil) {
```
woudini: Should these parameters be changed? - here a `KYCAccountStatus` is being passed in but it's not being used: https://github.com/blockchain/My-Wallet-V3-iOS/pull/396/files#diff-e638a3d37f8b87c67b0a015b111bece5R146

So for this `KYCInformationViewModel`, the badge will have a white background, hiding the text "Failed":
```
▿ KYCInformationViewModel
  - title : "Verification Failed"
  - subtitle : nil
  - description : "Unfortunately we had some trouble verifying your identity with the documents you\'ve supplied."
  - buttonTitle : nil
  ▿ badge : Optional<String>
    - some : "Failed"

PR #398, opened 2018-08-14
feat(KYC): IOS-1104 add user authentication API

woudini:
## Objective

To (loosely) organize the user authentication process for KYC.

## Description

WIP for client authentication - it does not include fully working requests but opening this now to keep later PRs small.

I'm not totally clear on what to pass into the requests yet which is why `KYCCoordinator`'s `authenticate()` is long and incomplete.

This PR also includes adding public key pinning and user agent to `KYCNetworkRequest`.

## How to Test
If you _really_ want to test...
Run `sh scripts/install-js.sh && sh scripts/build-js.sh`.
Add the following to `start()` in `KYCCoordinator`.

    authenticate()
    return

Add the following to `handleWebLogin()` in `AppCoordinator`.

    let kycCoordinator = KYCCoordinator()
    kycCoordinator.start()
    return

Tapping on "Log in to web wallet" in the side menu should show a `Logger` error
## Screenshot

N/A

## Related Information

* Ticket Number: 1104

## Merge Checklist

- [x] All unit tests pass.


Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -85,9 +89,17 @@ final class KYCNetworkRequest {
     // MARK: - Initialization
 
     private init(url: URL, httpMethod: String) {
+        super.init()
+
+        let sessionConfiguration = URLSessionConfiguration.default
```
woudini: This is going to end up being pretty wasteful - a new session for every request. Not sure where the best place to store this is

Blockchain/Models/Wallet.m
```diff
@@ -2589,6 +2589,33 @@ - (BOOL)isWithdrawalTransaction:(NSString *)txHash
     return NO;
 }
 
+# pragma mark - Retail Core
+
+- (void)updateKYCUserCredentials:(NSString *)userId lifetimeToken:(NSString *)lifetimeToken success:(void (^)(NSString *))success error: (void (^)(NSString *))error
+{
+    if ([self isInitialized]) {
+        self.context[@"objc_updateUserCredentials_success"] = success;
```
bchrisarriola: That was easy 👍

Blockchain/Network/HttpHeader.swift
```diff
@@ -0,0 +1,22 @@
...
+struct HttpHeaderField {
+    static let userAgent = "User-Agent"
+    static let accept = "Accept"
+    static let contentLength = "Content-Length"
+    static let contentType = "Content-Type"
+}
+
+struct HttpHeaderValue {
+    static let json = "application/json"
+    static let accept = "Accept"
```
bchrisarriola: Remove?

Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -168,4 +168,53 @@ protocol KYCCoordinatorDelegate: class {
 
         return .address
     }
+
+    func authenticate() {
```
bchrisarriola: Just to clarify I understand what this method is doing. So when we are starting KYC and we need to obtain an authentication token from the server we are:
1. creating/fetching the KYC user ID
    1a. if we are creating a new KYC user, we also update the wallet metadata with the KYC user ID
2. given the KYC user ID, we request for a new session token

I definitely see the value in using Rx here to coordinate all these async calls and make it more understandable (we can get to that after)

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -177,3 +189,11 @@ final class KYCNetworkRequest {
         task.resume()
     }
 }
+
+extension KYCNetworkRequest: URLSessionDelegate {
+    func urlSession(_ session: URLSession,
+                    didReceive challenge: URLAuthenticationChallenge,
+                    completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void) {
+        CertificatePinner.shared.didReceive(challenge, completion: completionHandler)
```
bchrisarriola: I'm guessing this we wouldn't need to explicitly handle this if we had used `NetworkManager` directly?

woudini: Correct, however I did realize just now that the `NetworkManager`'s `URLSession` is public. Thoughts on using it directly in `KYCNetworkRequest`? We're going to have to refactor this anyway, so it will be less code to manage overall.

bchrisarriola: Yeah, might as well update KYCNetworkRequest to use `NetworkManager.shared.session` 👍

Blockchain/KYC/KYCCoordinator.swift
woudini: Yes, although we aren't reloading metadata to fetch the KYC user ID, just checking if it's there.

bchrisarriola:
Going to work off of this and integrate this in `BlockchainDataRepository`

woudini:
Feature implemented by https://github.com/blockchain/My-Wallet-V3-iOS/pull/405

PR #399, opened 2018-08-14
feat(KYC): Creating BlockchainDataRepository.

bchrisarriola:
## Objective

Creating KYCDataRepository. A component that handles fetching KYC-related models (user, countries, etc.)

## Description

`KYCDataRepository` is a component wherein layers can retrieve KYC-related data, such as the current `KYCUser`, while being agnostic as to where data is being fetched from. The 1st time a property in `KYCDataRepository` is accessed, it will be retrieved over the network followed by caching in memory. Callers are all provided a `Single<KYCData: Decodable>` type regardless if data is coming from memory or over the network.

Open to feedback if you guys think something like this is a good idea for other components to use (e.g. we can update `KYCCoordinator` to use this as well as the settings view for accessing a `KYCUser`)

## How to Test

Pull in these changes and launch the KYC feature. At the moment, only `KYCCountrySelectionController` is using this component. To verify that it's accessing cached data, enter/exit a couple times from the `KYCWelcomeController` to `KYCCountrySelectionController`.

## Screenshot

N/A

## Related Information

* Ticket Number: N/A

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/KYCDataRepository.swift
+++ b/Blockchain/KYC/KYCDataRepository.swift
@@ -1,0 +1,75 @@
+//
+//  KYCDataRepository.swift
+//  Blockchain
+//
+//  Created by Chris Arriola on 8/14/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import RxSwift
+
+/// Repository for KYC related data. Accessing properties in this repository
+/// will be fetched from the cache (if available), otherwise, data will be fetched over
+/// the network and subsequently cached for faster access.
+@objc class KYCDataRepository: NSObject {
+
+    static let shared = KYCDataRepository()
+
+    // MARK: - Public Properties
+
+    var user: Single<KYCUser> {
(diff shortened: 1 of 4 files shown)


thisisalexmcgregor:
This is really cool.

Blockchain/KYC/KYCDataRepository.swift
```diff
@@ -0,0 +1,75 @@
...
+//  Created by Chris Arriola on 8/14/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import RxSwift
+
+/// Repository for KYC related data. Accessing properties in this repository
+/// will be fetched from the cache (if available), otherwise, data will be fetched over
+/// the network and subsequently cached for faster access.
+@objc class KYCDataRepository: NSObject {
```
jstxx: Very good stuff.  Seeing as I'm already modifying settings to accept @thisisalexmcgregor structs this will go a long way to minimizing network calls in settings.

PR #402, opened 2018-08-15
style(KYC): UI polish for Address input scren

thisisalexmcgregor:
## Objective

This implements a few polish items for the address input screen.

## Description

* The `primaryButton` is always above the keyboard
* The currently in focus textField is at the top of the `UIScrollView` 
* The description labels are within the `UIScrollView`
* The `UISearchBar` is now above the `UIScrollView`

## How to Test

1. Build and run
2. Go to the debug KYC flow
3. Go to the address screen. (Note: you may need to hardcode this in `screenFor(pageType:)` in the `KYCCoordinator`)
4. Use the `UISearchBar`.
5. Enter values in the fields.
6. Observe the `primaryButton`. Observe the scrolling behavior.

## Screenshot

screen shot 2018-08-15 at 10 36 03 am
screen shot 2018-08-15 at 10 36 08 am

## Related Information

* Ticket Number: N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/Address/KYCAddressController.swift
+++ b/Blockchain/KYC/Address/KYCAddressController.swift
@@ -218,6 +218,10 @@ extension KYCAddressController: LocationSuggestionInterface {
         tableView.alpha = visibility.defaultAlpha
     }
 
+    func primaryButton(_ visibility: Visibility) {
+        primaryButtonContainer.alpha = visibility.defaultAlpha
+    }
+
     func searchFieldActive(_ isFirstResponder: Bool) {
         switch isFirstResponder {
         case true:
--- a/Blockchain/KYC/Address/KYCAddressController.swift
+++ b/Blockchain/KYC/Address/KYCAddressController.swift
@@ -145,6 +145,7 @@ class KYCAddressController: KYCBaseViewController, ValidationFormView, BottomBut
             self?.scrollView.setContentOffset(.zero, animated: true)
             guard let keyboard = self?.keyboard else { return }
             self?.keyboardWillHide(with: keyboard)
+            self?.keyboard = nil
         }
         NotificationCenter.when(.UIKeyboardWillShow) { [weak self] notification in
             let keyboard = KeyboardPayload(notification: notification)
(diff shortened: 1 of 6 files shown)


Blockchain/KYC/Address/KYCAddressController.swift
```diff
@@ -137,9 +143,12 @@ class KYCAddressController: KYCBaseViewController, ValidationFormView {
         NotificationCenter.when(.UIKeyboardWillHide) { [weak self] _ in
             self?.scrollView.contentInset = .zero
             self?.scrollView.setContentOffset(.zero, animated: true)
+            guard let keyboard = self?.keyboard else { return }
+            self?.keyboardWillHide(with: keyboard)
```
bchrisarriola: Should `self?.keyboard` be set to nil here?

Blockchain/KYC/Address/KYCAddressController.swift
```diff
@@ -154,7 +163,7 @@ class KYCAddressController: KYCBaseViewController, ValidationFormView {
             postalCode: postalCodeTextField.text ?? "",
             city: cityTextField.text ?? "",
             state: stateTextField.text ?? "",
-            country: countryTextField.text ?? ""
+            country: ""
```
bchrisarriola: Can this value be inferred from the search result?

thisisalexmcgregor: Yeah, it can be, *or* we can get it from the country picker. Though, we'd only get it if they used the auto-complete feature. So I guess we have to always get it from the country picker and inject that.

bchrisarriola: Yeah. I guess we can use the value taken from the country picker but there is the case where that won't match... I feel like we should keep the country field to avoid weird cases like that now thinking about it. Wdyt?

thisisalexmcgregor: Ah, so add the country field back in? I'm fine with that, just need to check with creative and get their input.

Blockchain/KYC/Address/KYCAddressController.swift
thisisalexmcgregor: Done 396f86a

Blockchain/KYC/Address/KYCAddressController.swift
thisisalexmcgregor: @bchrisarriola Per slack:
Definitely an edge case and legit. For now, we won't have the country input field and we'll inject the country that is provided from the country selection screen. Added a `// TODO:` and will circle back once all the screens are stitched together (auth etc.)

bchrisarriola: 💯

PR #408, opened 2018-08-16
feat(Homebrew-Shapeshift): IOS-1152 begin adding view controller architecture

woudini:
## Objective

Added files to be used with services and APIs with architecture similar to `KYCAddressController` and `KYCPersonalDetailsViewController`. Also added `FromToViewDelegateIntermediate` class to use in both the old and new exchange implementations.

## Description

The general approach is to have a view controller `HomebrewExchangeCreateViewController` interact with a coordinator `ExchangeTradeCoordinator` through a `ExchangeTradeDelegate` (from controller to coordinator) and `ExchangeTradeInterface` (from coordinator to controller).

Will add service and API properties to `ExchangeTradeCoordinator` and tests in future PRs (unless reviewers feel that it makes sense to add tests at this point).

## How to Test

Go to exchange (you may have to modify the `if` statement around this if you are in `dev` env https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/View%20Controllers/ConfirmStateViewController.m#L73) and select various `From` and `To` destinations. You should be able to execute an trade.

## Screenshot

N/A

## Related Information

* Ticket Number: 1152

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Homebrew/Exchange/FromToButtonDelegateIntermediate.swift
```diff
@@ -0,0 +1,54 @@
...
+//
+//  Created by kevinwu on 8/16/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+@objc class FromToButtonDelegateIntermediate: NSObject {
+    private let wallet: Wallet
+    private let navigationController: BCNavigationController
```
bchrisarriola: weak?

Blockchain/Homebrew/Exchange/ExchangeTradeCoordinator.swift
```diff
@@ -0,0 +1,28 @@
...
+        super.init()
+
+        if let controller = interface as? HomebrewExchangeCreateViewController {
+            controller.delegate = self
+        }
+    }
+}
+
+extension ExchangeTradeCoordinator: ExchangeTradeDelegate {
+    func onContinueButtonClicked() {
```
thisisalexmcgregor: Nitpick: `onContinueButtonTapped`

Pray we never have to support macOS. :P

thisisalexmcgregor:
Seems pretty straight forward to me. 

@bchrisarriola Yes I believe the intention is to have two separate VCs and the view that is shared is the `ExchangeCreateView` and likely the transaction list screen.

Blockchain/Homebrew/Exchange/FromToButtonDelegateIntermediate.swift
```diff
@@ -0,0 +1,54 @@
...
+//  FromToButtonCoordinator.swift
+//  Blockchain
+//
+//  Created by kevinwu on 8/16/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+@objc class FromToButtonDelegateIntermediate: NSObject {
```
thisisalexmcgregor: Pulled and tested, makes sense now what it does but initially I wasn't exactly sure what was going on, might be worth putting a comment above this class saying where it's used and what it's for.

Blockchain/Homebrew/Exchange/HomebrewExchangeCreateViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+extension HomebrewExchangeCreateViewController: ContinueButtonInputAccessoryViewDelegate {
+    func closeButtonClicked() {
+        exchangeCreateView.hideKeyboard()
+    }
+}
+
+extension HomebrewExchangeCreateViewController: UITextFieldDelegate {
```
thisisalexmcgregor: I know it's out of scope for this PR and you probably know this already however, I think that the `ExchangeCreateView` should have a protocol that houses all the functions that its delegate can conform to. That would allow us to not have `UITextFieldDelegate` but instead just one protocol and a function like:
`func exchangeView(_ exchangeView: ExchangeCreateView, didEnterText text: String)` and then maybe an enum type indicating which of the fields. 

Thoughts? If you like or prefer that, could be a good candidate for `// TICKET`

woudini:
>the intention is to have two separate VCs and the view that is shared is the ExchangeCreateView

Yes, added some docs.
>and likely the transaction list screen.

The code for this screen may not necessarily be shared due to rigidity of the existing code.

Blockchain/Homebrew/Exchange/HomebrewExchangeCreateViewController.swift
woudini: Created ticket and referenced 06a65e06587a63ecc1fca8082469d94705d1fb3e

Blockchain/Homebrew/Exchange/ExchangeTradeCoordinator.swift
woudini: Quite a few methods are named this way, will address as part of the refactoring ticket.

Blockchain/Homebrew/Exchange/HomebrewExchangeCreateViewController.swift
thisisalexmcgregor: Sounds good 👍

PR #409, opened 2018-08-16
feat(KYC Authentication): IOS-1104 caching authentication session token

bchrisarriola:
## Objective

Cache the KYC authentication session token in KYCAuthenticationService.

## Description

Previously, a new session token for authentication is requested each time when invoking `getKycSessionToken()`. With this change, we are now caching the previously fetched `KYCSessionTokenResponse` and so subsequent calls will now use that cached response. However, if the cached session token is expired, we'll fetch another one.

Writing tests for this is a bit involved—`KYCNetworkRequest` cannot be easily mocked and creating the changes to make it mock-able involves refactoring the network layer a bit. I can make that change in another PR, but since @jstxx & @achtenhagen are working on refactoring the network layer, I figured it would be good to hold off until that's in a good place (something to consider in the new arch).

Other changes:
* clearing the `BlockchainDataRepository` when logging out. Looking for suggestions as well on how we can genericize the cache.
* `KYCAuthenticationService` is no longer a singleton. I'm thinking `BlockchainDataRepository` should also not be a singleton and instead the lifecycle should live within `KYCCoordinator`. Looking for input/thoughts here.

## How to Test

* Add `|| host == "api.dev.blockchain.info"` in line 108 in NetworkManager to get around the cert pinning issue
* Comment out line 70 in `BlockchainDataRepository` so that a `KYCUser` is always fetched over the network
* Put a breakpoint on line 65 in `KYCAuthenticationService`, go in, out, and back in KYC and verify that the breakpoint is reached

## Screenshot

N/A

## Related Information

* Ticket Number: IOS-1104

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/Authentication/AuthenticationCoordinator.swift
+++ b/Blockchain/Authentication/AuthenticationCoordinator.swift
@@ -216,6 +216,8 @@ import RxSwift
 
         WalletManager.shared.close()
 
+        BlockchainDataRepository.shared.clearCache()
+
         let appCoordinator = AppCoordinator.shared
         appCoordinator.tabControllerManager.clearSendToAddressAndAmountFields()
         appCoordinator.closeSideMenu()
--- a/Blockchain/Data/BlockchainDataRepository.swift
+++ b/Blockchain/Data/BlockchainDataRepository.swift
@@ -18,7 +18,7 @@ import RxSwift
 
     private let authenticationService: KYCAuthenticationService
 
-    init(authenticationService: KYCAuthenticationService = KYCAuthenticationService.shared) {
+    init(authenticationService: KYCAuthenticationService = KYCAuthenticationService()) {
         self.authenticationService = authenticationService
     }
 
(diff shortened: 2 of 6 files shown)


woudini:
I assume `go in, out, and back in KYC` means using the `Debug` side menu item? It's crashing for me so I can't look at it right now (also happening on `dev`):
```
Reason: Could not find a storyboard named 'KYCVerifyIdentity' in bundle NSBundle

bchrisarriola:
Hm not sure why that's been crashing. I've been launching KYC through "Settings" -> "Terms of Service" -> "Debug" -> "Launch KYC". Keep in mind that you need to be authenticated into the wallet now to access KYC so launching KYC from the debug view in swipe to receive will not work.

woudini:
I see. Breakpoint is being reached as expected. Is there an easier way to back out of KYC other than by backgrounding the app?

bchrisarriola:
> I see. Breakpoint is being reached as expected. Is there an easier way to back out of KYC other than by backgrounding the app?

Not yet. :/ We should probably has a close button there.

woudini:
Added a comment on the design doc regarding a close button, but we should probably do it anyway.

PR #412, opened 2018-08-18
refactor(Side Menu): IOS-1094 side menu logic for showing exchange

bchrisarriola:
## Objective

Updating the logic in the side menu of the app for whether or not the "Exchange" option should be displayed.

## Description

Previously, we would show the "Exchange" option in the side menu if the inferred country of the wallet is _not_ in the list of blacklisted countries for shapeshift. Given our new homebrew support, this logic needs to be updated to do a union against countries that are in the "KYC" scope.

Changes:
* Rewrote `SideMenuViewController` in Swift and used storyboard
* Moved logic of what items to present in the side menu to `SideMenuPresenter`
* When tapping on "Debug KYC" in the side menu, this will now use `KYCCoordinator` instead of presenting `KYCVerifyIdentity`

There are still more changes that I'd like to make but in the interest of keeping this PR small, I'll do those changes in another PR. 

Upcoming changes:
* Write unit tests for `SideMenuPresenter`
* Update `KYCCountrySelectorPresenter` to use the logic in `WalletService` so we don't have duplicate logic on whether or not a country supports crypto-to-crypto exchange
* There's a bit of a lag when you open the side menu—"Exchange" will show up after a few seconds. This is because we now need to do a network call to get all the countries in the homebrew region. We can definitely prefetch. I'll get to that optimization step next.

## How to Test

Pull in these changes and poke around the side menu

## Screenshot

simulator screen shot - iphone 5s - 2018-08-17 at 19 25 19

## Related Information

* Ticket Number: IOS-1094

## Merge Checklist

- [X] All unit tests pass.


achtenhagen:
Currently all I am seeing is this:
simulator screen shot - iphone x - 2018-08-20 at 11 37 57

Is this PR currently out of date?

Blockchain/Network/API/WalletService.swift
```diff
@@ -55,18 +60,41 @@ import RxSwift
...
+    ///            and state is supported by a partner
+    func isInPartnerRegionForExchange(countryCode: String, state: String?) -> Single<Bool> {
+        return walletOptions.map { walletOptions in
+            guard let shapeshift = walletOptions.shapeshift else {
+                return false
             }
-            guard let json = $1 as? JSON else {
-                throw NetworkError.jsonParseError
+
+            let blacklistedCountries = shapeshift.countriesBlacklist ?? []
```
woudini: @jstxx pointed out that `blacklistedCountries ` should probably be `nil` instead of `[]` and that this function should return `false` if so. Probably the safe thing to do even though the equivalent JS function doesn't handle a `null` or `undefined` `countriesBlacklist`

Blockchain/Side Menu/SideMenuViewController.storyboard
```diff
@@ -0,0 +1,87 @@
...
+                                    </subviews>
+                                    <color key="backgroundColor" red="0.0" green="0.29019607843137252" blue="0.48627450980392156" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                    <constraints>
+                                        <constraint firstItem="OrS-f1-zpW" firstAttribute="centerY" secondItem="0df-tG-myy" secondAttribute="centerY" id="Jg6-41-c4C"/>
+                                        <constraint firstItem="OrS-f1-zpW" firstAttribute="leading" secondItem="0df-tG-myy" secondAttribute="leading" constant="58" id="xvR-61-rq8"/>
+                                    </constraints>
+                                </view>
+                                <prototypes>
+                                    <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" preservesSuperviewLayoutMargins="YES" selectionStyle="none" indentationWidth="10" reuseIdentifier="SideMenuCell" textLabel="GOI-Ww-29D" imageView="KYL-By-psN" rowHeight="54" style="IBUITableViewCellStyleDefault" id="YYw-Di-RgP" customClass="SideMenuCell" customModule="Blockchain" customModuleProvider="target">
```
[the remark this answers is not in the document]
bchrisarriola: Agreed that a non-none selectionStyle is preferred. I did this mostly because that's what the previous implementation was doing and didn't want to change it. I'll update that

Blockchain/Network/API/WalletService.swift
bchrisarriola: if `blacklistedCountries` is `nil` then shouldn't this function return true and not false? a `nil` `blacklistedCountries` should mean that all countries are whitelisted and so SS should be available everywhere

woudini: I think it should return `true` only if `blacklistedCountries` is an empty array. If it's `nil` that suggests something went wrong in the wallet-options endpoint, so it should be handled conservatively.

bchrisarriola: Cool. Makes sense

woudini:
Good catch on adding `showShapeshift`

PR #414, opened 2018-08-20
feat(Promises): Add currency pair model & promise based networking

jstxx:
## Objective

To simplify async network operations and give us a succinct way to deal with chaining API calls.
  
## Description

Using two dispatch queues, we pass the completion handler from the promised object down the chain so a CompleteFuture can have infinite .then methods chained. 

This is intended to work for the new exchange/KYC flow and be a scaffold for networking requests going forward.  Feedback is welcomed/encouraged while we consolidate and do away with the duplicitous networking stack and move towards a more universally extensible way to manage networking going forward with the goal of deprecating `NetworkManager` and `KYCNetworkRequest`.   

## How to Test

Open Settings and then tap `Identity Verification` It should start the KYC flow and should make the network call to the exchange. 

An example can be found in `KYCCordinator.swift`.  The main goal here is to start consolidation of network calls using a generic implementation of `URLSession` taking in a generic, decodable model.  Currently the network client supports `GET` requests.  

## Related Information

* Ticket Number:
IOS-1153

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/KYCCoordinator.swift
+++ b/Blockchain/KYC/KYCCoordinator.swift
@@ -86,9 +86,9 @@ protocol KYCCoordinatorDelegate: class {
                     
                     var url = URL(string: BlockchainAPI.shared.retailCoreUrl)!
                     url.appendPathComponent("markets/quotes/pairs")
-
+                    
                     var req = URLRequest(url: url)
-                    req.setValue(WalletManager.shared.wallet.kycLifetimeToken(), forHTTPHeaderField: "authorization")
+ 
                     self.exchangeFuture.then { (client) in
                         client.getAndParse(request: req, model: CurrencyPair.self) { result in
                             switch result {
--- a/Blockchain/KYC/KYCCoordinator.swift
+++ b/Blockchain/KYC/KYCCoordinator.swift
@@ -75,6 +75,17 @@ protocol KYCCoordinatorDelegate: class {
         var pairs: [String]
     }
     
+    func getUserVerificationStatus(handler: @escaping (KYCUser?, Bool) -> Void) {
+        disposable = BlockchainDataRepository.shared.kycUser
+            .subscribeOn(MainScheduler.asyncInstance) // network call will be performed off the main thread
+            .observeOn(MainScheduler.instance) // closures passed in subscribe will be on the main thread
+            .subscribe(onSuccess: { user in
+                handler(user, true)
+            }, onError: {  error in
+                handler(nil, false)
+            })
+    }
+    
     @objc func start(from viewController: UIViewController) {
         if user == nil {
             disposable = BlockchainDataRepository.shared.kycUser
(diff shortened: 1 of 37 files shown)


Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -51,22 +51,61 @@ protocol KYCCoordinatorDelegate: class {
...
     @objc func start(from viewController: UIViewController) {
         if user == nil {
             disposable = BlockchainDataRepository.shared.kycUser
                 .subscribeOn(MainScheduler.asyncInstance)
                 .observeOn(MainScheduler.instance)
                 .subscribe(onSuccess: { [unowned self] in
                     self.user = $0
+                    self.exchangeService = NetworkClient(session: URLSession.shared)
+                    
+                    var exchangeService: NetworkClient
```
achtenhagen: Why is this redeclared here?

jstxx: Bad merge, plz ignore.  Fixing.

Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -51,22 +51,56 @@ protocol KYCCoordinatorDelegate: class {
...
                 .observeOn(MainScheduler.instance)
                 .subscribe(onSuccess: { [unowned self] in
                     self.user = $0
+                    self.exchangeService = NetworkClient(session: URLSession.shared)
+                    
+                    var url = URL(string: BlockchainAPI.shared.retailCoreUrl)!
+                    url.appendPathComponent("markets/quotes/pairs")
+
+                    var req = URLRequest(url: url)
+                    req.setValue(WalletManager.shared.wallet.kycLifetimeToken(), forHTTPHeaderField: "authorization")
```
bchrisarriola: Shouldn't use lifetime token here

Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -51,22 +51,56 @@ protocol KYCCoordinatorDelegate: class {
...
     func start() {
         guard let rootViewController = UIApplication.shared.keyWindow?.rootViewController else {
             Logger.shared.warning("Cannot start KYC. rootViewController is nil.")
             return
         }
         start(from: rootViewController)
     }
-
+    
+    struct CurrencyPair: Decodable {
```
bchrisarriola: Move to its own file?

Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -51,22 +51,56 @@ protocol KYCCoordinatorDelegate: class {
     // MARK: Public
 
     weak var delegate: KYCCoordinatorDelegate?
+    var exchangeService: NetworkClient!
+    let exchangeFuture = CompletableFuture<NetworkClient>()
```
bchrisarriola: What's the reasoning for making `NetworkClient` into a `CompletableFuture`? The instantiation of this service seems like it can be synchronous.

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -114,6 +114,7 @@ final class KYCNetworkRequest {
         request.httpMethod = httpMethod
         request.addValue(HttpHeaderValue.json, forHTTPHeaderField: HttpHeaderField.accept)
         request.timeoutInterval = timeoutInterval
+        request.setValue(WalletManager.shared.wallet.kycLifetimeToken(), forHTTPHeaderField: "authorization")
```
bchrisarriola: This value should be provided as a dependency since retrieving the token can be async.

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -156,6 +157,7 @@ final class KYCNetworkRequest {
                 }
             }
             request.allHTTPHeaderFields = allHeaders
+            request.setValue(WalletManager.shared.wallet.kycLifetimeToken(), forHTTPHeaderField: "authorization")
```
bchrisarriola: Same comment here

Blockchain/KYC/KYCNetworkRequest.swift
```diff
@@ -181,6 +183,9 @@ final class KYCNetworkRequest {
                 HttpHeaderField.contentType: HttpHeaderValue.json,
                 HttpHeaderField.accept: HttpHeaderValue.json
             ]
+            
+            request.setValue(WalletManager.shared.wallet.kycLifetimeToken(), forHTTPHeaderField: "authorization")
```
bchrisarriola: Same

Blockchain/Network/NetworkClient.swift
```diff
@@ -0,0 +1,67 @@
...
+                default:
+                    ApiResult.failure(.serverError(http.statusCode)) -=> completion
+                }
+            }
+        }
+        task.resume()
+    }
+}
+
+func -=><T>(result: ApiResult<T>, completion: @escaping ApiCompletionBlock<T>) {
```
bchrisarriola: This is nice

Blockchain/Network/Promise.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+    
+    public func finally() {
+        // TODO: Implement
+        //regardless of errors, do this when everything completes
+    }
+    
+    public func failure() -> NSError {
+        promiseSequence.sync {
+            guard self.doneBlocks != nil else {
```
thisisalexmcgregor: Not sure what's happening here. You check if `doneBlocks` is not `nil` and then you don't do anything with them?

PR #421, opened 2018-08-21
feat(Homebrew-Shapeshift): add new subviews to exchange create view

woudini:
## Objective
Add additional subviews to Homebrew Shapeshift view as specified by design. 

## Description

Added a new class `AssetConversionRateView` and its viewModel `AssetConversionRateViewModel`, and used it conditionally in `ExchangeCreateView` with the parameter `withConversionView`.

## How to Test

To test Homebrew, cherry-pick 8cfb01692a4e2043ad11a1d8622813e0e3f5001f and go to Exchange
To test Shapeshift, cherry-pick 969038f31bd4451ff004ff511616d0940394f004 on top of 8cfb01692a4e2043ad11a1d8622813e0e3f5001f and go to Exchange

The above commits enable access to Shapeshift and add a mock view model for the `AssetConversionRateView`.

## Screenshot

simulator screen shot - iphone 6 - 2018-08-21 at 23 01 21

## Related Information

* Ticket Number: IOS-1159

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Extensions/UIView+Center.swift
```diff
@@ -0,0 +1,26 @@
...
+//  UIView+Center.swift
+//  Blockchain
+//
+//  Created by kevinwu on 8/18/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+public extension UIView {
```
bchrisarriola: Nice extensions 👍

Blockchain/Homebrew/Exchange/HomebrewExchangeCreateViewController.swift
```diff
@@ -25,13 +25,15 @@ class HomebrewExchangeCreateViewController: UIViewController {
...
-        let exchangeCreateView = ExchangeCreateView(frame: view.bounds)
+        exchangeCreateView = ExchangeCreateView(frame: view.bounds)
         view.addSubview(exchangeCreateView)
 
         exchangeCreateView.setup(
+            withConversionView: true,
             delegate: self,
             navigationController: self.navigationController as! BCNavigationController
         )
+        LoadingViewPresenter.shared.hideBusyView()
```
bchrisarriola: Why is this here? This should be placed elsewhere (i.e. whichever component called `LoadingViewPresenter.shared.showBusyView()`)

Blockchain/Homebrew/Exchange/Views/AssetConversionRateView.swift
```diff
@@ -0,0 +1,79 @@
...
+//  AssetConversionRateView.swift
+//  Blockchain
+//
+//  Created by kevinwu on 8/18/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+class AssetConversionRateView: UIView {
```
bchrisarriola: Why is this created programmatically? This should be moved to be created using a xib. We're trying to move views away from programmatically creating them for maintainability/scalability reasons unless there's no way around it or it is very expensive/involved to do so.

bchrisarriola: Taking inspiration from `Storyboardable`, you can:
1. Create a protocol `Nibable` in a file called `Nibable.swift`
2. Create a .xib titled `AssetConversionRateView`
3. Instantiate an `AssetConversionRateView` using `makeFromNib()`

Code for Nibable:

```
protocol Nibable: class {
    static var defaultNibName: String { get }
}

extension Nibable where Self: UIView {
    static var defaultNibName: String {
        return String(describing: self)
    }

    static func makeFromNib() -> Self {
        let nib = UINib(nibName: defaultNibName, bundle: Bundle.main)
        let contents = nib.instantiate(withOwner: nil, options: nil)
        return contents.first { $0 is Self } as! Self
    }
}

extension UIView: Nibable { }```

thisisalexmcgregor: Actually, you can also use `NibBasedView`. 

Just subclass that, and make a xib with the same name. In the `File's Owner` set it as whatever class you're making. In this case it'd be `AssetConversionRateView`. I did this with `PrimaryButtonContainer` if you need an example.

Blockchain/Extensions/UIView+Center.swift
thisisalexmcgregor: Autolayout? 

Actually if you use `NibBasedView` you can probably do a lot of this in a xib vs. having to do all the below programmatically.

Blockchain/Homebrew/Exchange/Views/AssetConversionRateView.swift
woudini: I don't have a good reason, just wanted to get it done quickly. Thanks for the code snippet, it was a lot easier than I thought to add. @thisisalexmcgregor I couldn't find a manual instantiation of `PrimaryButtonContainer` which is what I need in order to use it in `ExchangeCreateView`.

bchrisarriola: Forgot about `NibBasedView`! I do like `Nibable` though since it's a protocol and won't need to rely on inheritance (`NibBasedView` can also conform to `Nibable` for consolidation)

Blockchain/Homebrew/Exchange/HomebrewExchangeCreateViewController.swift
woudini: This should have been in the test branch - will remove.

Blockchain/Extensions/UIView+Center.swift
woudini: Removed extension

PR #423, opened 2018-08-22
feat(KYC): IOS-1167 attach authorization token to KYC personal details screen

bchrisarriola:
## Objective

Complete personal detail screen

## Description

Completing the personal details screen—previously, the network call made after tapping on "next" did not work because it was blocked by KYC authentication.

Other changes:
* Creating `KYCPageViewFactory` which is in charge of constructing a `KYCBaseViewController` given a `KYCPageType` and `KYCPayload`
* Creating `KYCPayload` and updating the `nextPageFromPageType`. The idea behind this is that when a page completes it can pass in a `KYCPayload` (e.g. thought of this so that we can pass around the data selected from `KYCCountrySelectionController`

## How to Test

1. Pull these changes in
2. Add `|| host == "api.dev.blockchain.info"` to `NetworkManager`
3. Launch KYC -> Select "Germany" -> Fill out personal details screen -> verify that you can get passed this step

NOTE: if you fail to get a `KYCUser` when you create a new wallet, _make sure you use a unique email address._ This is because the backend will complain if you try to create a new `KYCUser` if one has already been created for the email address provided (an email address already exists for the address "test@doesnotexist.com")

## Related Information

* Ticket Number: IOS-1167

## Merge Checklist

- [X] The PR uses a title supported by [.changelogrc](https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/.changelogrc#L6...L69).
- [ ] Areas of technical debt are marked with a `// TICKET:` comment that includes a ticket number.
- [X] All unit tests pass.
- [ ] You have added unit tests.
- [X] New files added are within the correct directory. (e.g. if a file is required for unit tests to compile, be sure it is added to the tests target.)

PR #424, opened 2018-08-22
feat(KYC): IOS-1163 Complete update address network call.

bchrisarriola:
## Objective

Complete KYC update address network call.

## Description

This PR completes the address update network call when a user tries to update their address during KYC. This was done by:
1. Providing the selected `KYCCountry` into `KYCAddressController`
2. Using `KYCAuthenticationService` to retrieve a session token followed by using that token to attach it to the "Authorization" header of the `PUT /users/current/address` network call

NOTE: this is reviewable but would like to merge #423 first before this PR (once #423 is merged, I'll update the base branch for this PR, I chose to base it off of `personal_details` to make reviewing easier)

## How to Test

1. Pull in these changes
2. Launch KYC
3. Select a homebrew supported country (e.g. Germany)
4. Go through personal details screen and hit next
5. Select an address and hit next and verify that you can go to the next step (ie entering your phone number)

REMINDER: if you fail to get a KYCUser when you create a new wallet, make sure you use a unique email address. This is because the backend will complain if you try to create a new KYCUser if one has already been created for the email address provided (an email address already exists for the address "test@doesnotexist.com")

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/Address/KYCAddressController.swift
+++ b/Blockchain/KYC/Address/KYCAddressController.swift
@@ -51,22 +51,30 @@ class KYCAddressController: KYCBaseViewController, ValidationFormView, BottomBut
     /// This is just for convenience purposes when iterating over the fields
     /// and checking validation etc.
     var validationFields: [ValidationTextField] {
-        get {
-            return [addressTextField,
-                    apartmentTextField,
-                    cityTextField,
-                    stateTextField,
-                    postalCodeTextField
-            ]
-        }
+        return [addressTextField,
+                apartmentTextField,
+                cityTextField,
+                stateTextField,
+                postalCodeTextField
+        ]
     }
 
-    var keyboard: KeyboardPayload? = nil
+    var keyboard: KeyboardPayload?
 
     // MARK: Private Properties
 
     fileprivate var locationCoordinator: LocationSuggestionCoordinator!
     fileprivate var dataProvider: LocationDataProvider!
+    private var user: KYCUser?
+    private var country: KYCCountry?
+
+    // MARK: KYCCoordinatorDelegate
+
+    override func apply(model: KYCPageModel) {
+        guard case let .address(user, country) = model else { return }
+        self.user = user
+        self.country = country
+    }
 
     // MARK: Lifecycle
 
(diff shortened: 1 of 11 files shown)


Blockchain/KYC/Address/KYCAddressController.swift
```diff
@@ -51,22 +51,30 @@ class KYCAddressController: KYCBaseViewController, ValidationFormView, BottomBut
...
     fileprivate var locationCoordinator: LocationSuggestionCoordinator!
     fileprivate var dataProvider: LocationDataProvider!
+    private var user: KYCUser?
+    private var country: KYCCountry?
+
+    // MARK: KYCCoordinatorDelegate
+
+    override func apply(model: KYCPageModel) {
+        guard case let .address(user, country) = model else { return }
+        self.user = user
```
bchrisarriola: Note: I'm setting the user here in preparation for mid-flow drop off handling (not doing anything with it right now)

thisisalexmcgregor: Should the textfields be filled in? For example if the user goes from this screen to er... whatever is next (phone?), and then taps the back button because they want to change the address or name, I'd expect the fields to be filled in again.

thisisalexmcgregor: Ah, so it already does that if you go back. But, it wouldn't do this if you were returning to KYC at a later date.

Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -37,9 +37,19 @@ protocol KYCCoordinatorDelegate: class {
...
+    private(set) var country: KYCCountry?
+
+    weak var delegate: KYCCoordinatorDelegate?
+
+    // MARK: - Private Properties
+
     fileprivate var navController: KYCOnboardingNavigationController!
 
-    fileprivate var user: KYCUser?
+    private let pageFactory = KYCPageViewFactory()
```
thisisalexmcgregor: I'd make the function on `KYCPageFactory` a class function so you don't need to keep an instance of it around.

bchrisarriola: I had that as a class function initially but converted it to an instance method so that it would be easier to mock if we want to test a component that depends on a `KYCPageViewFactory` (not actually mocked or tested right now though :/).

Blockchain/KYC/Address/KYCAddressController.swift
bchrisarriola: Yeah, the textfields would be filled in if a non-nil KYCUser was provided. For mid-flow drop off handling I was thinking that the `KYCCoordinator` would push all the previous view controllers into the navigation controller's stack, this way, going back to the previous screen would already be handled. wdyt?

PR #427, opened 2018-08-23
feat(Homebrew-Shapeshift): IOS-1152 add markets service to exchange coordinator

woudini:
## Objective

To more clearly define architecture of `ExchangeCoordinator`. Looking for feedback especially on how to organize it.

## Description

- Added `MarketsService` and `ExchangeService` to `ExchangeCoordinator`, as well as related files. Need to create models to continue. Much of what is added is WIP.
- Added `TradingPair` struct and unit tests.
- `HomebrewExchangeCreateController` and `ExchangeCreateViewController` will be renamed in a separate PR.

## How to Test

Not testable at the moment.

## Screenshot

N/A

## Related Information

* Ticket Number: IOS-1152

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Coordinators/ExchangeCoordinator.swift
```diff
@@ -107,8 +92,48 @@ import RxSwift
...
+    private let marketsService: MarketsService
+    private let exchangeService: ExchangeService
+
+    // MARK: - Interfaces
+    fileprivate weak var createInterface: ExchangeCreateInterface?
+
+    // MARK: - Lifecycle
+    private init(
+        walletManager: WalletManager = WalletManager.shared,
+        walletService: WalletService = WalletService.shared
```
bchrisarriola: Pass in marketsService and exchangeService as dependencies as well.

```
private init(
        walletManager: WalletManager = WalletManager.shared,
        walletService: WalletService = WalletService.shared,
        marketService: MarketsService = MarketsService(),
        exchangeService: ExchangeService = ExchangeService()
)
```

The reason this is preferred is so that you can mock/configure the dependencies if we were to unit test `ExchangeCoordinator`.

Blockchain/Homebrew/Exchange/Services/MarketsService.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+    private var dataSource: DataSource = .socket
+
+    var pair: TradingPair? {
+        didSet {
+            fetchRates()
+        }
+    }
+
+    private var socketMessageSubject: Observable<SocketMessage> {
```
bchrisarriola: rename to `socketMessageObservable`?

Blockchain/Homebrew/Exchange/Services/MarketsService.swift
```diff
@@ -0,0 +1,72 @@
...
+                return false
+            })
+        }
+    }
+
+    func fetchRates() {
+        switch dataSource {
+        case .socket: do {
+            let message = Quote(parameterOne: "parameterOne")
+            message.tryToEncode(encoding: .utf8, onSuccess: { encoded in
```
bchrisarriola: This `tryToEncode` method is a bit funky. I thought this was an async call given that you have to pass completion/error blocks. Might be good to update that method to just `throw` if the encoding fails. So the method signature would be:

```
func tryToEncode(encoding: String.Encoding) throws -> String {
    // ...
}
```

BlockchainTests/Homebrew/TradingPairTests.swift
```diff
@@ -0,0 +1,54 @@
...
+//  TradingPairTests.swift
+//  BlockchainTests
+//
+//  Created by kevinwu on 8/23/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import XCTest
+
+class TradingPairTests: XCTestCase {
```
bchrisarriola: 💯

BlockchainTests/Homebrew/TradingPairTests.swift
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    func testInitializerError() {
+        let pair = TradingPair(from: .bitcoin, to: .bitcoin)
+        XCTAssert(pair == nil, "Initializer should allow different asset types")
+    }
+
+    func testInitializerSuccess() {
+        let pair = TradingPair(from: .bitcoin, to: .ethereum)
+        XCTAssert(pair != nil, "Initializer should allow different asset types")
```
achtenhagen: There is a convenience function named `XCTAssertNotNil` that can be used for this purpose.

BlockchainTests/Homebrew/TradingPairTests.swift
```diff
@@ -0,0 +1,54 @@
...
+        super.setUp()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+    }
+
+    func testInitializerError() {
+        let pair = TradingPair(from: .bitcoin, to: .bitcoin)
+        XCTAssert(pair == nil, "Initializer should allow different asset types")
```
achtenhagen: Similarly, there is a convenience function named `XCTAssertNil` that can be used here.

Blockchain/Coordinators/ExchangeCoordinator.swift
```diff
@@ -107,8 +92,48 @@ import RxSwift
...
     }
+
+    private func showCreateExchangetype(type: ExchangeType) {
+        switch type {
+        case .homebrew:
+            // show homebrew
+            let exchangeCreateViewController = HomebrewExchangeCreateViewController()
+            exchangeCreateViewController.delegate = self
+            self.createInterface = exchangeCreateViewController
+            // present view controller
```
woudini: Wanted to get thoughts on this: would we have to keep a strong reference to `HomebrewExchangeCreateViewController` here? Its `delegate` is weak and the coordinator's `createInterface` is also weak. It doesn't seem to make sense to make either the `delegate` or `createInterface` strong references.

Blockchain/Homebrew/Exchange/Services/MarketsService.swift
woudini: Should `restMessageSubject` be `restMessageObservable` also?

Blockchain/Coordinators/ExchangeCoordinator.swift
bchrisarriola: I don't think we should keep a reference of `HomebrewExchangeCreateViewController`. If that view is in the foreground, the window will have a strong reference to it so as long as that's show in the screen the delegate's won't be nil

Blockchain/Homebrew/Exchange/Services/MarketsService.swift
woudini: I think the original idea behind `tryToEncode` was to not expose the `throw`. It would make more sense to just use `encode() throws -> Data`.  Thoughts?

bchrisarriola: Ah, that's right. Callers can just invoke the `encode()` method directly. 

I've typically seen methods in Swift that accepts blocks to fall in one of the following categories:
1. it is async and the completion block(s) will be invoked after the async op is complete
2. the block is invoked on a functional operation (map/filter/etc.)

...but not really as a way to get around exception handling. I think it's clearer to handle the exception in line so that the caller doesn't need to pass all subsequent execution in the block which makes code harder to comprehend.

But, if you've seen this pattern before, I'm happy to not push too hard on this :)

PR #430, opened 2018-08-24
feat(KYC): add Application Complete view controller

achtenhagen:
## Objective

Adds the *Application Completed* screen to the KYC flow.

## Description

Created class + storyboard for this static screen.

## How to Test

Not possible at the moment, unless the VC is instantiated directly from the KYCCoordinator.

## Screenshot

simulator screen shot - iphone x - 2018-08-24 at 14 17 56

## Merge Checklist

- [x] All unit tests pass.


bchrisarriola:
Actually look like the last item is being cut off on iPhone SE: https://cloudup.com/cKglF5H7mZ7

To get around this—should wrap in a `UIScrollView`.

thisisalexmcgregor:
Is there ever a time when one of those things in the list isn't complete? If so, it may need to be either a `UITableViewCell` that we configure with a model that has a title and a completion sate, or a custom view. Also I'm guessing the header values wouldn't say `Done and Done!`. But maybe you can't get to this screen without completing all the steps. 

For localization should the string values be set in the codebase?

achtenhagen:
@thisisalexmcgregor I'm not aware of the possibility to reach this state other than when the flow is completed, hence why I made the UI as static as possible. The header values do indeed say `Done and Done!` as per design. The only way to reach this screen is to complete all of the previous steps.

achtenhagen:
@bchrisarriola Content is already wrapped in a UIScrollView to support smaller devices.

PR #435, opened 2018-08-28
feat(CreateWalletView): IOS-1189 add privacy policy

woudini:
## Objective

- Add Privacy Policy link to Create Wallet View.
- Added `.com` and `.info` distinctions to `BlockchainAPI` `Hosts`, since the `BCWebViewController` was leaving the app due to the URLs being from `.com` instead of `.info`

## Description

~~Attempted to use this extension (https://github.com/blockchain/My-Wallet-V3-iOS/blob/kevin/IOS-1189/add_privacy_policy/Blockchain/KYC/KYCWelcomeController.swift#L100) to add a tappable view over attributed text, but it ends up being very offset for `.alignmentLeft`, even when the `textContainerOffset` is set to `0` as a user in the stackoverflow link suggests. Ended up using a pattern similar to the one in `PartnerExchangeConfirmViewController.m` to create tappable views over the Terms of Service and Privacy Policy. Unfortunately spent a lot of time on and havent found a way to fix this offset bug with a xib, possibly due to autoresizing. Moving this to `layoutSubviews` produces similar results:~~

screen shot 2018-08-28 at 2 43 48 pm

~The implementation in `PartnerExchangeConfirmViewController.m` is exact in width:~
screen shot 2018-08-28 at 2 54 10 pm

~~I personally have preferred programmatic views to xibs/storyboards and this is an example of how precise control of views is much more difficult to achieve when using them.~~
Used a new `UILabel` instance separate from the xib initialization to fix:
screen shot 2018-08-29 at 11 38 22 am

## How to Test

Go to create wallet view, tap close to Terms of Service and Privacy Policy.

## Screenshot

simulator screen shot - iphone 6 - 2018-08-28 at 14 49 34

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Modal Content Views/BCCreateWalletView.h
+++ b/Blockchain/Modal Content Views/BCCreateWalletView.h
@@ -23,6 +23,7 @@
 #import "BCRecoveryView.h"
 #import "Wallet.h"
 
+// TICKET: IOS-1232 Refactor/Modernize BCCreateWalletView.
 @interface BCCreateWalletView : BCModalContentView <UITextFieldDelegate, WalletDelegate> {
     IBOutlet UITextField *emailTextField;
     IBOutlet UITextField *passwordTextField;
--- a/Blockchain/Modal Content Views/BCCreateWalletView.h
+++ b/Blockchain/Modal Content Views/BCCreateWalletView.h
@@ -29,8 +29,8 @@
     IBOutlet UITextField *password2TextField;
     IBOutlet UILabel *passwordFeedbackLabel;
     IBOutlet UIProgressView *passwordStrengthMeter;
+    IBOutlet UILabel *agreementTappablePlaceholderLabel;
     IBOutlet UILabel *agreementLabel;
-    IBOutlet UILabel *agreementTappableLabel;
 }
 
 + (nonnull BCCreateWalletView *)instanceFromNib;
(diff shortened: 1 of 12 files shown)


thisisalexmcgregor:
Definitely worth putting:
`// TICKET: [Ticket Number] Refactor/Modernize BCCreateWalletView`.  
Made in 2012! Doesn't look like it'd take long to do it.

Blockchain/Modal Content Views/BCCreateWalletView.h
```diff
@@ -47,4 +45,5 @@
 @property(nonatomic) float passwordStrength;
 @property(nonatomic) BOOL isRecoveringWallet;
 @property(nonatomic) UIButton *createButton;
+@property(nonatomic) UILabel *agreementTappableLabel;
```
thisisalexmcgregor: Since `agreementTappableLabel` is just a `UILabel` and not a subclass, I think we should just call it `agreementLabel`.

Actually, weird why is there an `agreementLabel` IBOutlet and property? 

Probably not worth fixing it up but we can document that it needs rewritten.

woudini: Sorry yeah it's confusing.
- `agreementLabel` is the label that says "By creating a wallet you agree to the". It's just a plain label.
- I added `agreementTappableLabel` as a separate property to get around the autoresizing that was causing the offset widths issue.
- `agreementTappablePlaceholderLabel` is an IBOutlet label with the same frame and autoresizing as the old "Terms of Service" button. It's used as a reference frame for `agreementTappableLabel`.

Agree that it will be better to rewrite the whole thing but this seems like the best way to have a correct implementation and preserve the `frame`s of the existing subviews.

PR #440, opened 2018-08-29
fix(KYC): update page title on address screen

achtenhagen:
## Objective

Update the navigation bar title to match the designs.

## Screenshot

simulator screen shot - iphone x - 2018-08-29 at 11 20 15

## Merge Checklist

- [x] All unit tests pass.


achtenhagen:
Note: The black status bar is the result of this screen being presented via the debug menu on the pin entry screen (which uses the dark style). This does not occur during the actual KYC flow.

thisisalexmcgregor:
Got an idea for applying the titles to each screen. 

In `private func handlePageWillAppear(for type: KYCPageType)` you'll see the delegate call `func apply(model: KYCPageModel)`. Each screen at a minimum could have a `KYCPageModel` with just a title. 

Then on `viewWillDisappear` we can have a similar hook for `coordinator.handle(event: .viewWillDisappear)` and the title can be set to `nil`. I think that would handle the back button issue you mentioned to me the other day.

PR #444, opened 2018-08-29
feat(HB): IOS-1231 - Models and REST endpoints for exchange

thisisalexmcgregor:
## Objective

Adding some models and hooking up some REST endpoints for HB. 

## Description

This includes two new API's that are added to `ExchangeDependencies` 

* `RatesAPI`
* `TradeExecutionAPI`

The `RatesAPI` may not be necessary but, `TradeExecutionAPI` definitely is. With this you can get `TradeLimits`, submit an `Order` and get back a `Trade`. 

`RatesAPI` returns all `TradePairs` and `TradePairConfigurations` for each `TradePair`. Again, this may not be needed on the client...

## How to Test

1. Put a `!` in front of `isHomebrewSupported` in the `ExchangeCoordinator` to force the new exchange screen to load. 
2. In the function where dependencies are setup in `ExchangeListViewController` you can access either of the new services. 
3. Try out the new services and `print` the result. 

## Screenshot

N/A

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Homebrew/Exchange/Services/RatesService.swift
```diff
@@ -0,0 +1,97 @@
...
+    
+    fileprivate func configuration(forPair: TradingPair) -> Single<TradingPairConfiguration> {
+        guard let baseURL = URL(
+            string: BlockchainAPI.shared.retailCoreUrl) else {
+                return .error(RatesAPIError.generic)
+        }
+        
+        guard let endpoint = URL.endpoint(
+            baseURL,
+            pathComponents: ["markets", "quotes", forPair.stringRepresentation, "config"],
```
woudini: Add a `private struct PathComponents` with these static strings?

Blockchain/Homebrew/Exchange/Services/RatesService.swift
```diff
@@ -0,0 +1,97 @@
...
+    fileprivate func rates() -> Single<Rates> {
+        
+        guard let baseURL = URL(
+            string: BlockchainAPI.shared.retailCoreUrl) else {
+                return .error(RatesAPIError.generic)
+        }
+        
+        guard let endpoint = URL.endpoint(
+            baseURL,
+            pathComponents: ["markets", "quotes", "pairs"],
```
woudini: `private struct PathComponents`?

Blockchain/Homebrew/Exchange/Services/TradeExecutionService.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    fileprivate func limits() -> Single<TradeLimits> {
+        guard let baseURL = URL(
+            string: BlockchainAPI.shared.retailCoreUrl) else {
+                return .error(TradeExecutionAPIError.generic)
+        }
+        
+        guard let endpoint = URL.endpoint(
+            baseURL,
+            pathComponents: ["trades", "limits"],
```
woudini: `private struct PathComponents`?

thisisalexmcgregor:
@woudini Yeah `NetworkRequest` isn't final, I just needed it up and running quickly to get these endpoints available. It's at least decoupled from `KYCNetworkRequest` and can work with RxSwift (we have to use that to leverage the `KYCAuthenticationService`.

thisisalexmcgregor:
@woudini comments addressed. @bchrisarriola is going to take a look as well to give his RxSwift take.

Blockchain/Homebrew/Exchange/Services/RatesService.swift
```diff
@@ -0,0 +1,120 @@
...
+    init(service: KYCAuthenticationService = KYCAuthenticationService.shared) {
+        self.authentication = service
+    }
+    
+    deinit {
+        disposable?.dispose()
+        disposable = nil
+    }
+    
+    func getRates(withCompletion: @escaping ((Result<Rates>) -> Void)) {
```
bchrisarriola: Alternatively, you can also just have the component calling this method handle the subscription (so this method would just return `Single<Rates>`). This works too though 👍

Blockchain/Network/NetworkRequest.swift
```diff
@@ -0,0 +1,146 @@
...
+//  NetworkRequest.swift
+//  Blockchain
+//
+//  Created by Alex McGregor on 8/28/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import RxSwift
+
+struct NetworkRequest {
```
bchrisarriola: This is sweet! 💯

Blockchain/Network/NetworkRequest.swift
```diff
@@ -0,0 +1,146 @@
...
+        case put = "PUT"
+        case delete = "DELETE"
+    }
+    
+    let method: NetworkMethod
+    let endpoint: URL
+    let token: String?
+    let body: Data?
+    
+    init(endpoint: URL, method: NetworkMethod, body: Data?, authToken: String? = nil) {
```
bchrisarriola: Thoughts on making this networking component generic? Meaning that this shouldn't "know" about the concept of an authentication token. Instead, this will accept a map of headers instead.

Blockchain/Network/NetworkRequest.swift
```diff
@@ -0,0 +1,146 @@
...
+        type: ResponseType.Type
+        ) -> Single<ResponseType> {
+        let request = self.init(endpoint: url, method: .get, body: body, authToken: token)
+        return Single.create(subscribe: { (observer) -> Disposable in
+            request.execute(expecting: ResponseType.self, withCompletion: { (result, responseCode) in
+                switch result {
+                case .success(let value):
+                    observer(.success(value))
+                case .error(let error):
+                    if let value = error {
```
bchrisarriola: Should this just call `observer(.error(error)) ` regardless if `error == nil`? This way, the observer's _onError` method will always be invoked when there is an error. At the moment, that isn't the case.

Blockchain/Network/NetworkRequest.swift
```diff
@@ -0,0 +1,146 @@
...
+        guard let urlRequest = URLRequest() else {
+            withCompletion(.error(nil), responseCode)
+            return
+        }
+        guard let session = NetworkManager.shared.session else {
+            withCompletion(.error(nil), responseCode)
+            return
+        }
+        
+        session.dataTask(with: urlRequest) { (payload, response, error) in
```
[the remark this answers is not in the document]
thisisalexmcgregor: Done. Side effect, that function is now a `mutating` function. I don't think that'll be an issue though.

Blockchain/Network/NetworkRequest.swift
```diff
@@ -0,0 +1,155 @@
...
+//  Blockchain
+//
+//  Created by Alex McGregor on 8/28/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import RxSwift
+
+/// TICKET: IOS-1242 - Condense HttpHeaderField
+/// and HttpHeaderValue into enums and inject in a token
```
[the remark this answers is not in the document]
bchrisarriola: Sounds good

Blockchain/Homebrew/Exchange/Models/TradeLimits.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+extension String {
+    
+    enum ConversionError: Error {
+        case generic
+    }
+    
+    func toDouble() throws -> Double {
```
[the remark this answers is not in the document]
bchrisarriola: Were you thinking of converting this to `NSDecimalNumber` as @woudini suggested in this PR? Not sure if you were planning on adding more changes. Otherwise, this PR lgtm.

PR #445, opened 2018-08-30
fix(KYC): add close button to account status view

achtenhagen:
## Objective

Add the missing close button to the account status view.

## How to Test

Launch the account status screen and dismiss it. Also observe that the entire KYC flow will be dismissed as well.

## Screenshot

simulator screen shot - iphone x - 2018-08-30 at 09 47 07

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/Information/KYCInformationController.swift
+++ b/Blockchain/KYC/Information/KYCInformationController.swift
@@ -40,6 +40,14 @@ final class KYCInformationController: KYCBaseViewController {
         return controller
     }
 
+    // MARK: - IBActions
+
+    @IBAction func dismiss(_ sender: Any) {
+        dismiss(animated: false, completion: {
+            self.coordinator.end()
+        })
+    }
+
     // MARK: - Lifecycle Methods
 
     override func viewDidLoad() {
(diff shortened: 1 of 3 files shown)


Blockchain/KYC/Information/KYCInformationController.swift
```diff
@@ -40,6 +40,14 @@ final class KYCInformationController: KYCBaseViewController {
         return controller
     }
 
+    // MARK: - IBActions
+
+    @IBAction func dismiss(_ sender: Any) {
+        dismiss(animated: false, completion: {
```
woudini: I think `animated: true` is slightly better, but this flow is designed pretty oddly (ideally wouldn't have to end up with a presented controller dismissing its parent). Will leave up to you.

achtenhagen: I decided to use `animated: false` such that the flow is dismissed more quickly. Otherwise the user needs to witness two animations (one of them reverting back to the previous screen, which feels odd). But I agree, the flow is not designed in an ideal way. Prefer to *not* animate the presented controller at this time (unless/until the design changes at some point).

PR #448, opened 2018-08-30
feat(Homebrew-Shapeshift): IOS-1176 connect keypad and amount labels through interactor and presenter

woudini:
## Objective

Connect number keypad output to amount label.

## Description

- Connected interactor with presenter
- Added a dependency `ExchangeInputs` to house the model and model interaction logic (this should be separated later as one or the other becomes more complex).
- Subviews need to be moved up, but that should be done in a later PR since this one is pretty big already.

## How to Test

Open exchange with a wallet without any trade history, then tap on numbers, decimals, and backspace.

## Screenshot

simulator screen shot - iphone 6 - 2018-08-30 at 17 09 00

## Merge Checklist

- [x] All unit tests pass.


BlockchainTests/Homebrew/NumberInputTests.swift
```diff
@@ -0,0 +1,81 @@
...
+//  NumberInputTests.swift
+//  BlockchainTests
+//
+//  Created by kevinwu on 8/29/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import XCTest
+
+class NumberInputTests: XCTestCase {
```
thisisalexmcgregor: 🎉 🎈

Blockchain/Coordinators/ExchangeCoordinator.swift
```diff
@@ -127,45 +134,56 @@ struct ExchangeServices: ExchangeDependencies {
...
+            } else {
+                navigationController?.pushViewController(exchangeCreateViewController, animated: true)
+            }
         default:
             // show shapeshift
             Logger.shared.debug("Not yet implemented")
         }
     }
 
+    func showCreateExchange() {
```
thisisalexmcgregor:
```
func showCreateExchange() {
        showCreateExchangetype(type: .homebrew)
    }
```
A few things here. The function is a bit misleading. I guess it should maybe be renamed `showHomebrewExchange`. That being said, I think the coordinator should have a `handle(_ event: ExchangeCoordinatorEvent)` function. When we wrap the feature, if there's only one or two events, then we can probably just make two separate functions. 

Also this function is called here:
```
func showNewExchange(animated: Bool) {
        coordinator.showCreateExchange()
    }
```

That `animated` property should be passed up. This is for if they tap on `Exchange` and *do not* have any trades in their history, it should show the HB exchange screen (rather than the exchange list). This would present _without_ an animation to give the appearance that it was never pushed on, if that makes sense. We do this in Shapeshift. 

So it'd look like this:
`coordinator(handle: .createHomebrewExchange(animated))`

Blockchain/Coordinators/ExchangeCoordinator.swift
```diff
@@ -198,32 +216,13 @@ struct ExchangeServices: ExchangeDependencies {
...
-            self.createInterface?.exchangeRateUpdated("rate")
-        }, onError: { (error) in
-            Logger.shared.debug("Could not get exchange rates: \(error.localizedDescription)")
-        })
+//        disposable = self.marketsService.rates.subscribe(onNext: { [unowned self] rate in
+//            // WIP
+//            self.createInterface?.exchangeRateUpdated("rate")
+//        }, onError: { (error) in
+//            Logger.shared.debug("Could not get exchange rates: \(error.localizedDescription)")
+//        })
```
thisisalexmcgregor: ❓

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Created by kevinwu on 8/28/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+protocol ExchangeCreateInterface: class {
+    func expandRatesView()
+    func updateInputLabels(primary: String?, secondary: String?)
+    func updateRates(first: String, second: String, third: String)
```
thisisalexmcgregor: Minor thing but, there's some slight inconsistency here. One function is updating input labels, the other is updating rates but, I don't know if it's labels, a button, etc. Since it's functions for an interface it should be a bit more specific.

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
```diff
@@ -0,0 +1,27 @@
...
+}
+
+protocol ExchangeCreateInput: NumberKeypadViewDelegate {
+    func toggleFiatInput()
+    func ratesButtonTapped()
+    func useMinimumButtonTapped()
+    func useMaximumButtonTapped()
```
thisisalexmcgregor: These functions kinda read like an interface. Maybe something like `useMinimumAssetAmount()` or something like that? 

Is `fiatToggle` a `UISwitch`? Or is it a `UIButton` that displays either `To Fiat` or `To Crypto`? (I'm paraphrasing here.) If it's a `UIButton` maybe it should be `displayCurrencyTypeTapped`? Toggle just feels like a switch, and it feels like if it _is_ a switch then a parameter is missing with the source of truth as to whether or not the switch is on or off. 

Lastly, probably worth a comment above this protocol indicating why it needs to conform to `NumberKeypadViewDelegate`

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
```diff
@@ -0,0 +1,27 @@
...
+}
+
+protocol ExchangeCreateOutput: class {
+    func updatedInput(primary: String?, secondary: String?)
```
thisisalexmcgregor: Is the behavior different if both of these values are `nil`? Can they be? If so, maybe a separate function should exist that better defines that behavior?

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
thisisalexmcgregor: Oh, and are "first, second, third" labels? Implementation is empty in the `UIViewController` is why I ask

Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
```diff
@@ -12,18 +12,19 @@ class ExchangeCreateViewController: UIViewController {
...
+    @IBOutlet private var secondaryAmountLabel: UILabel!
+    @IBOutlet private var useMinimumButton: UIButton!
+    @IBOutlet private var useMaximumButton: UIButton!
+    @IBOutlet private var exchangeRateButton: UIButton!
+    @IBOutlet private var exchangeButton: UIButton!
     // MARK: - IBActions
 
-    @IBAction func fiatToggleTapped(_ sender: Any) {
-
+    @IBAction private func fiatToggleTapped(_ sender: Any) {
```
thisisalexmcgregor: Nit, but a good example as to why `toggle` can be misleading.

Since it's `sender: Any`, I can't tell if it's a `UISwitch` or if it's a `UIButton`. I wonder which one it is because the function signature is `toggle` 🙇

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Blockchain
+//
+//  Created by kevinwu on 8/28/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+protocol ExchangeCreateInterface: class {
+    func expandRatesView()
```
thisisalexmcgregor: `updateExpandRatesView(visibility: Visibility)`?
I'm assuming they'll want to be able to hide and show it, not just show it

Blockchain/Homebrew/Exchange/DataProviders/ExchangeInputs.swift
```diff
@@ -0,0 +1,58 @@
...
+
+protocol ExchangeInputsAPI: class {
+    var activeInput: NumberInputDelegate { get }
+    var lastOutput: String? { get }
+    var conversionRate: NSDecimalNumber? { get set }
+
+    func add(character: String)
+    func backspace()
+    func addDecimal()
+    func toggleInput()
```
thisisalexmcgregor: `toggle` feels like there should be a parameter in there. Something somewhere is maintaining state I'm guessing

Blockchain/Homebrew/Exchange/DataProviders/ExchangeInputs.swift
```diff
@@ -0,0 +1,58 @@
...
+import Foundation
+
+protocol ExchangeInputsAPI: class {
+    var activeInput: NumberInputDelegate { get }
+    var lastOutput: String? { get }
+    var conversionRate: NSDecimalNumber? { get set }
+
+    func add(character: String)
+    func backspace()
+    func addDecimal()
```
thisisalexmcgregor: Does `backspace` account for removing the decimal?

Blockchain/Homebrew/Exchange/DataProviders/ExchangeInputs.swift
```diff
@@ -0,0 +1,58 @@
...
+//  Created by kevinwu on 8/27/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+protocol ExchangeInputsAPI: class {
+    var activeInput: NumberInputDelegate { get }
+    var lastOutput: String? { get }
+    var conversionRate: NSDecimalNumber? { get set }
```
thisisalexmcgregor: I think `Decimal` is the Swift version right? Should bridge between the APIs. (I think)

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -0,0 +1,54 @@
...
+    init(dependencies: ExchangeDependencies) {
+        self.markets = dependencies.markets
+        self.inputs = dependencies.inputs
+    }
+}
+
+extension ExchangeCreateInteractor: ExchangeCreateInput {
+    func toggleFiatInput() {
+        inputs.toggleInput()
+        self.output?.updatedInput(primary: inputs.activeInput.input, secondary: inputs.lastOutput)
```
thisisalexmcgregor: ~self.~

Blockchain/Coordinators/ExchangeCoordinator.swift
woudini: Will delete

Blockchain/Homebrew/Exchange/DataProviders/ExchangeInputs.swift
woudini: I think `Decimal` is the swift version of `NSDecimal`. Tried to create an NSDecimal and the compiler errored `'NSDecimal' has been renamed to 'Decimal'`

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
woudini: Retracted feedback?

thisisalexmcgregor: oh, no, sorry, was wondering why we need `self.` here. I can see how that'd be confusing lol.

Blockchain/Homebrew/Exchange/DataProviders/ExchangeInputs.swift
woudini: It does, although looking at this again I think it would be better to remove `addDecimal()` and handle it in `add(character: String)`. Will update.

Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
woudini: I see. Doesn't seem like there's a good way to name it and imply a boolean state. Will rename

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
woudini: Not sure yet, depends on how we will want to handle input conversion errors. What separate function did you have in mind?

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
woudini: Yes, will update to `updateRateLabels(`. Thanks for the feedback

Blockchain/Homebrew/Exchange/DataProviders/ExchangeInputs.swift
woudini: Will rename to `changeInput()`, although we lose the boolean implication.

PR #449, opened 2018-08-31
feat(HB): Adding TradingPairView

thisisalexmcgregor:
## Objective

`TradingPairView` is used in a few different places in the app. It supports applying animatable presentation updates and transition updates. 

## Description

This PR adds:
* `AnimatablePresentationUpdate` - This makes applying animated UI updates a bit simpler
* `TransitionPresentationUpdate` - This is similar to `AnimatablePresentationUpdate` but is for animating updating a `UIImage` or a button title (e.g. on a `UIButton`)
* `TradingPairView` - Used in a few different places in HB.

How do you use it?

There's a few ways.

This will just give you the generic view with the `AssetTypes` logos and brand names:

`let pair = TradingPair(from: .bitcoin, to: .ethereum)`
`tradingPairView.apply(pair: pair)`

If you want to animate changes to the view, this is how you do it:

tradingpairviewdemo

These update types aren't final and we may want to make them more granular and specific to HB but for now these at least demonstrate how we want to handle updating this view.

Tapping on the buttons calls its `TradingPairViewDelegate`

## How to Test

N/A

## Screenshot

N/A - See Channel

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Models/ViewModels/PresentationUpdate/AnimatablePresentationUpdate.swift
```diff
@@ -0,0 +1,95 @@
...
+            )
+            
+        case .standard(let duration):
+            UIView.animate(
+                withDuration: duration,
+                delay: 0,
+                options: [
+                    .beginFromCurrentState,
+                    .allowUserInteraction
+                ],
```
woudini: Thoughts on combining `.beginFromCurrentState` and `.allowUserInteraction` into something like `standardOptions`? Not sure if it's very clean to do add an enum `AnimationParameterOptions` with something like `case .crossFade: standardOptions.append(.transitionCrossDissolve)`

thisisalexmcgregor: Agreed that it's in a couple of places so kinda not really doing DRY there. I would but since the call-site still only has `.crossFade` or `.none` I wouldn't worry about it too much. If we end up have a bunch more than these than yeah, I agree, probably worth buttoning it up a bit. 

Also, possible we may cut the animation altogether if creative isn't down 🤷‍♂️  which in that case we'd just use `.none`.

Blockchain/Models/ViewModels/PresentationUpdate/TransitionPresentationUpdate.swift
```diff
@@ -0,0 +1,55 @@
...
+
+/// This is used in `TransitionPresentationUpdate`.
+/// You use this parameter value to change the style
+/// of the animation applied to the UI update.
+enum TransitionParameter {
+    
+    case crossFade(duration: TimeInterval)
+    case none
+    
+    func perform(with: UIView, animations: @escaping () -> Void) {
```
woudini: Thoughts on renaming to `(with view: UIView`? `with: with` on line 40 is a bit difficult to understand.

thisisalexmcgregor: yep, I can do that.

Blockchain/Homebrew/Exchange/Views/TradingPairView.swift
```diff
@@ -0,0 +1,140 @@
...
+    
+    // MARK: Private
+    
+    fileprivate func handle(_ update: ViewUpdate) {
+        switch update {
+        case .statusTintColor(let color):
+            iconStatusImageView.tintColor = color
+            
+        case .statusVisibility(let visibility):
+            iconStatusImageView.alpha = visibility.defaultAlpha
```
woudini: There needs to be another `iconStatusImageView` next to the `Receive` label or this one needs to somehow be able to move back and forth between the labels - not sure if I missed that functionality here but just making sure.

woudini: Confirmed with creative that it moves to indicate that the input describes the desired amount to either exchange or receive.

thisisalexmcgregor: Yup, updating now.

thisisalexmcgregor:
@woudini Done:
d86228d
73b8790

Blockchain/Homebrew/Exchange/Views/TradingPairView.swift
```diff
@@ -0,0 +1,147 @@
...
+    // MARK: Public
+    
+    func apply(pair: TradingPair, animation: AnimationParameter = .none, transition: TransitionParameter = .none) {
+        let presentationUpdate = TradingPresentationUpdate(
+            animations: [
+                .backgroundColors(from: pair.from.brandColor, to: pair.to.brandColor),
+                .statusTintColor(.green),
+                .swapTintColor(.grayBlue),
+                .toStatusVisibility(.visible),
+                .fromStatusVisibility(.hidden)
```
woudini: Was thinking that `toIconStatusImageView` and `fromIconStatusImageView` would be switched - it seems more intuitive for the `Exchange` dot to be the `from` and the `Receive` dot would be the `to`, and for the visibilities to be switched here. Thoughts? Looks good otherwise.

thisisalexmcgregor: Ah, yeah I guess my `to` `from` paradigm got thrown off with this new behavior. So it can take a `TradingPair` which has a `to` and `from`. I always assumed that the `to` would be on the left and the `from` be on the right. 

Let me rename some of these parameters and outlets.

thisisalexmcgregor: @woudini 
Alrighty, made it basic and just used left and right. Might change it later to give it more context but, doesn't look like I can or should make much assumptions there lol. 
a1113ea

PR #457, opened 2018-09-04
fix(KYC): IOS-1187 show correct verification status for new wallets

achtenhagen:
## Objective

Fix the default identity verification status for newly created wallets.

## Description

Previously, the case for a new user was handled in the same way as a user with a `failed` state.

## How to Test

Create a new wallet and launch the settings view. Observe that the label inside the identity verification cell reads `Unverified` as opposed to `Failed`.

simulator screen shot - iphone x - 2018-09-04 at 16 04 25

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/View Controllers/Settings/SettingsCell.swift
+++ b/Blockchain/View Controllers/Settings/SettingsCell.swift
@@ -89,18 +89,14 @@ import UIKit
 
         self.detailTextLabel?.frame.origin.x = screenWidth - detailWidth! - 40
 
-        
         self.detailTextLabel?.font = UIFont(name: Constants.FontNames.montserratSemiBold, size: Constants.FontSizes.Tiny)
 
-        
 //        self.detailTextLabel?.layer.cornerRadius = 4
 //        self.detailTextLabel?.layer.masksToBounds = true
 //        self.detailTextLabel?.backgroundColor = .orange
 //        self.detailTextLabel?.textColor = .white
 //        sizeToFit()
 //        layoutIfNeeded()
-    
-        
     }
 }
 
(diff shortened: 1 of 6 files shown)


Blockchain/View Controllers/Settings/SettingsCell.swift
```diff
@@ -89,18 +89,14 @@ import UIKit
 
         self.detailTextLabel?.frame.origin.x = screenWidth - detailWidth! - 40
 
-        
         self.detailTextLabel?.font = UIFont(name: Constants.FontNames.montserratSemiBold, size: Constants.FontSizes.Tiny)
 
-        
 //        self.detailTextLabel?.layer.cornerRadius = 4
```
bchrisarriola: Delete?

Blockchain/View Controllers/Settings/SettingsViewController.swift
```diff
@@ -79,7 +77,6 @@ MobileNumberDelegate, WalletAccountInfoDelegate {
         super.viewDidAppear(animated)
         tableView.reloadData()
         self.walletManager.accountInfoDelegate = self
-        preparedIdentityStatus == false
```
bchrisarriola: 🤔

achtenhagen: What are your thoughts 💭?

bchrisarriola: Just strange that this was even checked in (this line wasn't doing anything useful and the property wasn't actually used).

achtenhagen: Ah yes, completely unnecessary 🕵️‍♂️

PR #465, opened 2018-09-07
feat(Homebrew): Exchange Detail Screens and related UI components

thisisalexmcgregor:
## Objective

This PR includes all ViewModels, custom cells, and the `ExchangeDetailViewController`. Note that how the view models are build in `ExchangeDetailCoordinator` (or even that `ExchangeDetailCoordinator` exists) may change once this is hooked up to a live endpoint/flow. I had to make some assumptions to get this built and test all the custom views. 

## Description

There's only one `UIViewController` and that's `ExchangeDetailViewController`. To build it, you need a `ExchangeDetailViewController.PageModel`. There are three types, one for each screen. 

The `ExchangeDetailCoordinator` makes all the `ExchangeCellModels`. There are three types:
1. `Plain` - This is the cell with a subject on the left and a value on the right. It's used on all three screens.
2. `Text` - This is the text at the bottom of two of the screens. 
3. `TradingPair` - This is a cell that houses a `TradingPairView`

Each `ExchangeCellModel` maps to one and only one cell. All cells are subclasses of `ExchangeDetailCell`. It has two public functions.

```
func configure(with model: ExchangeCellModel) {
        assertionFailure("Should be implemented by subclasses")
    }
class func heightForProposedWidth(_ width: CGFloat, model: ExchangeCellModel) -> CGFloat {
        // Cells should override this method.
        return 0.0
    }
```

The first is for configuring the cell. The second is for calculating its height (which is dynamic). 

The `ExchangeDetailViewController` tells the `ExchangeDetailCoordinator` when it has loaded and passes up its `PageModel`. From that the coordinator builds all the models and calls its delegate:
```
func coordinator(_ detailCoordinator: ExchangeDetailCoordinator, updated models: [ExchangeCellModel])
```
It also updates the `ExchangeDetailInterface`.

That's pretty much it. Some screens have header views and footer views and the controller knows what to show based on its `PageModel`. 

## How to Test

In the `AppCoordinator` at line 227 you can see the debug code for displaying these screens. If you launch the app and select `Debug KYC` in the hamburger menu, you'll see the screen. You'll need to swap out the correct controller type to see each screen. 

Note that all screens show stubbed out sample data. Most of the sample data will be removed once this is merged. 

Also note, some of the models aren't final given that this isn't actually hooked up to the API yet. I imagine they'll need to be changed a bit. 

I would say for all these screens the UI is at about 90-95% accurate. Granted there's three different screens so I probably missed something. 
simulator screen shot - iphone 8 - 2018-09-06 at 21 22 30

simulator screen shot - iphone 8 - 2018-09-06 at 21 25 14

simulator screen shot - iphone 8 - 2018-09-06 at 21 26 01

## Merge Checklist

- [x] All unit tests pass.


woudini:
Swiftlint violations
screen shot 2018-09-07 at 5 01 05 pm

Blockchain/Coordinators/AppCoordinator.swift
```diff
@@ -224,7 +224,16 @@ extension AppCoordinator: SideMenuViewControllerDelegate {
             handleExchange()
         // TODO: remove item
         case .kyc:
-            handleLaunchKYC()
+            // TODO: Do not merge into `dev`.
```
woudini: Remove? I can revert the commit to test.

thisisalexmcgregor: Done

thisisalexmcgregor:
@woudini Ah, kk. Creative can take a look. That bit of extra scrollable space is only there because the spacing of all the items matches the comps. They may have not considered smaller displays. Good call out though, I'll let them know.

PR #469, opened 2018-09-10
fix(KYC): pass country code instead of name (IOS-1270)

achtenhagen:
## Objective

Pass the country code instead of the country name when updating the user.

## Merge Checklist

- [ ] All unit tests pass.


Blockchain/KYC/Address/KYCAddressController.swift
```diff
@@ -179,7 +179,7 @@ class KYCAddressController: KYCBaseViewController, ValidationFormView, BottomBut
             postalCode: postalCodeTextField.text ?? "",
             city: cityTextField.text ?? "",
             state: stateTextField.text ?? "",
-            country: country?.name ?? user?.address?.country ?? ""
+            country: country?.code ?? user?.address?.country ?? ""
```
bchrisarriola: What about also renaming `UserAddress.country` to `UserAddress.countryCode`? Either that or adding a comment to that field so it's clear that's what that value represents

PR #471, opened 2018-09-10
fix(KYC): IOS-1269 avoid using cached token if user ID does not match.

bchrisarriola:
## Objective

Check that the cached Nabu session token user_id matches the current wallet nabu user_id.

## Description

We need to check that the cached session token matches the current wallet nabu user_id so that when you log-out and create a new wallet, the cached session token for the previous wallet/user is not used.

## How to Test

1. Create a wallet
2. Go through KYC and complete all the steps
3. Log-out
4. Create a new wallet
5. Try to go through KYC again (verify that going through KYC will start the process from the start)

N/A

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Authentication/NabuAuthenticationService.swift
+++ b/Blockchain/Homebrew/Authentication/NabuAuthenticationService.swift
@@ -47,27 +47,42 @@ final class NabuAuthenticationService {
     // MARK: - Private Methods
 
     private func getSessionTokenIfNeeded(from userResponse: NabuCreateUserResponse) -> Single<NabuSessionTokenResponse> {
-        // Use cached session token if not expired, otherwise, request a new one
-        guard let sessionToken = cachedSessionToken.value,
-            let expiresAt = sessionToken.expiresAt, Date() < expiresAt else {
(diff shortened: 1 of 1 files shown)


Blockchain/Homebrew/Authentication/NabuAuthenticationService.swift
```diff
@@ -47,29 +47,44 @@ final class NabuAuthenticationService {
...
-                ).do(onSuccess: { [unowned self] in
-                    self.cachedSessionToken.accept($0)
-                })
+
+        guard let sessionToken = cachedSessionToken.value else {
+            return requestNewSessionToken(from: userResponse)
+        }
+
+        // Make sure cached session token is for this user
+        guard userResponse.userId == sessionToken.userId else {
```
achtenhagen: Do you want to combine these two `guard` statements?

bchrisarriola: I definitely considered that but I prefer to keep these as separate for (1) readability and (2) in case we want to handle these differently in the future.

PR #478, opened 2018-09-11
fix(KYC): remove back button from application completed screen (IOS-1268)

achtenhagen:
## Objective

Remove the back button from the _Application Completed_ screen.

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/KYCApplicationCompleteController.swift
+++ b/Blockchain/KYC/KYCApplicationCompleteController.swift
@@ -27,6 +27,7 @@ final class KYCApplicationCompleteController: KYCBaseViewController, Progressabl
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        self.navigationItem.leftBarButtonItem = UIBarButtonItem()
         setupProgressView()
     }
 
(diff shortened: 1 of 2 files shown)


Blockchain/KYC/KYCApplicationCompleteController.swift
```diff
@@ -27,6 +27,7 @@ final class KYCApplicationCompleteController: KYCBaseViewController, Progressabl
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        self.navigationItem.leftBarButtonItem = UIBarButtonItem()
```
bchrisarriola: `KYCApplicationCompleteController` probably shouldn't be aware of where it is in the navigation stack—should be the role of the component pushing a `KYCApplicationCompleteController` in the navigation stack.

How about updating `handle(even: KYCEvent)` method in `KYCCoordinator` to:

```
func handle(event: KYCevent) {
    switch event {
        // ...
        case .nextPageFromPageType(let type, let payload):
            handlePayloadFromPageType(type, payload)
            guard let nextPage = type.nextPage(for: self.user) else { return }
            let controller = pageFactory.createFrom(
                pageType: nextPage,
                in: self,
                payload: payload
            )
            controller.navigationItem.hidesBackButton = (nextPage == .applicationComplete)
            navController.pushViewController(controller, animated: true)
}
```

?

PR #480, opened 2018-09-11
feat(Debug): IOS-1263 optionally allow prefilled fields per QA request

woudini:
## Objective

Optionally allow prefilled fields for the Create Wallet View in Debug mode.

## Description

Added Debug menu entry for enabling prefilled fields for the Create Wallet View.

## How to Test

Run app in a debug scheme, go to Create Wallet screen, see that fields are not pre-populated. Go to Debug menu, toggle the Create Wallet Prefill option ON, and see that fields are pre-populated.

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/View Controllers/DebugTableViewController.m
+++ b/Blockchain/View Controllers/DebugTableViewController.m
@@ -22,8 +22,8 @@ typedef NS_ENUM(NSInteger, DebugTableViewRow) {
     RowSecurityReminderTimer,
     RowZeroTickerValue,
     RowKYC,
+    RowCreateWalletPrefill,
     RowTotalCount,
-    RowCreateWalletPrefill
 };
 
 typedef enum {
--- a/Blockchain/View Controllers/DebugTableViewController.m
+++ b/Blockchain/View Controllers/DebugTableViewController.m
@@ -205,6 +205,7 @@ typedef enum {
         }
         case RowKYC: {
             cell.textLabel.text = @"Launch KYC";
+            break;
         }
         case RowCreateWalletPrefill: {
             cell.textLabel.text = @"Create wallet prefill";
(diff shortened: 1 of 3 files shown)


bchrisarriola:
Hm I think this is really useful for us devs though. 

I think we just need to make sure that when we archive a build, and submit to the app store, we make sure that we have `DEBUG = 0`. We can make sure to just make note of that in our release process doc.

Thoughts?

achtenhagen:
Agreed with @bchrisarriola, we should not remove this and instead ensure the debug flag is disabled for production builds (which is how it was before the cocoapods integration). A simple tweak to the .xcconfig files should do the trick. QA should expect the sign up fields to be populated for as long as we serve them builds that use the dev environment.

woudini:
Added as debug menu entry. QA loses time by having to clear the fields manually to enter a KYC-compatible email so it makes sense for them to have this option.

Just a clarification - the debug flag is checking whether `DEBUG` is defined, so `DEBUG=0` will have the same effect as `DEBUG=1`.

Blockchain/View Controllers/DebugTableViewController.m
```diff
@@ -22,7 +22,8 @@ typedef NS_ENUM(NSInteger, DebugTableViewRow) {
     RowSecurityReminderTimer,
     RowZeroTickerValue,
     RowKYC,
-    RowTotalCount
+    RowTotalCount,
+    RowCreateWalletPrefill
```
bchrisarriola: Should go above `RowTotalCount `, otherwise, `RowTotalCount` will evaluate to the wrong value.

woudini:
Can we make `BlockchainDebugSettings` a TODO? Having this merged will be helpful for QA in the short term.

woudini:
Created story and added comment. Thanks!

PR #483, opened 2018-09-11
fix(KYC): take user directly to status screen (IOS-1266)

achtenhagen:
## Objective

Upon KYC completion, take the user back to the status screen instead of showing the _Application Completed_ screen.

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/KYCCoordinator.swift
+++ b/Blockchain/KYC/KYCCoordinator.swift
@@ -85,8 +85,12 @@ protocol KYCCoordinatorDelegate: class {
                 Logger.shared.debug("Got user with ID: \($0.personalDetails?.identifier ?? "")")
                 LoadingViewPresenter.shared.hideBusyView()
                 self.user = $0
-                self.initializeNavigationStack(viewController)
-                self.restoreToMostRecentPageIfNeeded()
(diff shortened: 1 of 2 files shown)


Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -85,8 +85,12 @@ protocol KYCCoordinatorDelegate: class {
                 Logger.shared.debug("Got user with ID: \($0.personalDetails?.identifier ?? "")")
                 LoadingViewPresenter.shared.hideBusyView()
                 self.user = $0
-                self.initializeNavigationStack(viewController)
-                self.restoreToMostRecentPageIfNeeded()
+                if self.pageTypeForUser() == .accountStatus {
+                    self.presentAccountStatusView(for: self.user!.status, in: viewController)
```
bchrisarriola: Just use `$0` (i.e. the user) which is not an optional so you don't need to do force unwrapping

```
self.presentAccountStatusView(for: $0.status, in: viewController)
```

achtenhagen: Will update, thanks!

PR #484, opened 2018-09-11
fix(KYC): only show passport and driver's license options (IOS-1275)

achtenhagen:
## Objective

Only show supported document types during identity verification step (passport and driver's license). Note that other document types are still in the code in case we add support at a later point.

## How to Test

Launch KYC flow and proceed to identity verification step. Verify that the there are only two supported document types.

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/Identity Verification/KYCVerifyIdentityController.swift
+++ b/Blockchain/KYC/Identity Verification/KYCVerifyIdentityController.swift
@@ -60,14 +60,14 @@ final class KYCVerifyIdentityController: KYCBaseViewController {
     // MARK: - Private Methods
 
     private func setUpAndShowDocumentDialog() {
-        let documentDialog = UIAlertController(title: "Which document are you using?", message: nil, preferredStyle: .actionSheet)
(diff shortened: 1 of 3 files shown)


Blockchain/KYC/Identity Verification/KYCVerifyIdentityController.swift
```diff
@@ -67,17 +67,9 @@ final class KYCVerifyIdentityController: KYCBaseViewController {
         let driversLicenseAction = UIAlertAction(title: "Driver's License", style: .default, handler: { _ in
             self.didSelect(.driversLicense)
         })
-        let identityCardAction = UIAlertAction(title: "Identity Card", style: .default, handler: { _ in
```
bchrisarriola: I know this isn't related to your changes, but just realized that some of the strings here aren't using `NSLocalizedString`. Mind fixing as well?

achtenhagen: Sure, will do.

PR #482, opened 2018-09-11
feat(Homebrew-Exchange): configure keypad UI

woudini:
## Objective

`kevin/IOS-1176/fractional_input_updated` (now merged and deleted) and `kevin/IOS-1176/exchange_UI_updated` (this branch) should be merged into dev before falling too far behind. Currently another branch `kevin/IOS-1176/mock_exchange` has many changes that I'd like to open a PR for to merge into `dev`.

Configure keypad UI for Exchange Creation screen. Demo code is deleted in another branch that will be opened shortly.

## Description
Most changes in this PR are in https://github.com/blockchain/My-Wallet-V3-iOS/pull/477

Keypad:
- Added borders
- Rounded corners
- Set fonts

View Controller:
- Set view controller appearance

## How to Test

Add `showExchange(type: .homebrew); return;` to `ExchangeCoordinator`'s `start()`.

## Screenshot/Design assessment

Design is still pretty off in measurement, but closer in attributes. See https://github.com/blockchain/My-Wallet-V3-iOS/pull/477s

## Merge Checklist

- [x] The PR uses a title supported by [.changelogrc](https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/.changelogrc#L6...L69).
- [ ] Areas of technical debt are marked with a `// TICKET:` comment that includes a ticket number.
- [x] All unit tests pass.
- [ ] You have added unit tests.
- [ ] New files added are within the correct directory. (e.g. if a file is required for unit tests to compile, be sure it is added to the tests target.)

PR #485, opened 2018-09-11
feat(Homebrew-Shapeshift): update values from conversion socket message

woudini:
Please merge #482 before this one.

## Objective

Update labels from conversion socket messages.

### User Input

 
- `PrimaryLabel` is the larger label - this shows the current input (what the user is typing). This label's text is assigned to `ExchangeInputs`'s `activeInput.input`.
- `SecondaryLabel` is the smaller label below the `PrimaryLabel` - this also shows the current input, but in crypto or fiat (whatever is opposite to the kind `PrimaryLabel` is displaying) - this label's text is assigned to `ExchangeInputs`'s `lastOutput`.
- The `ExchangeConversionService` is a class that helps with sanitizing data from a `Conversion` struct.
- The `Conversion` socket message is received to the `PublishSubject` and is sanitized by removing the trailing zeros and possibly trailing decimal point (performed by the `ExchangeConversionService`).
- The `SecondaryLabel`'s text is updated with the `output` value.
- The sanitized `input` value is compared to a sanitized copy of the current `activeInput.input` - if the values are the same, then the label is not updated to allow uninterrupted keypad input (so that the user can keep deleting zeros from `50.00`, for example, but the output value is still the latest from the socket).
- This still needs some tweaks, since in some cases user input is still being interrupted.

### TradingPairView 
- `TradingPairView` is (somewhat) hooked up through changing the button titles to reflect crypto values (also provided by the `ExchangeConversionService`).

Especially interested in feedback on the `ExchangeConversionService`.

## How to Test

- Revert 6e49088e24b0012d841c12a39870bdca0fbc4b2b
- Open Exchange
- Start typing in values on the keypad and see that the `SecondaryLabel` updates. Can also tap the right-hand side button to switch between crypto and fiat.

Measurements are still off, but getting closer to design specs. GIF is sped up for easier viewing.
56yqkT

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -12,24 +12,23 @@ import RxSwift
 class ExchangeCreateInteractor {
     var disposable: Disposable?
     weak var output: ExchangeCreateOutput?
-    fileprivate let inputs: ExchangeInputsAPI
+    fileprivate var inputs: ExchangeInputsAPI
```
bchrisarriola: Seems like we'd want this to be immutable. In what scenario does this get changed?

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -12,24 +12,23 @@ import RxSwift
...
     fileprivate var markets: ExchangeMarketsAPI
-    private let inputsState: InputsState
+    fileprivate var conversions: ExchangeConversionAPI
```
bchrisarriola: Same comment about immutability here.

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -103,11 +143,28 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
-        updateInput()
+        updatedInput()
     }
     
     func onAddInputTapped(value: String) {
+        guard let model = model else {
+            Logger.shared.error("Updating conversion with no model")
+            return
+        }
+        if model.isUsingFiat == true {
```
bchrisarriola: Nit: can be simplified to just `if model.isUsingFiat {`

Blockchain/Homebrew/Exchange/Services/ExchangeConversionService.swift
```diff
@@ -0,0 +1,102 @@
...
+    // The amount being typed in by the user (what goes in the primary or larger, more prominent label)
+    var input: String { get }
+
+    // The conversion result amount from the input amount (what goes in the secondary or smaller, less prominent label)
+    var output: String { get }
+
+    // The base output is what should be displayed on the left hand side of the trading pair view.
+    var baseOutput: String { get }
+
+    // The counter fix output is what should be displayed on the right hand side of the trading pair view.
```
bchrisarriola: Thanks for adding these comments. Super useful 👍

Blockchain/Homebrew/Exchange/Services/ExchangeConversionService.swift
```diff
@@ -0,0 +1,102 @@
...
+    // Method used to remove trailing zeros and decimals for true value comparison
+    // Primariy used to allow the user to keep typing uninterrupted
+    func removeInsignificantCharacters(input: String) -> String
+}
+
+class ExchangeConversionService: ExchangeConversionAPI {
+    var input: String = ""
+    var output: String = ""
+    var baseOutput: String = ""
+    var counterOutput: String = ""
```
bchrisarriola: Should these all be scoped as `private(set)` so that these values are only updated internally (i.e. via `update(with:)`)

Blockchain/Models/SocketMessage.swift
```diff
@@ -95,30 +98,31 @@ struct Conversion: SocketMessageCodable {
     typealias JSONType = Conversion
 
     let sequenceNumber: Int
-    let channel, type, pair, fiatCurrency: String
-    let fix: Fix
-    let volume: Double
-    let currencyRatio: CurrencyRatio
+    let channel, type: String
```
bchrisarriola: Defining each property in a new line is much more readable

i.e.

```
let channel: String
let type: String
```

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
woudini: It doesn't, but I remember the Swift compiler complaining when I changed set a property of `inputs` - now it doesn't 😕. Will change

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
woudini: Sorry I think I made changes after you reviewed - this line is now `model?.isUsingFiat == true`

Blockchain/Network/SocketManager.swift
```diff
@@ -134,9 +136,10 @@ extension SocketManager: WebSocketAdvancedDelegate {
 
         // Optimization: avoid retyping "tryToDecode(data: data, onSuccess: onSuccess, onError: onError)" for each case
         switch type {
-        case "conversion": Conversion.tryToDecode(data: data, onSuccess: onSuccess, onError: onError)
-        case "heartbeat", "subscribed", "authenticated": HeartBeat.tryToDecode(data: data, onSuccess: onSuccess, onError: onError)
-        case "error": onError("Error returned")
+        case "currencyRatio": Conversion.tryToDecode(socketType: socketType, data: data, onSuccess: onSuccess, onError: onError)
+        case "currencyRatioError": onError("Currency ratio error: \(json)")
+        case "heartbeat", "subscribed", "authenticated": HeartBeat.tryToDecode(socketType: socketType, data: data, onSuccess: onSuccess, onError: onError)
```
bchrisarriola: Actually, getting a swiftlint error here (need a newline)

PR #497, opened 2018-09-14
feat(Exchange): IOS-1243 Exchange picker screen

bchrisarriola:
## Objective

Allowing user to select account to exchange into/out from. This action sheet can be triggered by tapping on the "Exchange" or "Receive" button in the exchange create view.

## Description

**Summary:**
* Created a new struct `AssetAccount` which encapsulates a Blockchain account (this way, we no longer need to pass indices around)
* Created `AssetAccountRepository` which is in charge of fetching `AssetAccount` objects.

**Looking for input:**
* Not sure where the value for "fix" should be stored/coming from (see TODO statement). I'm always passing in `.base` when invoking `updateTradingPair`
* The type for `AssetAccount.balance` is CLongLong. However, looks like the ethereum balance is stored in ether(?) and not the smallest unit, wei so CLongLong won't work. I think this value should be converted to wei in `AssetAccount.balance` so that we are always storing the smallest unit for all assets. Wanted to get input first before doing that (at the moment I just set the balance for Eth to 0) 
* Need to also get legacy addresses for BTC and BCH

## How to Test

Go to exchange view, tap "Exchange"/"Receive" button and see picker

img_0045

## Merge Checklist

- [X] All unit tests pass.


Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
```diff
@@ -63,47 +66,79 @@ class ExchangeCreateViewController: UIViewController {
...
+
+    private func onExchangeAccountChanged() {
+        guard let tradingPair = TradingPair(
+            from: fromAccount.address.assetType,
+            to: toAccount.address.assetType
+        ) else {
+            return
+        }
+        // TODO: where should the value of `fix` come from?
+        presenter.updateTradingPair(pair: tradingPair, fix: .base)
```
woudini: Normally gestures go from the controller to the presenter (via `ExchangeCreateDelegate`) to the interactor (via `ExchangeCreateInput`) where the `MarketsModel` (which should probably default to `.base` on initialization) is accessible, then back to the presenter (via `ExchangeCreateOutput`) and then to the controller (via `ExchangeCreateInterface`). But that's kind of a lot to do.

woudini:
Re the second point, the app uses `NSDecimalNumber` for ether values since the backend typically gives values in ether units (usually as strings). A couple of issues I foresee using wei:
- Could be a bit confusing as values smaller than 0.00000001 ETH (past eight decimal places) would basically be treated as 0 since we use 8 decimal places for crypto.
- It also would introduce some conversion overhead in the short term since sending ether is done currently in ether units with 8 decimal places.
- 18 decimal places is also quite a lot to handle, so representing 1 ETH in wei could open the door for readability and overflow issues.

So although it does make sense to be consistent across crypto types and use wei, there would be friction created by how the backend and current state of the app function.

Blockchain/Accounts/AssetAccount.swift
```diff
@@ -0,0 +1,25 @@
...
+struct AssetAccount {
+
+    /// The index of this account in the wallet metadata (always 0 for ether)
+    let index: Int32
+
+    /// The AssetAddress for this account
+    let address: AssetAddress
+
+    /// The balance in this account
+    let balance: CLongLong
```
woudini: Should this be `CUnsignedLongLong` since the equivalent is `uint64_t`?

bchrisarriola: Given the discussion on ether though, which is a decimal, I think this should be changed to `Decimal`. Wdyt?

woudini: Ah right. Yeah let's use `Decimal`.

Blockchain/Accounts/AssetAccountRepository.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    private let wallet: Wallet
+
+    init(wallet: Wallet = WalletManager.shared.wallet) {
+        self.wallet = wallet
+    }
+
+    // MARK: Public Methods
+
+    func accounts(for type: AssetType) -> [AssetAccount] {
```
woudini: Thoughts on renaming `type` to `assetType` like in line 55?

Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
woudini: I'll do this in another PR. The trading pair update needs to reach the interactor anyway to update the socket.

bchrisarriola: K. So I'll leave this as is?

Blockchain/Accounts/AssetAccount.swift
bchrisarriola: Will change

Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
```diff
@@ -63,47 +66,79 @@ class ExchangeCreateViewController: UIViewController {
     // MARK: Lifecycle
 
     override func viewDidLoad() {
+        super.viewDidLoad()
         dependenciesSetup()
+        viewsSetup()
         delegate?.onViewLoaded()
-        
+        if let navController = navigationController as? BCNavigationController {
```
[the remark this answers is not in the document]
bchrisarriola: Yeah. I this `viewWillAppear` is a better place for it.

Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
woudini: Yeah

Blockchain/Accounts/AssetAccountRepository.swift
```diff
@@ -84,14 +87,15 @@ class AssetAccountRepository {
...
     static func create(assetType: AssetType, index: Int32, wallet: Wallet) -> AssetAccount? {
-        guard let address = wallet.getReceiveAddress(forAccount: index, assetType: type.legacy) else {
+        guard let address = wallet.getReceiveAddress(forAccount: index, assetType: assetType.legacy) else {
             return nil
         }
-        let name = wallet.getLabelForAccount(index, assetType: type.legacy)
-        let balance = wallet.getBalanceForAccount(index, assetType: type.legacy) as? CLongLong ?? 0
+        let name = wallet.getLabelForAccount(index, assetType: assetType.legacy)
+        let balanceLong = wallet.getBalanceForAccount(index, assetType: assetType.legacy) as? CUnsignedLongLong ?? 0
+        let balance = Decimal(balanceLong)
```
woudini: I think this would be better in BTC units instead of satoshis (using `NSNumberFormatter+Currencies`'s `+ (NSString *)satoshiToBTC:(uint64_t)value`)

I can do this in another PR though if going from `uint64_t` to `CUnsignedLongLong` presents issues

PR #501, opened 2018-09-17
feat(Exchange): IOS-1195 swap currencies

bchrisarriola:
## Objective

Ability to swap currencies in the exchange create screen

## Description

Functionality was already there but you couldn't see the swap button because it was white on a white background (the tint wasn't being applied). This PR fixes that and now shows the swap button with the  tint color provided in `.swapTintColor` (i.e. the blockchain blue-ish gray color) which matches the color in the design spec.

## How to Test

Go to exchange create view

## Merge Checklist

- [X] All unit tests pass.


bchrisarriola:
> Might be a good idea to save the overhead of another PR and decrease the ImageEdgeInsets of that button in this PR - the image is small compared to the design.

K, let me tweak that

PR #505, opened 2018-09-18
feat(Exchange): IOS-1195 swap fix.

bchrisarriola:
## Objective

Fixing behavior of tapping on the swap button in `TradingPairView` to toggle the `Fix` value, and not the trading pairs.

## Description

**Primary Changes:**
* Added the appropriate methods in `MarketsModel`, `ExchangeCreatePresenter` and `ExchangeCreateInteractor` to toggle the Fix.

**Related Changes:**
* refactored the way the trading pairs are selected and refreshed in the views so that it propagates from the view controller, to the presenter, to the interactor, and back.

**Other notes:**
I noticed a few bugs and architecture issues that I didn't address to avoid blowing up the scope but wanted to call out. Namely:

(1) The didSet property observer method in `ExchangeCreateInteractor` for the property `model` might not be working as intended. Specifically, this block of code:

```
private(set) var model: MarketsModel? {
        didSet {
            didSetModel(oldModel: oldValue)
        }
    }
```

The didSet method will only be invoked _if the property itself was changed_. So, if somewhere within `ExchangeCreateInteractor` a method did something like:

```
func foo() {
    model = MarketsModel()
}
```

...only then would the `didSetModel` be invoked. However, if you changed a property within model, such as:

```
func foo() {
    model.property = Property1()
}
```

...`didSet` won't be invoked. 

But there is an exception: If `MarketsModel` was a struct then `didSet` would actually be invoked. The reason this exception happens is because structs are _value types_, which is different from classes which are _referenced types_. This means that changing a property within a struct will actually create a new struct, whereas with classes, changing a property won't create a whole new class, only the property is changed.

(2) The `ExchangeCreateInteractor` has a few oddities which I think break the interactor pattern and make using it a bit confusing. The interactor is supposed to be independent of any UI. All it should do is provide methods that perform CRUD operations to the underlying entities/models. At the moment, this isn't the case and has some verbs like "show", "tapped", it also uses `AlertViewPresenter`, etc. Instead, it seems like it's also taking the role of the presenter which is ultimately the component in charge of taking actions like that. This isn't really a big issue but thought I'd bring up to help think about this pattern (need a reminder for myself, too).

## How to Test

Launch exchange and toggle.

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
+++ b/Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
@@ -18,10 +18,10 @@ class ExchangeCreateViewController: UIViewController {
     
     // MARK: Private Static Properties
     
-    static let primaryFontName: String = Constants.FontNames.montserratRegular
+    static let primaryFontName: String = Constants.FontNames.montserratMedium
     static let primaryFontSize: CGFloat = Constants.FontSizes.Huge
-    static let secondaryFontName: String = Constants.FontNames.montserratRegular
-    static let secondaryFontSize: CGFloat = Constants.FontSizes.SmallMedium
+    static let secondaryFontName: String = Constants.FontNames.montserratMedium
+    static let secondaryFontSize: CGFloat = Constants.FontSizes.MediumLarge
 
     // MARK: - IBOutlets
 
@@ -109,9 +109,6 @@ class ExchangeCreateViewController: UIViewController {
 
         tradingPairView.delegate = self
         exchangeButton.layer.cornerRadius = Constants.Measurements.buttonCornerRadius
-
-        setAmountLabelFont(label: primaryAmountLabel, size: Constants.FontSizes.Huge)
-        setAmountLabelFont(label: secondaryAmountLabel, size: Constants.FontSizes.MediumLarge)
     }
 
     fileprivate func dependenciesSetup() {
@@ -147,10 +144,6 @@ extension ExchangeCreateViewController {
         viewToEdit.layer.cornerRadius = 4.0
         viewToEdit.layer.borderWidth = 1.0
         viewToEdit.layer.borderColor = UIColor.brandPrimary.cgColor
-    }
-
-    private func setAmountLabelFont(label: UILabel, size: CGFloat) {
-        label.font = UIFont(name: Constants.FontNames.montserratRegular, size: size)
     }
 }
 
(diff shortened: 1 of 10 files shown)


Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
```diff
@@ -206,7 +184,7 @@ extension ExchangeCreateViewController: ExchangeCreateInterface {
         let secondary = UIFont(
             name: ExchangeCreateViewController.secondaryFontName,
             size: ExchangeCreateViewController.secondaryFontSize
-            ) ?? UIFont.systemFont(ofSize: 17.0)
+        ) ?? UIFont.systemFont(ofSize: 17.0)
```
woudini: This and `primary` should probably be Montserrat as in the design, and then we can remove https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift#L116-L117

woudini:
>This means that changing a property within a struct will actually create a new struct, whereas with classes, changing a property won't create a whole new class, only the property is changed.

Thanks for explaning. Sounds like we want to keep `MarketsModel` as a class?

woudini:
When I swap with the `BTC-ETH` trading pair from `Exchange` to `Receive`, the `BTC` symbol in the primary input label persists - seems like the issue is in this line https://github.com/blockchain/My-Wallet-V3-iOS/blob/77619617381aa6bc819aed090f4a2d15a7595878/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift#L146
So it should probably be
```
let assetType = model.isUsingBase ? model.pair.from : model.pair.to
let symbol = assetType.symbol
```
And add a method `isUsingBase` to `MarketsModel`:
```
var isUsingBase: Bool {
    return fix == .base || fix == .baseInFiat
}

Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
bchrisarriola: Not really part of this PR but I can change

bchrisarriola:
> Thanks for explaning. Sounds like we want to keep MarketsModel as a class?

Yeah, I think that's fine. But if we do what `didSetModel` to be invoked after mutating a property in `MarketsModel`, we'll need to invoke that method directly.

> When I swap with the BTC-ETH trading pair from Exchange to Receive, the BTC symbol in the primary input label persists - seems like the issue is in this line

Will look into

PR #508, opened 2018-09-18
fix(Homebrew-Exchange): IOS-1305 label and increase size of fiat button

woudini:
Please merge #507 first

## Objective

Expand the tappable area of fiat button.

## Description

Adjusted image insets to give the appearance of an image with a Y position centered with the secondary amount label.

This is cleaner than having a `UIImageView` inside a tappable `UIView`, but still needs maintenance if the height of the `secondaryAmountLabel` changes (in which case, the `centerY` constraint offset will need to be changed).

Interestingly, the button's constraints change after pausing and unpausing with the view debugger. Not sure if this is really a bug since a user theoretically would not be able to execute an action like this.

## How to Test

Open exchange, tap fiat button - should be easier to tap.

View Debugger:
screen shot 2018-09-18 at 2 39 00 pm

6 plus:
toggle6plus

6: 
toggle6

5s: 
toggle5s

## Merge Checklist

- [x] All unit tests pass.


woudini:
VPN not working at the moment, will have to test this tomorrow before merging to ensure the  storyboard conflicts were resolved correctly.

PR #510, opened 2018-09-18
fix(Settings): correct typo in settings view (IOS-1316)

achtenhagen:
## Objective

Rename Recovery Phase -> Recovery Phrase.

## Merge Checklist

- [ ] All unit tests pass.


bchrisarriola:
We have a backlog task to move strings away from storyboard to `LocalizationConstants` because of how our localization process works. Might as well move this string and set it programmatically?

achtenhagen:
@bchrisarriola Correct, but as that is a different task, I'd like to get this merged and then begin moving the strings away from storyboard files in a separate pull request.

PR #513, opened 2018-09-19
feat(HB): Fetching Homebrew trades with new HB model type.

thisisalexmcgregor:
## Objective

The HB models for a trade have been changed. This PR fixes our model. We now have `ExchangeTradeModel`. 

## Description

`ExchangeTradeModel` has two types, `.partner(ExchangeTrade)` and `.homebrew(ExchangeTradeCellModel)`. More than likely, this is only temporary as there will eventually be only a homebrew type. 

I updated the `HomebrewExchangeService` to now use `RxSwift`. Previously it wasn't and it also wasn't including the token in the request. 

## How to Test

You can't for now. This is marked DNM until the endpoint returns models that reflect what is documented. 

N/A

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Homebrew/Exchange/Services/ExchangeService.swift
```diff
@@ -88,20 +88,24 @@ extension ExchangeService: ExchangeHistoryAPI {
         
         partnerOperation = AsyncBlockOperation(executionBlock: { [weak self] complete in
```
bchrisarriola: Should partnerOperation be cancelled as well (similar to how homebrewOperation is cancelled if it's not nil?

bchrisarriola:
> Opted to check and see if the operations are in progress (still in the queue) rather than cancel them. Added another function for cancelling all operations but, can't think of a time when we necessarily need to do that. Maybe on cleanup/deinit.

Makes sense to me to add canceling to `deinit` + maybe also when you pull to refresh?

thisisalexmcgregor:
@bchrisarriola Yeah, good idea. Will include it in the `Exchange Locked` PR

PR #514, opened 2018-09-20
fix(NetworkRequest): error if bad response or status code

woudini:
## Objective

- Handle errors in a way similar to `KYCNetworkRequest`.
- Make `OrderResult` (the response from the trades endpoint) non-optional.
- Update `OrderResult` properties to match server response.

## Description

- Added error handling to `NetworkRequest`'s `execute` method.
- Removed optionals and error-related properties from `OrderResult`.
- Removed force unwraps of `OrderResult` properties from `TradeExecutionService`.

## How to Test

Hit a bad endpoint and see an error instead of an attempt to decode to a certain type.

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Network/Models/Result.swift
+++ b/Blockchain/Network/Models/Result.swift
@@ -16,3 +16,8 @@ enum Result<A> {
     case success(A)
     case error(Error?)
 }
+
+enum HTTPRequestResult<A> {
+    case success(A)
+    case error(HTTPRequestError?)
+}
--- a/Blockchain/Network/Models/Result.swift
+++ b/Blockchain/Network/Models/Result.swift
@@ -16,8 +16,3 @@ enum Result<A> {
     case success(A)
     case error(Error?)
 }
-
-enum HTTPRequestResult<A> {
-    case success(A)
-    case error(HTTPRequestError?)
-}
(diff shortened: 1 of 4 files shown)


Blockchain/Network/Models/Result.swift
```diff
@@ -16,3 +16,8 @@ enum Result<A> {
     case success(A)
     case error(Error?)
 }
+
+enum HTTPRequestResult<A> {
```
bchrisarriola: Not sure I see the advantages of this. Why not just use `Result` and if the case is `error`, check if the value is of type `HTTPRequestError`?

woudini: Removed

Blockchain/Network/NetworkRequest.swift
```diff
@@ -77,25 +77,33 @@ struct NetworkRequest {
...
+                withCompletion(.error(HTTPRequestClientError.failedRequest(description: error.localizedDescription)), responseCode)
+                return
+            }
+
+            guard let httpResponse = response as? HTTPURLResponse else {
+                withCompletion(.error(HTTPRequestServerError.badResponse), responseCode)
+                return
+            }
+
+            guard (200...299).contains(httpResponse.statusCode) else {
```
thisisalexmcgregor: `var resonseCode: Int = 0` on line 68 can be removed

woudini: Can I change to just `let`? Otherwise I'll be passing in `0` a few times in this function.

thisisalexmcgregor: Oh yeah, that's what I meant. Whoops.

PR #515, opened 2018-09-20
fix(Exchange): unsubscribe from unused currency pair.

bchrisarriola:
## Objective

Unsubscribe to rates for currency pairs that are different from the pair the user is trying to exchange in.

## Description

Prevents chatty/extraneous websocket messages where conversion rates are still being received from previous currency pairs. This can happen if the user was previously trying to exchange in a currency pair (e.g. "BTC-ETH") but then changed to a different pair (e.g. "BCH-ETH"). 

## How to Test

Pull in changes, go to create exchange flow, change currency pairs, and verify that you see the logged message: "Pair <old pair> is different from model pair <new pair>"

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
+++ b/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
@@ -95,12 +95,8 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
 
             guard model.pair.stringRepresentation == conversion.quote.pair else {
                 Logger.shared.warning(
-                    """
-                    Pair '\(conversion.quote.pair)' is different from model pair '\(model.pair.stringRepresentation)'.
-                    Unsubscribing from pair '\(conversion.quote.pair)'
-                    """
+                    "Pair '\(conversion.quote.pair)' is different from model pair '\(model.pair.stringRepresentation)'."
                 )
-                this.markets.unsubscribeToCurrencyPair(pair: conversion.quote.pair)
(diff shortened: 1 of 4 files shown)


Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -90,8 +90,17 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
     func subscribeToConversions() {
         disposable = markets.conversions.subscribe(onNext: { [weak self] conversion in
             guard let this = self else { return }
-            guard let model = this.model, model.pair.stringRepresentation == conversion.quote.pair else {
-                Logger.shared.error("Pair returned from conversion is different from model pair")
+
+            guard let model = this.model else { return }
+
+            guard model.pair.stringRepresentation == conversion.quote.pair else {
```
woudini: I think the overall client logic will be cleaner/more intuitive if it rejected conversions whose trading pair didn't match the `model`'s and instead unsubscribed when `model.pair` changes. Thoughts?

bchrisarriola: Yeah that's probably a more proactive strategy. Will update.

woudini:
Currently not receiving conversion socket messages even though the app is subscribing and receiving other socket messages.

bchrisarriola:
Looks like conversions are coming through dev though. Will test on that env.

PR #516, opened 2018-09-20
feat(HB): Hooking up Exchange Locked Screen - IOS-1210

thisisalexmcgregor:
## Objective

Hooks up the `Exchange Locked` screen and handles the empty state of the `Exchange List` screen.

## Description

The `Exchange List` screen (Order History) will now show a loading indicator initially. The `Create New Order` button is always visible regardless whether or not trades have been placed. The user must tap this button in order to create a new trade. 

After submitting a trade, we pop back to the rootVC and then present the `Exchange Locked` screen modally. 

## How to Test

1. Build an run
2. Copy your BTC address
3. Enter into the settings debug menu
4. Go to the `Exchange` flow
5. Observer the `Exchange List` screen
6. Create a new exchange
7. Submit the exchange
8. View the `Exchange Locked` screen
9. Tap `Done` and you should be back at the `Exchange List` screen.

See ios-dev channel for demo. 

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Homebrew/Exchange/ExchangeDetailCoordinator.swift
```diff
@@ -80,9 +88,13 @@ class ExchangeDetailCoordinator: NSObject {
                     bold: true
                 )
                 
+                let accounts = accountRepository.allAccounts()
+                
+                guard let destination = accounts.filter({ return $0.address.address == orderTransaction.to.address }).first else { return }
```
woudini: Seeing this `return` for all cells so I get an empty confirm screen - I think the `OrderResult`'s `destinationAddress` should be used here instead of `orderTransaction.to.address`.

I think `OrderTransaction` might need another property like `destinationAddress` - right now the `from` and `to` are both in terms of the `base` assetType since I originally meant for it to describe mostly the details of the payment built in the JS. It's a bit confusing because "to" implies the same `AssetType` as the "from" in the context of a normal transaction but implies a different `AssetType` in the context of a `TradingPair`. So one way to fix this would be to assign the `OrderResult`'s `withdrawalAddress` as the value of a new property on `OrderTransaction` (e.g. `destinationAddress` or `withdrawalAddress`) that indicates where the user will receive the `counter` amount. Ping me if anything is unclear.

thisisalexmcgregor: ❓ 
lol will ping. 

Yeah got a blank screen there as well, but only after I had one successful trade

thisisalexmcgregor: @woudini Added `// TICKET:` and a fix to prevent early return: be456ac

woudini: Early return was removed from `.confirm` but should be removed from `.locked` also

thisisalexmcgregor: Fixed 
9b70e42

PR #517, opened 2018-09-20
feat(Exchange): IOS-1322 Checking balances when computing min/max

bchrisarriola:
## Objective

Checking account balances when populating min/max value that the user wishes to trade. Please review merge #515 before reviewing/merging this.

## Description

Previously, tapping on min/max will populate the trading limits for the user as provided get `GET /trades/limits`. However, this didn't take into account the balance of the user's account. In this PR, we are now computing the balance in fiat of the crypto the user wishes to trade.

This is done by:
* Getting the best exchange rates from the "exchange_rate" socket channel
* Taking the `AssetAccount` the user is exchanging from to retrieve the balance, followed by using the appropriate exchange rate to compute the balance in fiat.

Note that I exposed Rx methods in `TradeLimitsAPI` and `ExchangeMarketsAPI`. The reason I decided to Rx-ify here is two-fold:
(1) computing the min/max now needs to wait for 2 asynchronous calls (i.e. trading limits and best exchange rates
(2) exchange rates change dynamically, and so with an Rx approach, we can get the user's account balance in fiat as a reactive stream and re-compute it as the rates change

## How to Test

Pull in changes, go to exchange create view, tap min/max.

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/Accounts/AssetAccountRepository.swift
+++ b/Blockchain/Accounts/AssetAccountRepository.swift
@@ -60,9 +60,7 @@ class AssetAccountRepository {
         return AssetAccount.create(assetType: assetType, index: index, wallet: wallet)
     }
 
-    // MARK: Private Methods
-
-    private func defaultEthereumAccount() -> AssetAccount? {
+    func defaultEthereumAccount() -> AssetAccount? {
         guard let ethereumAddress = wallet.getEtherAddress(), wallet.hasEthAccount() else {
             Logger.shared.debug("This wallet has no ethereum address.")
             return nil
@@ -86,6 +84,8 @@ class AssetAccountRepository {
 
 extension AssetAccount {
 
+    /// Creates a new AssetAccount. This method only supports creating an AssetAccount for
+    /// BTC or BCH. For ETH, use `defaultEthereumAccount`.
     static func create(assetType: AssetType, index: Int32, wallet: Wallet) -> AssetAccount? {
         guard let address = wallet.getReceiveAddress(forAccount: index, assetType: assetType.legacy) else {
             return nil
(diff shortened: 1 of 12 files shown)


Blockchain/Network/Models/SocketMessage.swift
```diff
@@ -160,6 +184,11 @@ extension Conversion {
 
 // MARK: - Associated Models
 
+struct CurrencyPairRate: Codable {
+    let pair: String
```
woudini: Possible to use `TradingPair` here?

bchrisarriola: Hmm that would be ideal but currently `TradingPair` contains `AssetType` (i.e. only crypto). However, `pair` here can be crypto to fiat (e.g. "BTC-USD")

woudini: Ah right, gotcha.

Blockchain/Accounts/AssetAccountRepository.swift
```diff
@@ -93,7 +92,12 @@ extension AssetAccount {
         }
         let name = wallet.getLabelForAccount(index, assetType: assetType.legacy)
         let balanceLong = wallet.getBalanceForAccount(index, assetType: assetType.legacy) as? CUnsignedLongLong ?? 0
-        let balance = Decimal(balanceLong)
+        let balance: Decimal
+        if assetType == .bitcoin || assetType == .bitcoinCash {
+            balance = Decimal(balanceLong) / Decimal(Constants.Conversions.satoshi)
+        } else {
+            balance = Decimal(balanceLong)
```
woudini: This line doesn't appear to be called but would be critically wrong if it were - line 94 would truncate all ETH values to an integer. I do realize this is a bit older code though so I can fix in another PR.

bchrisarriola: Let me add a comment here that this isn't actually called if the `assetType == .ethereum` (i.e. this method `create` is not invoked). If you notice, `defaultEthereumAccount` is called is the asset type is ethereum.

Let me add a comment to clarify that.

woudini: I still think these lines should be changed but let's get this PR in.

PR #518, opened 2018-09-21
fix(AssetAccountRepository): cast balance to string for ethereum and …

woudini:
…return empty array if wallet is not initialized

## Objective

Fix bugs in `AssetAccountRepository`

## Description

Fixes two bugs
- One crash when the wallet is not initialized
- One bug where the ether balance is casted to an integer but it should be casted to a string

## How to Test

Test making an exchange and sending it (only flow where `AssetAccountRepository` is used

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Accounts/AssetAccountRepository.swift
+++ b/Blockchain/Accounts/AssetAccountRepository.swift
@@ -23,6 +23,12 @@ class AssetAccountRepository {
 
     func accounts(for assetType: AssetType) -> [AssetAccount] {
 
+        // A crash occurs in the for loop if wallet.getActiveAccountsCount returns 0
+        // "Fatal error: Can't form Range with upperBound < lowerBound"
+        if !wallet.isInitialized() {
+            return []
+        }
+
         // Handle ethereum
         if assetType == .ethereum {
             if let ethereumAccount = defaultEthereumAccount() {
@@ -91,12 +97,14 @@ extension AssetAccount {
             return nil
         }
         let name = wallet.getLabelForAccount(index, assetType: assetType.legacy)
-        let balanceLong = wallet.getBalanceForAccount(index, assetType: assetType.legacy) as? CUnsignedLongLong ?? 0
+        let balanceFromWalletObject = wallet.getBalanceForAccount(index, assetType: assetType.legacy)
         let balance: Decimal
         if assetType == .bitcoin || assetType == .bitcoinCash {
+            let balanceLong = balanceFromWalletObject as? CUnsignedLongLong ?? 0
             balance = Decimal(balanceLong) / Decimal(Constants.Conversions.satoshi)
         } else {
-            balance = Decimal(balanceLong)
+            let balanceString = balanceFromWalletObject as? String ?? "0"
+            balance = NSDecimalNumber(string: balanceString).decimalValue
         }
         return AssetAccount(
             index: index,


Blockchain/Accounts/AssetAccountRepository.swift
```diff
@@ -23,6 +23,12 @@ class AssetAccountRepository {
 
     func accounts(for assetType: AssetType) -> [AssetAccount] {
 
+        // A crash occurs in the for loop if wallet.getActiveAccountsCount returns 0
+        // "Fatal error: Can't form Range with upperBound < lowerBound"
```
achtenhagen: The reason is because `getActiveAccountsCount` will return 0 if there are no active accounts and your for loop should not alter the index as you are not looking up items by their indexes. Removing the `-1` from:
```
for index in 0...wallet.getActiveAccountsCount(assetType.legacy)-1
...
```
should fix the issue.

achtenhagen: > upperBound < lowerBound

Your upperBound is therefore `-1` and is less than your lowerBound which is `0`.

woudini: Wouldn't `wallet.getActiveAccountsCount(assetType.legacy)` be an index out of bounds? The last index of a collection should be the collection's `count - 1` right?

woudini: Seems like `0..<wallet.getActiveAccountsCount(assetType.legacy)` is a better way?

achtenhagen: From what I can see in:
```
- (int)getActiveAccountsCount:(LegacyAssetType)assetType
{
    if (![self isInitialized]) {
        return 0;
    }

    if (assetType == LegacyAssetTypeBitcoin) {
        return [[[self.context evaluateScript:@"MyWalletPhone.getActiveAccountsCount()"] toNumber] intValue];
    } else if (assetType == LegacyAssetTypeBitcoinCash) {
        return [[[self.context evaluateScript:@"MyWalletPhone.bch.getActiveAccountsCount()"] toNumber] intValue];
    }
    return 0;

}
```
It will only return the number of accounts or zero. I don't think this would ever result in an out of bounds exception since your `for` loop would say: `from zero to zero, do...`, which is the same as `do nothing`.

achtenhagen: Since you are checking for initialization in `getActiveAccountsCount`, I think you should be good to remove it from the `accounts` function on line 28.

woudini: Correct it wouldn't be an exception, just an invalid index.
>Since you are checking for initialization in getActiveAccountsCount, I think you should be good to remove it from the accounts function on line 28.

Removing this check would still allow creating an `AssetAccount` at index 0 which is a false representation of the wallet's accounts - if the wallet is initialized it isn't decrypted, so anyone wanting to view the asset accounts should just get an empty array.

Feel free to open a PR about using `0..<wallet.getActiveAccountsCount(assetType.legacy)`, but I'd like to keep the initialization check + return at the beginning.

achtenhagen: Oh, so you mean if the wallet is initialized, the index will always be greater than zero?

woudini: Not sure where you're picking that up from, can you elaborate?

achtenhagen: The part I'm hung up on is
> at index 0 which is a false representation of the wallet's accounts
Why is creating an account at index 0 a false representation?

woudini: Try not to get hung up on which index is being used. Think of the bigger picture. Why would an API user want to call this method?
Imagine you're working with a bank API. You use the API to try to get the balance of an account that you know has a balance of 50, and it returns to you a balance of 0. You later find out that the reason it returned a balance of 0 instead of 50 is because you haven't authenticated with your user credentials.
It's deceptive to return an `AssetAccount` if the wallet isn't decrypted because it more closely mimics the behavior of a successful call. It's more truthful to return nothing because if the wallet is still encrypted you aren't supposed to know anything about it.

achtenhagen: Got it, thanks for the follow up explanation 👍

PR #520, opened 2018-09-21
feat(Homebrew-Exchange): use fiat symbol from wallet settings

woudini:
## Objective

Use correct fiat symbols in the exchange flow.

## Description

- Used fiat symbol from wallet object through `BlockchainSettings` instead of USD.
- Used the `fiatCurrency` stored in `MarketsModel` instead of `NumberFormatter`'s `currencySymbol`
- Used `fiat.value` from `conversion.quote.currencyRatio`'s `base` or `counter` for display
- Replaced the ambiguous `fiatCurrency` property on `MarketsModel` with `fiatCurrencySymbol` ($, £) and `fiatCurrencyCode` (USD, GBP).

## How to Test

- Go to Exchange with a few different Local Currency options set in Settings. See the currency symbol displayed in the input labels.
- Go to the confirm screen and see that the appropriate currency symbol is displayed.

Not able to test at the moment due to server issues.
Brazil Real as an example:
screen shot 2018-09-21 at 11 47 25 am

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/ExchangeDetailCoordinator.swift
+++ b/Blockchain/Homebrew/Exchange/ExchangeDetailCoordinator.swift
@@ -70,7 +70,7 @@ class ExchangeDetailCoordinator: NSObject {
                     model: TradingPairView.confirmationModel(for: conversion)
                 )
                 
-                let symbol = NumberFormatter.localCurrencyFormatter.currencySymbol ?? ""
+                let symbol = conversion.quote.currencyRatio.counter.fiat.symbol
                 
                 let value = ExchangeCellModel.Plain(
                     description: LocalizationConstants.Exchange.value,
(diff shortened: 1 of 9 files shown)


Blockchain/Homebrew/Exchange/ExchangeDetailCoordinator.swift
```diff
@@ -138,7 +138,7 @@ class ExchangeDetailCoordinator: NSObject {
                     model: TradingPairView.confirmationModel(for: conversion)
                 )
                 
-                let symbol = NumberFormatter.localCurrencyFormatter.currencySymbol ?? ""
+                let symbol = conversion.quote.currencyRatio.counter.fiat.symbol
```
thisisalexmcgregor: We can revisit this later but we'll wanna probably add some brief descriptions about the different models/variables on `Conversion` and maybe some convenience extensions on `conversion` to get the correct symbol. [See Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter)

woudini: Agree, accessing properties this deep into the struct is a code smell.

PR #522, opened 2018-09-21
fix(HB): Exchange List Shown Always

thisisalexmcgregor:
## Objective

The user's exchange history should be shown always, even if there aren't any trades.

## Description

Fixes the issue where tapping back showed a white screen on the `Create Exchange` view.

## How to Test

1. Build and run
2. Go to the exchange with no trades in your history.
3. It should show an empty exchange history screen. 
4. Tap New Exchange
5. Tap the back button
6. You should return to the empty exchange history screen. 

simulator screen shot - iphone x - 2018-09-21 at 13 13 49

**NOTE: We are still waiting on copy for an empty history**

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Presenters/ExchangeListPresenter.swift
+++ b/Blockchain/Homebrew/Exchange/Presenters/ExchangeListPresenter.swift
@@ -59,12 +59,7 @@ extension ExchangeListPresenter: ExchangeListOutput {
     
     func loadedTrades(_ trades: [ExchangeTradeModel]) {
         interface?.refreshControlVisibility(.hidden)
-        if trades.count == 0 {
-            interface?.showNewExchange(animated: false)
-        } else {
-            interface?.enablePullToRefresh()
-            interface?.display(results: trades)
-        }
+        interface?.display(results: trades)
     }
     
     func appendTrades(_ trades: [ExchangeTradeModel]) {


Blockchain/Homebrew/Exchange/Presenters/ExchangeListPresenter.swift
```diff
@@ -59,12 +59,7 @@ extension ExchangeListPresenter: ExchangeListOutput {
     
     func loadedTrades(_ trades: [ExchangeTradeModel]) {
         interface?.refreshControlVisibility(.hidden)
-        if trades.count == 0 {
-            interface?.showNewExchange(animated: false)
-        } else {
-            interface?.enablePullToRefresh()
-            interface?.display(results: trades)
-        }
+        interface?.display(results: trades)
```
woudini: `interface?.enablePullToRefresh()`?

thisisalexmcgregor: Right, wasn't sure initially if we wanted users to `pullToRefresh` when we first load the screen. However, we have safe guards to prevent multiple requests from firing, and we use that `UIRefreshControl` to show that we are fetching their latest trades when the screen first loads so we call: `interface?.enablePullToRefresh()` in `onLoaded()`

PR #523, opened 2018-09-21
feat(HB): IOS-1312 - All currencies should show when selecting source

thisisalexmcgregor:
## Objective

All currency options should show when selecting a trade option. 

## Description

We're no longer filtering against the selected asset. Also, if they select the asset that matches the other asset in the pair, we swap the two assets. 

## How to Test

1. Build and run
2. Go to the exchange
3. Tap either asset
4. All asset types should be shown

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Accounts/AssetAccount.swift
+++ b/Blockchain/Accounts/AssetAccount.swift
@@ -23,3 +23,12 @@ struct AssetAccount {
     /// The name of this account
     let name: String
 }
+
+extension AssetAccount: Equatable {
+    static func == (lhs: AssetAccount, rhs: AssetAccount) -> Bool {
+        return lhs.index == rhs.index &&
+        lhs.address.address == rhs.address.address &&
+        lhs.balance == rhs.balance &&
+        lhs.name == rhs.name
+    }
+}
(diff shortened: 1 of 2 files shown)


Blockchain/Accounts/AssetAccount.swift
```diff
@@ -23,3 +23,12 @@ struct AssetAccount {
     /// The name of this account
     let name: String
 }
+
+extension AssetAccount: Equatable {
+    static func == (lhs: AssetAccount, rhs: AssetAccount) -> Bool {
+        return lhs.index == rhs.index &&
+        lhs.address.address == rhs.address.address &&
```
woudini: How does this work? When I compare two `AssetAddresses` I get `error: binary operator '==' cannot be applied to two 'AssetAddress' operands` since there's no `Equatable` for it?

thisisalexmcgregor: 🤔  I'll double check but `address.address` is a string value

thisisalexmcgregor: Yeah, it's a string value. I'm only comparing their addresses. Figured if all those properties matched, they're likely the same `AssetAccount`

woudini: Ah I misread this line. Thanks

woudini:
Looks great 👍

PR #527, opened 2018-09-24
fix(KYC): correctly present ExchangeCoordinator

achtenhagen:
## Objective

Correctly present the `ExchangeCoordinator` on the _account approved_ screen, when the user taps on **Get Started**.

## Description

By storing a reference to the rootViewController, the `ExchangeCoordinator` is able to be presented correctly.

## How to Test

Using a verified user, open Settings and tap on the _Identity Verification_ cell. On the _Account Approved_ screen, tap **Get Started**. Observe that it presents the Exchange correctly.

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/KYC/KYCCoordinator.swift
+++ b/Blockchain/KYC/KYCCoordinator.swift
@@ -53,6 +53,8 @@ protocol KYCCoordinatorDelegate: class {
 
     private(set) var country: KYCCountry?
 
+    private var rootViewController: UIViewController?
+
     fileprivate var navController: KYCOnboardingNavigationController!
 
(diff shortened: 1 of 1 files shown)


Blockchain/KYC/KYCCoordinator.swift
```diff
@@ -53,6 +53,8 @@ protocol KYCCoordinatorDelegate: class {
 
     private(set) var country: KYCCountry?
 
+    private var rootViewController: UIViewController?
```
bchrisarriola: This should probably be `weak` since `KYCCoordinator` is a singleton

PR #531, opened 2018-09-25
fix(Order history): IOS-1339 - Not showing pending trade

thisisalexmcgregor:
## Objective

Fixes issue where we were not showing the user's order history.

## Description

This endpoint requires a different date format than our session token endpoint. Additionally, some of the values on our trade model need to be optional. Lastly, there's an issue with the `pair` property. You're not supposed to ever have a pair with two of the same asset types. However, there was a server side bug where this could feasibly happen. This has been fixed on the server however, if you did submit trades while this bug was still around, you would get an assertion as `pair` would be nil. 

I'm waiting to hear from the backend team to see if these trades could be purged. 

## How to Test

1. Build and run
2. Go to the Exchange List screen
3. If you've submitted trades on HB, you should see those trades now.
4. Try pulling to refresh. 
5. If you've submitted over 50 trades, try paginating. 

simulator screen shot - iphone x - 2018-09-25 at 12 55 32

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Controllers/ExchangeListViewController.swift
+++ b/Blockchain/Homebrew/Exchange/Controllers/ExchangeListViewController.swift
@@ -51,6 +51,7 @@ class ExchangeListViewController: UIViewController {
         dependenciesSetup()
         dataProvider?.delegate = self
         delegate?.onLoaded()
+        registerForNotifications()
     }
     
     fileprivate func dependenciesSetup() {
@@ -61,9 +62,20 @@ class ExchangeListViewController: UIViewController {
         delegate = presenter
     }
     
+    fileprivate func registerForNotifications() {
+        NotificationCenter.when(Constants.NotificationKeys.exchangeSubmitted) { [weak self] _ in
+            guard let this = self else { return }
+            this.delegate?.onLoaded()
+        }
+    }
+    
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         delegate?.onDisappear()
+    }
+    
+    deinit {
+        NotificationCenter.default.removeObserver(self)
     }
 }
 
--- a/Blockchain/Homebrew/Exchange/Models/CellModels/ExchangeCellModel.swift
+++ b/Blockchain/Homebrew/Exchange/Models/CellModels/ExchangeCellModel.swift
@@ -30,7 +30,14 @@ enum ExchangeCellModel {
 
         var descriptionActionBlock: LabelAction?
 
-        init(description: String, value: String, backgroundColor: UIColor = .white, statusVisibility: Visibility = .hidden, bold: Bool = false, statusTintColor: UIColor = .green) {
+        init(
+            description: String,
+            value: String,
+            backgroundColor: UIColor = .white,
+            statusVisibility: Visibility = .hidden,
+            bold: Bool = false,
+            statusTintColor: UIColor = .green
+            ) {
             self.description = description
             self.value = value
             self.backgroundColor = backgroundColor
--- a/Blockchain/Homebrew/Exchange/Models/CellModels/ExchangeCellModel.swift
+++ b/Blockchain/Homebrew/Exchange/Models/CellModels/ExchangeCellModel.swift
@@ -25,16 +25,18 @@ enum ExchangeCellModel {
         let value: String
         let backgroundColor: UIColor
         let statusVisibility: Visibility
+        let statusTintColor: UIColor
         let bold: Bool
 
         var descriptionActionBlock: LabelAction?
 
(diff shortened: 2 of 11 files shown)


Blockchain/Homebrew/Exchange/Models/TradingPair.swift
```diff
@@ -13,9 +13,14 @@ struct TradingPair {
...
 
     init?(from: AssetType, to: AssetType) {
-        guard from != to else {
+        // TODO: This used to return early with `nil`.
+        // Some have already executed trades though when
+        // there was a server side bug that allowed for the
+        // two asset types to be the same. So, when those of
+        // us who have that in our trade history GET trades
+        // we hit an assertion error. 
+        if from == to {
```
woudini: I'm hesitant to merge this as we would be making our own code more error prone in order to accommodate to the backend's errors. Where exactly is this assertion error? Is it possible to throw the error a bit closer to the UI rather than change this model?

thisisalexmcgregor: Yeah, hesitant as well. Didn't hear back yet from the backend team regarding the trade. 

What I can do is remove this and whenever I need to look at my trades, I can comment this out. If it's happening to all of us, then we can address that then.

woudini: Saw your comment about this possibly being blocked, I would definitely wait

thisisalexmcgregor: Sounds good, I'll flag it as such and we can follow up tomorrow.

thisisalexmcgregor: 💥 backend fixed this issue so I reverted this change.

thisisalexmcgregor:
@woudini the ones that I noticed were null I put in `decodeIfPresent`, and actually I don't believe we are using `rate` anywhere, or the two `txhash` values.

Blockchain/Homebrew/Exchange/Controllers/ExchangeListViewController.swift
```diff
@@ -61,10 +62,21 @@ class ExchangeListViewController: UIViewController {
         delegate = presenter
     }
     
+    fileprivate func registerForNotifications() {
+        NotificationCenter.when(Constants.NotificationKeys.exchangeSubmitted) { [weak self] _ in
```
woudini: Can this event be handled by the `ExchangeCoordinator` instead of using notifications? It has a direct reference to the `ExchangeListViewController`.
https://github.com/blockchain/My-Wallet-V3-iOS/pull/531/files#diff-19e39398b6c5b87d10760ec1601dfd1dR264

thisisalexmcgregor: Didn't know we had that. Though it doesn't look like we assign `exchangeListViewController` to anything.

thisisalexmcgregor: Also `ExchangeDetailCoordinator` is using the `shared` version of `ExchangeCoordinator`. I'm guessing that'll change to being injected? I think since we plan on decoupling and not using `ExchangeCoordinator` as a singleton we keep this as is for now.

Blockchain/Homebrew/Exchange/Models/CellModels/ExchangeCellModel.swift
```diff
@@ -25,16 +25,18 @@ enum ExchangeCellModel {
...
         let backgroundColor: UIColor
         let statusVisibility: Visibility
+        let statusTintColor: UIColor
         let bold: Bool
 
         var descriptionActionBlock: LabelAction?
 
-        init(description: String, value: String, backgroundColor: UIColor = .white, statusVisibility: Visibility = .hidden, bold: Bool = false) {
+        init(description: String, value: String, backgroundColor: UIColor = .white, statusVisibility: Visibility = .hidden, bold: Bool = false, statusTintColor: UIColor = .green) {
             self.description = description
```
woudini: Swiftlint line length violation here, try
```
        init(
            description: String,
            value: String,
            backgroundColor: UIColor = .white,
            statusVisibility: Visibility = .hidden,
            bold: Bool = false,
            statusTintColor: UIColor = .green
        ) {
```

thisisalexmcgregor: I really need to setup swift lint...

thisisalexmcgregor: Done.

PR #533, opened 2018-09-25
fix(Homebrew-Exchange): IOS-1358 make ExchangeCoordinator inject ExchangeServices into view controllers

woudini:
## Objective

Make sure all amount label are cleared when returning to the New Exchange screen from the list controller.

## Description

~Cleared the secondary amount label on `viewDidLoad`. Looking for feedback on whether this is the best place for it inside this method.~

Found out that the state of the secondary output is retained in the `ExchangeConversionService`. Discussion with @bchrisarriola concluded that it would be best to inject `ExchangeServices` to ensure new instances for each.

Also did a little cleaning up:
- The `ExchangeDetailInteractor` doesn't need the `ExchangeConversionService` so I removed that.
- The `ExchangeDetailViewController.PageModel` doesn't need to be instantiated with dependencies now that the `ExchangeDetailCoordinator` has it, so I removed the dependencies parameter from the `.confirm(...)` case.

## How to Test

Go to Exchange, type in a value, go back one screen, go back to Exchange - input labels should be zeroed out and reset.

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Coordinators/ExchangeCoordinator.swift
+++ b/Blockchain/Coordinators/ExchangeCoordinator.swift
@@ -63,7 +63,6 @@ struct ExchangeServices: ExchangeDependencies {
     private let walletManager: WalletManager
 
     private let walletService: WalletService
-    private let dependencies: ExchangeDependencies = ExchangeServices()
 
     private var disposable: Disposable?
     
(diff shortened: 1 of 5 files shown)


woudini:
Looked further into this and I'm seeing `ExchangeCreateViewController` being `deinit`ed - but the real issue is that the services aren't reset - they're initialized in the `ExchangeServices` `init` and passed everywhere: https://github.com/blockchain/My-Wallet-V3-iOS/blob/b78794acb7927a17aa5c86fe96718fd40f21e9e7/Blockchain/Coordinators/ExchangeCoordinator.swift#L37
So the `ExchangeCreateViewController` views are being re-created but right after that the label text is being re-populated with an older conversion output.

Maybe these services should conform to a protocol that has a method `reset()`, and the `ExchangeCoordinator` should be responsible for calling `reset()`?

woudini:
Although one could also argue that having to reset APIs is another symptom, since API's shouldn't really have any state in the first place

PR #535, opened 2018-09-25
fix(Account): remove 'Invalid email address' alert when creating another account

bchrisarriola:
## Objective

Fixes bug where user cannot create another account.

## Description

User was not able to create a new account because the entered account label was being evaluated as an email address. Looks like it may have been added by accident in bf0eb1d. Found this issue while testing #534 

## How to Test

Pull in changes, try to create account (Tap "Addresses" in the side menu -> "+" button in the Wallets header for Bitcoin), and verify that it succeeds.

## Merge Checklist

- [X] All unit tests pass.


bchrisarriola:
> The HD account has always been difficult to name - good example of how just naming it account causes confusion

💯 agree

PR #537, opened 2018-09-26
fix(Exchange): fix issue where partner exchange can't be launched.

bchrisarriola:
## Objective

Fixes issue where partner exchange cannot be launched when tapping on the exchange side menu. This was a regression introduced in 91891143.

## How to Test

Launch the app with an approved nabu user (but have "enable homebrew" in the debug options turned off), tap on "Exchange", verify that the partner exchange is launched.

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/View Controllers/PartnerExchangeListViewController.m
+++ b/Blockchain/View Controllers/PartnerExchangeListViewController.m
@@ -34,7 +34,7 @@
 
 @implementation PartnerExchangeListViewController
 
-+ (PartnerExchangeListViewController * _Nonnull)createWithCountryCode:(NSString *_Nonnull)countryCode
++ (PartnerExchangeListViewController * _Nonnull)createWithCountryCode:(NSString *_Nullable)countryCode
 {
     PartnerExchangeListViewController *controller = [[PartnerExchangeListViewController alloc] init];
     controller.countryCode = countryCode;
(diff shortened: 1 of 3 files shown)


Blockchain/View Controllers/PartnerExchangeListViewController.m
```diff
@@ -49,13 +49,15 @@ - (void)viewDidLoad
     
     self.view.backgroundColor = UIColor.lightGray;
 
-    if ([self.countryCode  isEqual: @"US"]) {
+    Wallet *wallet = WalletManager.sharedInstance.wallet;
+    NSString *countryCode = (self.countryCode != nil) ? self.countryCode : wallet.countryCodeGuess;
```
woudini: The KYC country code probably shouldn't be used here - the `wallet.countryCodeGuess` should only apply, since the user can't choose it and try to pretend they're in a different country.

Blockchain/View Controllers/PartnerExchangeListViewController.m
```diff
@@ -49,13 +49,15 @@ - (void)viewDidLoad
...
+    NSArray *availableStates = wallet.availableUSStates;
+    if ([countryCode  isEqual: @"US"] && availableStates.count > 0) {
```
woudini: Might sound trivial but adding `availableStates.count > 0` here does change the potential behavior slightly (though probably not in any meaningful way) but if the goal is to keep the same behavior it should be removed.

Blockchain/View Controllers/PartnerExchangeListViewController.m
bchrisarriola: @woudini the reason this was added was because the user can select a country like "Andorra" but then asked what US state they are in, which is odd.

Blockchain/View Controllers/PartnerExchangeListViewController.m
bchrisarriola: I added that because it looks like `wallet.availableUSStates` will return an empty array when the user has already selected a USA state for shapeshift. So we need to check that that array is non-empty, otherwise, you see the state selection screen but w/ no states displayed.

Blockchain/View Controllers/PartnerExchangeListViewController.m
woudini: Had to dig further to see that `isInPartnerRegionForExchange` is called before this anyway - all good then 👍

PR #545, opened 2018-09-26
feat(NabuNetworkError): unmarshall error payloads returned by Nabu into NabuNetworkError, if possible

bchrisarriola:
## Objective

Created `NabuNetworkError` to allow closer inspection on 4xx type HTTP errors returned by the server.

## Description

Nabu returns us a very specific error code when a network request fails. For example, we get the following error message when trying to go through the exchange flow with an email associated to a wallet:

```
{
    "id":"someId",
    "code":33,
    "type":"CONFLICT",
    "description":"User already exists"
}
```

At the moment, the client simply logs these kind of messages and does not propagate it back up to the caller of the network request.

With `NabuNetworkError`, callers can now have closer inspection on the error message returned by the server which in turn allows us to show a more user-friendly error message to the user. At the moment, it isn't used anywhere but wanted to get this PR in so that we can start using it.

## How to Test

0. Put a breakpoint on line 226 in `KYCNetworkRequest`
1. Create a wallet with an existing user
2. Tap on "Exchange"
3. Notice that the breakpoint is hit, and that `errorPayload` has a value.

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Models/NabuNetworkError.swift
+++ b/Blockchain/Homebrew/Models/NabuNetworkError.swift
@@ -1,0 +1,107 @@
+//
+//  NabuNetworkError.swift
+//  Blockchain
+//
+//  Created by Chris Arriola on 9/26/18.
+//  Copyright © 2018 Blockchain Luxembourg S.A. All rights reserved.
+//
+
+import Foundation
+
+/// Describes an error returned by Nabu
+struct NabuNetworkError: Codable, Error {
+    let code: NabuNetworkErrorCode
+    let type: NabuNetworkErrorType
+    let description: String
+}
+
+enum NabuNetworkErrorCode: Int, Codable {
+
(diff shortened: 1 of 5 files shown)


Blockchain/Homebrew/Models/NabuNetworkError.swift
```diff
@@ -0,0 +1,107 @@
...
+    case notSupportedCurrencyPair = 16
+    case unknownCurrencyPair = 17
+    case unknownCurrency = 18
+    case currencyIsNotFiat = 19
+    case tooSmallVolume = 26
+    case tooBigVolume = 27
+    case resultCurrencyRatioTooSmall = 28
+
+    // conversion errors
+    case providedVolumeIsNotDouble = 20
```
woudini: Wondering where you found these. Are all of these documented in the wiki somewhere?

PR #546, opened 2018-09-27
fix(HB): IOS-1368 - `Withdrawal` SymbolValue can sometimes be nil resulting in no trade history

thisisalexmcgregor:
## Objective

Fixes an issue where `withdrawal` is `nil` in the event of a failed trade. Also fixes some UI issues including a incorrect color for a back button and missing currency symbols. 

## How to Test

1. Build and run
2. Go to the Exchange List
3. See that you can now see your prior exchanges (even the failed ones)

simulator screen shot - iphone se - 2018-09-27 at 12 48 15

Design is still TBD for handling `Failed` state. Likely remains the same but possible that `zero` is inserted vs `nil` to account for a trade that didn't complete. 

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Controllers/ExchangeListViewController.swift
+++ b/Blockchain/Homebrew/Exchange/Controllers/ExchangeListViewController.swift
@@ -52,6 +52,10 @@ class ExchangeListViewController: UIViewController {
         dataProvider?.delegate = self
         delegate?.onLoaded()
         registerForNotifications()
+        
+        if let controller = navigationController as? BCNavigationController {
+            controller.applyDarkAppearance()
+        }
     }
     
     fileprivate func dependenciesSetup() {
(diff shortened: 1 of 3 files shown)


Blockchain/Homebrew/Exchange/Controllers/ExchangeListViewController.swift
```diff
@@ -52,6 +52,10 @@ class ExchangeListViewController: UIViewController {
         dataProvider?.delegate = self
         delegate?.onLoaded()
         registerForNotifications()
+        
+        if let controller = navigationController as? BCNavigationController {
+            controller.applyDarkAppearance()
```
bchrisarriola: Should this be moved to `viewWillAppear`? This way, in case another view controller is added to the stack that changes the appearance, it can be reverted to dark when popped from the stack.

thisisalexmcgregor: Done ✔️

PR #547, opened 2018-09-27
fix(Websocket): close exchange WS when no longer needed.

bchrisarriola:
## Objective

Close exchange WS when no longer needed.

## Description

Fixes issue where quote related WS messages were still received even when the user is no longer on the exchange create screen.

Note about unit tests and directory structures... The location of the test file should mimic the location and directory structure of the file being tested (e.g. https://cloudup.com/cD9DjzafCct).

## How to Test

1. Go to exchange create
2. Verify that you see "Websocket connected..."
3. Exit out of exchange
4. Verify that you see "Websocket disconnected..."

## Merge Checklist

- [X] All unit tests pass.
- [X] You have added unit tests.

### DIFF
--- a/Blockchain/Network/SocketManager.swift
+++ b/Blockchain/Network/SocketManager.swift
@@ -101,6 +101,7 @@ class SocketManager {
 
 extension SocketManager: WebSocketAdvancedDelegate {
     func websocketDidConnect(socket: WebSocket) {
+        Logger.shared.debug("Websocket connected to: \(socket.currentURL.absoluteString)")
         if socket == self.exchangeSocket {
             pendingSocketMessages.forEach { [unowned self] in
                 self.send(message: $0)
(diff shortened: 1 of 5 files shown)


Blockchain/Network/SocketManager.swift
```diff
@@ -160,6 +161,7 @@ extension SocketManager: WebSocketAdvancedDelegate {
 
     func websocketDidDisconnect(socket: WebSocket, error: Error?) {
         // Required by protocol
+        Logger.shared.debug("Websocket disconnected to: \(socket.currentURL.absoluteString)")
```
woudini: "disconnected from"

BlockchainTests/Mock/MockSocketManager.swift
```diff
@@ -0,0 +1,24 @@
...
+
+    var didCallSetup: XCTestExpectation?
+    var didCallDisconnect: XCTestExpectation?
+
+    override func setupSocket(socketType: SocketType, url: URL) {
+        didCallSetup?.fulfill()
+    }
+
+    override func disconnect(socketType: SocketType) {
+        didCallDisconnect?.fulfill()
```
woudini: Should mock classes ever call their corresponding superclass methods? Seems like most of them in the app don't.

bchrisarriola: I think that depends on the method being mocked. If calling the superclass's method has a side-effect that we want to test, then yeah, it should call super's method. In this case, we don't want to call super because we don't actually want the test to make websocket connection/disconnections (just need to validate that those methods are invoked).

PR #549, opened 2018-09-27
fix(Homebrew-Exchange): IOS-1348 add dismiss handler to password confirm view

woudini:
## Objective

Allow creating and sending exchange order with second password.

## Description
- Added property `dismissHandler` to `PasswordConfirmView` to allow setting `tradeExecution.isExecuting` back to `false` after dismissing `PasswordConfirmView`
- Fixed issue where `bitcoincash:` prefix could be added twice
- Fixed some missing parameter bugs in JS

## How to Test

Submit and send an exchange order with second password enabled.

N/A

## Merge Checklist

- [X] All unit tests pass.

### DIFF
--- a/Blockchain/Authentication/AuthenticationCoordinator.swift
+++ b/Blockchain/Authentication/AuthenticationCoordinator.swift
@@ -629,14 +629,14 @@ extension AuthenticationCoordinator: SetupDelegate {
 }
 
 extension AuthenticationCoordinator: WalletSecondPasswordDelegate {
-    func getSecondPassword(success: WalletSuccessCallback, dismiss: WalletDismissCallback) {
+    func getSecondPassword(success: WalletSuccessCallback, dismiss: WalletDismissCallback?) {
         showPasswordConfirm(withDisplayText: LocalizationConstants.Authentication.secondPasswordDefaultDescription,
                             headerText: LocalizationConstants.Authentication.secondPasswordRequired,
                             validateSecondPassword: true,
                             confirmHandler: { (secondPassword) in
                                 success.success(string: secondPassword)
                             },
-                            dismissHandler: { dismiss.dismiss() }
+                            dismissHandler: { dismiss?.dismiss() }
         )
     }
 
--- a/Blockchain/Authentication/AuthenticationCoordinator.swift
+++ b/Blockchain/Authentication/AuthenticationCoordinator.swift
@@ -427,7 +427,8 @@ import RxSwift
         withDisplayText displayText: String,
         headerText: String,
         validateSecondPassword: Bool,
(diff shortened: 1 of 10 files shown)


Blockchain/Coordinators/ExchangeCoordinator.swift
```diff
@@ -110,10 +110,12 @@ struct ExchangeServices: ExchangeDependencies {
                 AuthenticationCoordinator.shared.showPasswordConfirm(
                     withDisplayText: LocalizationConstants.Authentication.etherSecondPasswordPrompt,
                     headerText: LocalizationConstants.Authentication.secondPasswordRequired,
-                    validateSecondPassword: true
-                ) { (secondPassword) in
-                    WalletManager.shared.wallet.createEthAccount(forExchange: secondPassword)
-                }
+                    validateSecondPassword: true,
+                    confirmHandler: { (secondPassword) in
+                        WalletManager.shared.wallet.createEthAccount(forExchange: secondPassword)
```
bchrisarriola: This should use the `walletManager` instance defined in `ExchangeCoordinator` and not reference the `WalletManager` singleton directly (same comment goes for all other `WalletManager.shared` references).

Blockchain/Authentication/AuthenticationCoordinator.swift
```diff
@@ -427,7 +427,8 @@ import RxSwift
         withDisplayText displayText: String,
         headerText: String,
         validateSecondPassword: Bool,
-        confirmHandler: @escaping PasswordConfirmView.OnPasswordConfirmHandler
+        confirmHandler: @escaping PasswordConfirmView.OnPasswordConfirmHandler,
+        dismissHandler: PasswordConfirmView.OnPasswordDismissHandler?
```
bchrisarriola: This can be set to `= nil` so that if the caller doesn't need to do anything upon dismissal, it doesn't need to pass a block (only on Swift, not obj-c, unfortunately).

Blockchain/Models/Wallet.h
```diff
@@ -462,8 +462,26 @@
 - (void)updateKYCUserCredentialsWithUserId:(NSString *)userId lifetimeToken:(NSString *)lifetimeToken success:(void (^ _Nonnull)(NSString *_Nonnull))success error: (void (^ _Nonnull)(NSString *_Nullable))error;
 - (NSString *_Nullable)KYCUserId;
 - (NSString *_Nullable)KYCLifetimeToken;
+
+/// Call this method to build an Exchange order.
+/// It constructs and stores a payment object with a given AssetType, to, from, and amount (properties of OrderTransactionLegacy).
+/// To send the order, call sendOrderTransaction:completion:success:error:cancel.
```
bchrisarriola: Thanks for these 🔥

PR #562, opened 2018-10-02
fix(Homebrew-Exchange): IOS-1382 remove order ID placeholder and display order ID in Exchange Locked screen

woudini:
## Objective

Show order ID in the Exchange Locked screen.

## Description

- Added mutable string property `orderIdentifier` to `OrderTransaction`.
- Set `orderIdentifier` from the `OrderResult`'s `id` returned from the trades endpoint.
- Set the `orderIdentifier` on the `OrderTransaction` just before calling the `.sentTransaction` event.
- Added order ID to the Exchange Locked screen as a new cell.

## How to Test

Send an Exchange Order and see order ID displayed. You should be able to tap and copy the order ID to the clipboard.

Not styled the same way as in the spec (in a single line).
5S:
5s
6:

6 plus:
6 plus

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Models/Order.swift
+++ b/Blockchain/Homebrew/Exchange/Models/Order.swift
@@ -72,6 +72,8 @@ struct OrderResult: Codable {
 }
 
 struct OrderTransaction {
+    var orderIdentifier: String?
+
     // The destination is where the user will ultimately receive
     // funds from the exchange.
     let destination: AssetAccount
--- a/Blockchain/Homebrew/Exchange/Models/Order.swift
+++ b/Blockchain/Homebrew/Exchange/Models/Order.swift
@@ -74,7 +74,7 @@ struct OrderResult: Codable {
 struct OrderTransaction {
     // The destination is where the user will ultimately receive
     // funds from the exchange.
-    let destination: String
+    let destination: AssetAccount
 
     // Details of payment constructed in Wallet JS
     let from: AssetAccount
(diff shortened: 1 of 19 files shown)


Blockchain/Homebrew/Exchange/Services/TradeExecutionService.swift
```diff
@@ -283,11 +285,13 @@ extension TradeExecutionService {
...
     // Used as the final step to convert Exchange information into built payment
     // and immediately sending the order.
+    //
+    // For the success handler, the string returned is the order identifier.
     func buildAndSend(
         with conversion: Conversion,
         from: AssetAccount,
         to: AssetAccount,
-        success: @escaping (() -> Void),
+        success: @escaping ((String?) -> Void),
```
bchrisarriola: If this call succeeds, shouldn't the ID be non-nil? In other words, shouldn't this pass a `String` and not a `String?`?

Alternatively, what about just passing the `OrderTransaction` object through?

So:
```
    succes: @escaping ((OrderTransaction) -> Void),
```

Blockchain/Homebrew/Exchange/Models/Order.swift
```diff
@@ -72,6 +72,8 @@ struct OrderResult: Codable {
 }
 
 struct OrderTransaction {
+    var orderIdentifier: String?
```
bchrisarriola: Although there are some cases where mutable properties are ok to use, I think for data models like this, we should always prefer immutable properties. With immutability, we can guarantee that something won't change and so reasoning about side-effects are easy (i.e. there are no side-effects since a model's properties are guaranteed not to change).

Long story short, I think this should be a `let` instead. Once the `orderIdentifier` is defined, just create a new `OrderTransaction` with the defined `orderIdentifier`.

PR #564, opened 2018-10-03
fix(Wallet): IOS-1387 fetch balances after sync

woudini:
## Objective

Fix issue where creating an HD account resulted in zero balances displayed in the dashboard and transactions tab.

## Description

After a successful wallet backup, the views were reloading while balances were `null`. Made `getHistoryForAllAssets` public and called it in `didBackupSuccess` and `didBackupError`. Also removed a couple of busy view and reload calls since `getHistoryForAllAssets` will ultimately hide the busy view.

## How to Test

Create an HD account in Addresses, and see that the balances are unchanged.

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Coordinators/AppCoordinator.swift
+++ b/Blockchain/Coordinators/AppCoordinator.swift
@@ -342,7 +342,7 @@ extension AppCoordinator: WalletAccountInfoAndExchangeRatesDelegate {
 
 extension AppCoordinator: WalletBackupDelegate {
     func didBackupWallet() {
-        reload()
+        walletManager.wallet.getHistoryForAllAssets()
     }
 
     func didFailBackupWallet() {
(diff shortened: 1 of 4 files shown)


woudini:
It's a poorly named method. You're correct, backup (as in writing down the mnemonic) is not related to creating a new HD account - "backup wallet" in the JS is synonymous with "sync wallet".

bchrisarriola:
> It's a poorly named method. You're correct, backup (as in writing down the mnemonic) is not related to creating a new HD account - "backup wallet" in the JS is synonymous with "sync wallet".

I see. We can at least change the Swift code to use "sync" instead (would be nice to rename in JS but I imagine that would be changing a lot of references in `My-Wallet-V3`?)

woudini:
We can do that. At first glance it doesn't look like My-Wallet-V3 has many instances of the name `backup`.

PR #565, opened 2018-10-03
feat(HB): IOS-1319 & IOS-1336 - Error handling in the Exchange

thisisalexmcgregor:
## Objective

⚠️Large PR

This PR includes a few things:
* Significant refactor of how we handle UI updates in `ExchangeCreateViewController`
* Additional classes for having a completion callback that executes an array of `CompletionEvent` types. You can use this on regular UIView animated updates and UIView transition animations. This was required for both as animating a color change of a UILabel is a transition. 
* Additional logic in how we fetch the `mix` and `max` values for trade limits. We now wrap these two requests in an `Observable`. When we receive them we compare them to the user input. **NOTE**: This only reports in fiat currently. 

⚠️This includes some changes from @woudini. Some of those changes have been removed as we learned that `TradeLimits` takes into account the user's balance. That said, not all of it was removed. I'll leave what should and shouldn't be removed up to @woudini.  

Lastly, this code isn't as documented as I would like, and the naming of certain `typealiases` need to be updated. I will do this in a follow up PR. 

## Description

`ExchangeCreatePresenter` now has an `InternalEvent` enum. Currently there's only one type of `.block`. It's a `CompletionEvent`. When a `TransitionUpdateGroup` or `ViewUpdateGroup` completes all of its animations, it calls `finish()`. If it has a completion handler, it calls all the `CompletionEvents`. Here's an example:

```
let completion: ViewUpdateBlock = { [weak self] internalEvents in
            guard let events = internalEvents else { return }
            events.forEach({ this.handle(internalEvent: $0) })
        }
        
        let block = { [weak self] in
            this.setErrorDisappearanceTimer(duration: 2.5)
        }
        
        let group = ViewUpdateGroup(
            preparations: [.secondaryLabel(.hidden)],
            animations: [.errorLabel(.visible)],
            animation: .standard(duration: 0.2),
            completionEvents: [.block(block)],
            completion: completion
        )
        interface?.apply(presentationUpdateGroup: group)
```
We show the `errorLabel` using a `standard` animation with a duration of `0.2`. When it completes it calls the `block` which sets a `Timer` for a duration of `2.5` seconds. When that timer completes, we hid the `errorLabel`:
```
fileprivate func setErrorDisappearanceTimer(duration: TimeInterval) {
        errorDisappearenceTimer?.invalidate()
        errorDisappearenceTimer = Timer(
            fire: Date(timeIntervalSinceNow: duration),
            interval: 0,
            repeats: false,
            block: { [weak self] timer in
                self?.errorDisappearanceTimerFired()
        })
        guard let timer = errorDisappearenceTimer else { return }
        RunLoop.main.add(timer, forMode: .commonModes)
    }
```

## How to Test

1. Build and run
2. Go to the exchange
3. Try to make trades above your min or max limit
4. The primary label should turn red, the primary label should wiggle, and the error label should appear. 

simulator screen shot - iphone x - 2018-10-03 at 13 16 00

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -236,33 +236,56 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
+                        }, error: { [weak self] errorMessage in
+                            guard let this = self else { return }
+                            AlertViewPresenter.shared.standardError(message: errorMessage)
+                            this.output?.loadingVisibility(.hidden)
+                        }
+                    )
+                }
+            })
+        
+        _ = disposables.insert(disposable)
```
bchrisarriola: You can use the helper method I added `insertWithDiscardableResult` so you don't have to set the value to `_`.

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -327,6 +352,42 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
+    }
+    
+    private func minTradingLimit() -> Maybe<Decimal> {
+        guard let model = model else {
+            return Maybe.empty()
+        }
+        
+        return tradeLimitService.getTradeLimits(
+            withFiatCurrency: model.fiatCurrencyCode
+            ).map { tradingLimits -> Decimal in
```
bchrisarriola: Nit: indentation

Blockchain/Network/Models/SocketMessage.swift
```diff
@@ -161,6 +161,55 @@ struct Conversion: SocketMessageCodable {
...
+    
+    var baseCryptoSymbol: String {
+        return quote.currencyRatio.base.crypto.symbol
+    }
+    
+    var baseCryptoValue: String {
+        return quote.currencyRatio.base.crypto.value
+    }
+    
+    var baseToFiatDescription: String {
```
bchrisarriola: `baseToFiatDescription`, `baseToCounterDescription` and `counterToFiatDescription` can be removed. I think these may have been added back by accident?

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -236,33 +236,56 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
+                switch candidate {
+                case ..<minValue:
+                    
+                    let value = NumberFormatter.localCurrencyFormatter.string(for: minValue) ?? ""
+                    let minimum = model.fiatCurrencySymbol + value
+                    
+                    output.entryBelowMinimumValue(minimum: minimum)
+                case maxValue..<Decimal.greatestFiniteMagnitude:
+                    guard let maximum = NumberFormatter.localCurrencyFormatter.string(for: maxValue) else { return }
+                    output.entryAboveMaximumValue(maximum: maximum)
```
woudini: Should this have the fiatCurrencySymbol also, similar to line 260? i.e.:
```
                    let value = NumberFormatter.localCurrencyFormatter.string(for: maxValue) ?? ""
                    let maximum = model.fiatCurrencySymbol + value
                    output.entryAboveMaximumValue(maximum: maximum)
```

thisisalexmcgregor: Done e0fcd37

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
```diff
@@ -40,10 +53,11 @@ protocol ExchangeCreateOutput: class {
     func entryRejected()
     func styleTemplate() -> ExchangeStyleTemplate
     func updatedInput(primary: NSAttributedString?, secondary: String?)
-    func updatedInput(primary: String?, primaryDecimal: String?, secondary: String?)
     func updatedRates(first: String, second: String, third: String)
     func updateTradingPairValues(left: String, right: String)
     func updateTradingPair(pair: TradingPair, fix: Fix)
-    func loadingVisibility(_ visibility: Visibility, action: ExchangeCreateViewController.Action)
+    func entryBelowMinimumValue(minimum: String)
```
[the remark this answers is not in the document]
thisisalexmcgregor: It could be but, I wouldn't think the `ExchangeCreateOutput` would have a protocol that gave any indication of UI functionality. I try to keep UI looking functions in just the interface otherwise I get kind of lost when hoping between interactor, presenter, interface, etc. 

Then again we have `loadingVisibility` 🤷‍♂️

Blockchain/AssetType.swift
```diff
@@ -100,4 +100,44 @@ extension AssetType {
...
+        case .bitcoin:
+            let value = NumberFormatter.formatMoney(
+                amount.uint64Value,
+                localCurrency: true
+            )
+            return value
+        case .ethereum:
+            let value = NumberFormatter.formatEthToFiat(
+                withSymbol: amount.stringValue,
+                exchangeRate: WalletManager.shared.wallet.latestEthExchangeRate
```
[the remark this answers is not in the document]
thisisalexmcgregor: Done `a939785`

Blockchain/AssetType.swift
```diff
@@ -100,4 +100,44 @@ extension AssetType {
...
+        case .bitcoinCash:
+            let value = NumberFormatter.formatBch(
+                withSymbol: amount.uint64Value,
+                localCurrency: true
+            )
+            return value
+        }
+    }
+    
+    func toCrypto(amount: NSDecimalNumber) -> String? {
```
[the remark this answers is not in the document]
thisisalexmcgregor: Done `a939785`

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
thisisalexmcgregor: Done bff26f3

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
thisisalexmcgregor: Done bff26f3

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -236,33 +236,56 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
-                this.output?.loadingVisibility(.hidden, action: ExchangeCreateViewController.Action.createPayment)
-                this.output?.showSummary(orderTransaction: orderTransaction, conversion: conversion)
-            }, error: { [weak self] errorMessage in
+        guard let output = output else { return }
+        
+        let min = minTradingLimit().asObservable()
+        let max = maxTradingLimit().asObservable()
+        let disposable = Observable.zip(min, max) {
+            return ($0, $1)
+            }.subscribe(onNext: { [weak self] payload in
```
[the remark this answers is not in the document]
thisisalexmcgregor: be33729

Blockchain/AssetType.swift
```diff
@@ -100,4 +100,53 @@ extension AssetType {
...
+    func toFiat(
+        amount: Decimal,
+        from wallet: Wallet = WalletManager.shared.wallet
+        ) -> String? {
+        let input = amount as NSDecimalNumber
+        
+        switch self {
+        case .bitcoin:
+            let value = NumberFormatter.formatMoney(
+                input.uint64Value,
```
woudini: I think this and line 125 need to be something like
```
NumberFormatter.parseBtcValue(from: input.stringValue),
```
Because an example `input` can be `0.00015026` and calling `uint64Value` will change it to `0`.

thisisalexmcgregor: Line 125 is `BCH`. Is there a `BCH` equivalent?

woudini: BCH has the same number of decimal places as BTC, so you can use the same one. But you bring up a good point of that method having a not-so-clear name. BTC related methods are used for BCH in several places in the app.

thisisalexmcgregor: Done: 20200ec

PR #569, opened 2018-10-03
 fix(Settings): fix flickering account status badge (IOS-1286)

bchrisarriola:
## Objective

Fix flickering KYC account status badge in settings view.

## Description

The badge was previously flickering because each time that cell was being rendered, it would first set it to "Unverified", make a network call to get the current NabuUser, and finally update the state of the badge to the correct KYC account status.

This PR fixes that issue by (1) not initializing the badge to "Unverified", and (2) using BlockchainDataRepository's cached version of the `NabuUser`. Since it's using the cached value which is in memory, you should no longer see the badge flicker. Notice though that I changed the property in BlockchainDataRepository to now be an `Observable<NabuUser>` so that it will first emit the cached value if available, followed by the network value.

I've also added an initialization step in `AuthenticationCoordinator` so that upon authenticating, the BlockchainDataRepository cache would be initialized. One (positive) side-effect of this, is we also no longer see a delay in the "Exchange" side menu option.

## How to Test

Go to settings.

## Merge Checklist

- [X] All unit tests pass.
- [X] You have added sufficient documentation (descriptive comments).

### DIFF
--- a/Blockchain/Authentication/AuthenticationCoordinator.swift
+++ b/Blockchain/Authentication/AuthenticationCoordinator.swift
@@ -61,6 +61,8 @@ import RxSwift
         tabControllerManager.sendBitcoinViewController?.reload()
         tabControllerManager.sendBitcoinCashViewController?.reload()
 
+        strongSelf.dataRepository.prefetchData()
+
         /// Prompt the user for push notification permission
         PushNotificationManager.shared.requestAuthorization()
 
@@ -108,6 +110,8 @@ import RxSwift
         }
     }
 
+    internal let dataRepository: BlockchainDataRepository
+
     internal let walletManager: WalletManager
 
     private let walletService: WalletService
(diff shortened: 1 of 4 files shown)


woudini:
The flickering is gone, but the badge is still moving around when tapping on it and returning the Settings screen - which may or may not be in the scope of this PR.

movingbadge

bchrisarriola:
Hmmm something weird with constraints. will look into

thisisalexmcgregor:
@woudini Saw that as well. Wasn't sure if it was worth pointing out only because all the badges seem to be uneven across the cells that display a badge.

bchrisarriola:
K. Flickering badge should be fixed in 881b5682

Issue was that `sizeToFit` and `layoutIfNeeded` was called but the badge's text would be changed later on. Fix was to set the text first.

woudini:
881b568 fixes it for a KYC verified wallet but the badge flickers quite a bit still for a new wallet.

PR #571, opened 2018-10-04
fix(Homebrew-Exchange): IOS-1363 UI fixes to display rates

woudini:
## Description

- Remove placeholder text for conversion title label.
- Fade in display rate for conversion title label.
- Speed up toggle to display rates.

## How to Test

Go to Exchange.
- Observe that the display rate is not populated with placeholder text
- Observe that the display rate fades in
- When tapping on the display rate, the animation for showing the rates happens at a reasonably fast speed.

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
+++ b/Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
@@ -65,6 +65,7 @@ class ExchangeCreateViewController: UIViewController {
     }
     
     enum ViewUpdate: Update {
+        case conversionTitleLabel(Visibility)
         case conversionView(Visibility)
         case exchangeButton(Visibility)
         case ratesChevron(Visibility)
@@ -275,6 +276,8 @@ extension ExchangeCreateViewController: ExchangeCreateInterface {
     
     func apply(update: ViewUpdate) {
         switch update {
+        case .conversionTitleLabel(let visibility):
+            conversionTitleLabel.alpha = visibility.defaultAlpha
         case .conversionView(let visibility):
             conversionView.alpha = visibility.defaultAlpha
         case .exchangeButton(let visibility):
--- a/Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
+++ b/Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
@@ -281,6 +281,12 @@ extension ExchangeCreatePresenter: ExchangeCreateOutput {
     
     func updatedRates(first: String, second: String, third: String) {
         interface?.apply(presentationUpdates: [.updateRateLabels(first: first, second: second, third: third)])
+        interface?.apply(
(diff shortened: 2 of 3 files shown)


Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
```diff
@@ -275,6 +276,8 @@ extension ExchangeCreateViewController: ExchangeCreateInterface {
     
     func apply(update: ViewUpdate) {
         switch update {
+        case .conversionTitleLabel(let visibility):
+            conversionTitleLabel.alpha = visibility.defaultAlpha
```
bchrisarriola: What about just setting the `conversionTitleLabel` to the empty string? I think that would be much simpler so that we don't have to worry about toggling its visibility.

thisisalexmcgregor: You can animate hiding it but you can only animate updating it's `.text` property with a `UIView.transition` animation block. If we want to do that, we can with a `TransitionUpdate`, which has a `.crossFade` transition type.

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
```diff
@@ -165,7 +165,7 @@ extension ExchangeCreatePresenter: ExchangeCreateDelegate {
             animatedUpdate: ExchangeCreateInterface.AnimatedUpdate(
                 animations: [.exchangeButton(.hidden),
                              .conversionView(.hidden)],
-                animation: .easeIn(duration: 2.0)
+                animation: .easeIn(duration: 0.2)
```
thisisalexmcgregor: heh, my bad.

Blockchain/Homebrew/Exchange/Controllers/ExchangeCreateViewController.swift
bchrisarriola: Yeah, toggling the alpha from 0 to 1 on each update would be pretty distracting.

So is the reason for this just to animate and ease in the rate once we receive the first rate from the websocket (sorry can't test right now because i'm getting WS auth issues)? I guess using the empty string and always having the label set to alpha = 1 won't animate.

woudini: >So is the reason for this just to animate and ease in the rate once we receive the first rate from the websocket

Yes. I do realize now though that's also inconsistent with how the `TradingPairView` updates.
>using the empty string and always having the label set to alpha = 1 won't animate.

It won't unless we use `.crossFade`.

thisisalexmcgregor: IMO probably shouldn't worry too much about this until we get some creative input on things like fonts, colors, UX etc. I'm guessing we'll be getting that soon.

PR #572, opened 2018-10-04
fix(HB): 1338 - Checking user account balance before creating a payment

thisisalexmcgregor:
## Objective

We now check the user's balance of the `from` account prior to creating a payment.

## Description

When creating an exchange, if the user enters in a fiat or crypto amount that is greater than what they have in their account, an error will be displayed. We check the `volume` of the `Conversion` and compare it to `account.balance`. 

## How to Test

1. Build and run
2. Go to the exchange 
3. Enter in an amount that is greater than what's in your account
4. You should see an error showing your account's balance in crypto. 

simulator screen shot - iphone 8 plus - 2018-10-04 at 14 47 18

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
+++ b/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
@@ -245,8 +245,7 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
         
         let min = minTradingLimit().asObservable()
         let max = maxTradingLimit().asObservable()
-        let usingBase = model.isUsingBase
-        let account = usingBase ? model.marketPair.fromAccount : model.marketPair.toAccount
+        let account = model.marketPair.fromAccount
         
         let disposable = Observable.zip(min, max) {
             return ($0, $1)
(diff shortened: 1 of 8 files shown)


thisisalexmcgregor:
@woudini yuck. Not sure what to do there though. We don't have font sizes for the primary or secondary label for each screen size. The comps show 72 points for the primary label, which is obviously *way* too big, particularly if you're entering in crypto.

woudini:
Well thinking further, the amount label font sizes will need to be increased for the larger screens anyway, so I don't think it's blocking for this (no point in fixing it now if it will change later either way)

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -245,19 +245,32 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
+        
         let disposable = Observable.zip(min, max) {
             return ($0, $1)
         }.subscribe(onNext: { [weak self] payload in
             guard let this = self else { return }
             let minValue = payload.0
             let maxValue = payload.1
-                
+            
+            guard let volume = Decimal(string: conversion.quote.volume) else { return }
```
woudini: When I type in $5, `volume` here is assigned to `5.0`, which seems to be interpreted here as the 5 BTC.

thisisalexmcgregor: 👀 whoops, looking.

thisisalexmcgregor: Fixed: 884e327

Blockchain/LocalizationConstants.swift
```diff
@@ -475,6 +475,8 @@ struct LocalizationConstants {
             "Your Max is",
             comment: "Error that displays what the maximum amount of fiat allowed for a trade"
         )
+        static let notEnough = NSLocalizedString("Not enough", comment: "Not enough")
+        static let yourBalance = NSLocalizedString("Your balance is", comment: "Your balance is")
```
woudini: Can you use the `comment:` parameter for these? Something like "Part of error message shown when the user doesn't have enough funds to make an exchange"

thisisalexmcgregor: Done 589c65a

PR #573, opened 2018-10-04
fix(HB): Fix issue where receiving zero trades from SS showed an error.

thisisalexmcgregor:
## Objective

Fixes a bug where `.error(nil)` was returned in a completionBlock when really it should be `.success([])` as the call to fetch partner trades completed successfully but there were no trades. 

## How to Test

N/A (Unless you've never done an SS trade, which in that case go to the exchange history screen and refresh your trades. You shouldn't see an alert.)

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Services/PartnerExchangeService.swift
+++ b/Blockchain/Homebrew/Exchange/Services/PartnerExchangeService.swift
@@ -9,7 +9,6 @@
 import Foundation
 
 protocol PartnerExchangeServiceDelegate {
-    func partnerExchangeDidFailToGetTrades(_ service: PartnerExchangeService, errorDescription: String)
     func partnerExchange(_ service: PartnerExchangeService, didGet exchangeRate: ExchangeRate)
     func partnerExchange(_ service: PartnerExchangeService, didGetBTC availableBalance: NSDictionary)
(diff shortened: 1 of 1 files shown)


Blockchain/Homebrew/Exchange/Services/PartnerExchangeService.swift
```diff
@@ -45,15 +44,14 @@ class PartnerExchangeService: PartnerExchangeAPI {
...
+        if let block = completionBlock {
+            block(.error(nil))
+        }
     }
     
     func didGetExchangeTrades(trades: NSArray) {
         if let block = completionBlock, trades.count == 0 {
-            block(.error(nil))
-            return
+            block(.success([]))
```
bchrisarriola: Early `return` here?

thisisalexmcgregor: Ah yeah. 

Would `guard let input = trades as? [ExchangeTrade] else { return }` fail for an empty array tho? Either way, I stuck in a `return`.

bchrisarriola: I think an empty array will pass the guard statement

PR #574, opened 2018-10-04
fix(Session Token): subtract 30 seconds from expiration time when trying to reuse cached token

bchrisarriola:
## Objective

Attempt to fix the issue wherein nabu returns the following error:

```
["description": Can not process auth request, token can not be found, "channel": auth, "type": error, "sequenceNumber": 0]
```

## Description

One of the scenarios wherein the server can return that error is if the client sends an expired token to the server. Since the server's clock and the phone's clock can be slightly off (i.e. the server might read time at "4:00.00 pm" but the client's clock might be at "3.59.30 pm"), the client might be sending expired tokens which result in failed WS auth calls. To prevent this, this PR subtracts 30 seconds off the expiration time comparison on the client so that the client will just request a new session token if it's within < 30 seconds of the expiration time in the phone's frame of reference.

## How to Test

This one is a bit tricky to test apart from just verifying that you can still get to the exchange create view and receive websocket calls (and not get the error message above).

## Merge Checklist

- [X] All unit tests pass.
- [X] You have added sufficient documentation (descriptive comments).

### DIFF
--- a/Blockchain/Homebrew/Authentication/NabuAuthenticationService.swift
+++ b/Blockchain/Homebrew/Authentication/NabuAuthenticationService.swift
@@ -14,7 +14,7 @@ final class NabuAuthenticationService {
 
     static let shared = NabuAuthenticationService()
 
-    private var cachedSessionToken = BehaviorRelay<NabuSessionTokenResponse?>(value: nil)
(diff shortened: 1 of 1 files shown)


thisisalexmcgregor:
:shipit:

PR #575, opened 2018-10-05
fix(WalletManager): pass partner exchange fetch error from JS to dele…

woudini:
…gate

## Objective

Ensure that homebrew trades are fetched even if fetching partner trades fails.

## Description

Passed error from JS to `partnerExchangeDelegate`.

## How to Test

- Log into a wallet that has both homebrew trades and partner trades in a region that is not supported by Shapeshift.
- Go to Exchange, see error in trade list, but still see homebrew trades being populated.

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Wallet/WalletManager.swift
+++ b/Blockchain/Wallet/WalletManager.swift
@@ -452,6 +452,12 @@ extension WalletManager: WalletDelegate {
         }
     }
 
+    func didFail(toGetExchangeTrades errorDescription: String!) {
+        DispatchQueue.main.async { [unowned self] in
+            self.partnerExchangeDelegate?.didFailToGetExchangeTrades(errorDescription: errorDescription)
+        }
+    }
+
     func didGet(_ rate: ExchangeRate!) {
         DispatchQueue.main.async { [unowned self] in
             self.partnerExchangeDelegate?.didGetExchangeRate(rate: rate)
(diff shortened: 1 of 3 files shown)


Blockchain/Wallet/WalletManager.swift
```diff
@@ -452,6 +452,12 @@ extension WalletManager: WalletDelegate {
         }
     }
 
+    func didFail(toGetExchangeTrades errorDescription: String!) {
+        DispatchQueue.main.async { [unowned self] in
```
thisisalexmcgregor: Not sure if this is intentional or not but `unowned` makes `self` non-optional. Granted I think `WalletManager` is a singleton so, I guess it's okay then? Anywho, just pointing that out. `[weak self] != [unowned self]`

woudini: I copied from other methods using `[unowned self]`. It could be because if the `WalletManager` isn't around something is terribly wrong anyway. cc @bchrisarriola

bchrisarriola: I definitely use the reasoning of using `[unowned self]` when `self` will absolutely not be nil, otherwise, I use `[weak self]`. In this scenario, I think this should be `[weak self]`

PR #577, opened 2018-10-05
fix(Homebrew-Exchange): show error messages immediately and disable e…

woudini:
…xchange button until input is valid

## Objective

Prevent user from continuing to the confirm screen until the input being displayed is valid.

## Description

Currently, user is allowed to continue to the confirm screen when typing in a valid input and then deleting that input. This is because conversion volume errors (in this case, the volume is too small) are not being taken into account. Since errors are not displayed immediately, these errors would have to be stored, which would complicate state management.

Web and Android are also displaying errors immediately.

## How to Test

Go to Exchange. By the time you are able to tap on "Exchange X for Y", the input should be valid

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
+++ b/Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
@@ -36,6 +36,8 @@ protocol ExchangeCreateInterface: class {
     func apply(presentationUpdates: [PresentationUpdate])
     func apply(animatedUpdate: AnimatedUpdate)
     func apply(transitionPresentation: AnimatedTransitionUpdate)
+
+    func exchangeEnabled(_ enabled: Bool)
 }
 
 // Conforms to NumberKeypadViewDelegate to avoid redundancy of keypad input methods
@@ -57,8 +59,9 @@ protocol ExchangeCreateOutput: class {
     func updateTradingPairValues(left: String, right: String)
     func updateTradingPair(pair: TradingPair, fix: Fix)
     func insufficientFunds(balance: String)
-    func entryBelowMinimumValue(minimum: String)
-    func entryAboveMaximumValue(maximum: String)
+    func entryBelowMinimumValue(minimum: String?)
+    func entryAboveMaximumValue(maximum: String?)
     func loadingVisibility(_ visibility: Visibility)
+    func exchangeEnabled(_ enabled: Bool)
     func showSummary(orderTransaction: OrderTransaction, conversion: Conversion)
 }
--- a/Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
+++ b/Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
@@ -59,6 +59,7 @@ protocol ExchangeCreateOutput: class {
     func updateTradingPairValues(left: String, right: String)
     func updateTradingPair(pair: TradingPair, fix: Fix)
     func insufficientFunds(balance: String)
+    func genericSocketError()
     func entryBelowMinimumValue(minimum: String?)
     func entryAboveMaximumValue(maximum: String?)
     func loadingVisibility(_ visibility: Visibility)
--- a/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
+++ b/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
@@ -370,7 +370,7 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
             // is than parsing the string
             if socketError.description.contains("small") {
                 this.output?.entryBelowMinimumValue(minimum: nil)
-            } else if socketError.description.contains("large") {
+            } else if socketError.description.contains("big") {
                 this.output?.entryAboveMaximumValue(maximum: nil)
             } else {
                 this.output?.genericSocketError()
--- a/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
+++ b/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
@@ -366,12 +366,14 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
 
         let errorDisposable = markets.errors.subscribe(onNext: { [weak self] socketError in
             guard let this = self else { return }
+            // There needs to be a better way to find out what the error
+            // is than parsing the string
             if socketError.description.contains("small") {
                 this.output?.entryBelowMinimumValue(minimum: nil)
             } else if socketError.description.contains("large") {
                 this.output?.entryAboveMaximumValue(maximum: nil)
             } else {
-                this.validateInput()
+                this.output?.genericSocketError()
             }
         })
 
--- a/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
+++ b/Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
@@ -368,9 +368,10 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
             guard let this = self else { return }
             // There needs to be a better way to find out what the error
             // is than parsing the string
-            if socketError.description.contains("small") {
+            if socketError.code == .tooSmallVolume ||
+                socketError.code == .resultCurrencyRatioTooSmall {
                 this.output?.entryBelowMinimumValue(minimum: nil)
-            } else if socketError.description.contains("big") {
+            } else if socketError.code == .tooBigVolume {
                 this.output?.entryAboveMaximumValue(maximum: nil)
             } else {
                 this.output?.genericSocketError()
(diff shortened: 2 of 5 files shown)


Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
```diff
@@ -57,8 +59,10 @@ protocol ExchangeCreateOutput: class {
     func updateTradingPairValues(left: String, right: String)
     func updateTradingPair(pair: TradingPair, fix: Fix)
     func insufficientFunds(balance: String)
-    func entryBelowMinimumValue(minimum: String)
-    func entryAboveMaximumValue(maximum: String)
+    func genericSocketError()
+    func entryBelowMinimumValue(minimum: String?)
+    func entryAboveMaximumValue(maximum: String?)
     func loadingVisibility(_ visibility: Visibility)
+    func exchangeEnabled(_ enabled: Bool)
```
thisisalexmcgregor: `exchangeEnabled` could mean that the entirety of that screen can be enabled or disabled. Really this is just the button. So let's rename this function to `exchangeButtonEnabled(_ enabled: Bool)`

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
```diff
@@ -93,19 +93,15 @@ class ExchangeCreatePresenter {
...
-            guard let this = self else { return }
-            this.setErrorDisappearanceTimer(duration: 2.5)
-        }
-        
         let group = ViewUpdateGroup(
             preparations: [.secondaryLabel(.hidden)],
             animations: [.errorLabel(.visible)],
             animation: .standard(duration: 0.2),
-            completionEvents: [.block(block)],
+            completionEvents: [],
```
thisisalexmcgregor: This is a really big departure from the feature that was described in the ticket. Now when I put in an amount that's too large, an error shows but, it never dismisses. Also, if we're not using the `completionEvents` then we shouldn't be using the `ViewUpdateGroup`.

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
```diff
@@ -249,21 +245,24 @@ extension ExchangeCreatePresenter: ExchangeCreateOutput {
     func insufficientFunds(balance: String) {
         interface?.apply(presentationUpdates: [.updateErrorLabel(balance)])
         displayError()
-        triggerErrorFeedback()
     }
     
-    func entryBelowMinimumValue(minimum: String) {
-        let display = LocalizationConstants.Exchange.yourMin + " " + minimum
+    func entryBelowMinimumValue(minimum: String?) {
```
thisisalexmcgregor: Why is this now optional? I don't think it should be.

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
```diff
@@ -249,21 +245,24 @@ extension ExchangeCreatePresenter: ExchangeCreateOutput {
...
+        let display = minimum == nil ? LocalizationConstants.Exchange.belowTradingLimit : LocalizationConstants.Exchange.yourMin + " " + minimum!
         interface?.apply(presentationUpdates: [.updateErrorLabel(display)])
         displayError()
-        triggerErrorFeedback()
     }
     
-    func entryAboveMaximumValue(maximum: String) {
-        let display = LocalizationConstants.Exchange.yourMax + " " + maximum
+    func entryAboveMaximumValue(maximum: String?) {
```
thisisalexmcgregor: Same thing here.

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
```diff
@@ -249,21 +245,24 @@ extension ExchangeCreatePresenter: ExchangeCreateOutput {
...
+        let display = maximum == nil ? LocalizationConstants.Exchange.aboveTradingLimit : LocalizationConstants.Exchange.yourMax + " " + maximum!
+        interface?.apply(presentationUpdates: [.updateErrorLabel(display)])
+        displayError()
+    }
+
+    func genericSocketError() {
```
thisisalexmcgregor: This error should take a string value and the string value should be passed in from the `Interactor`.

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
```diff
@@ -304,6 +303,11 @@ extension ExchangeCreatePresenter: ExchangeCreateOutput {
         interface?.apply(presentationUpdates: [.loadingIndicator(visibility)])
     }
 
+    func exchangeEnabled(_ enabled: Bool) {
+        errorDisappearanceTimerFired()
```
thisisalexmcgregor: This is confusing. A timer didn't fire here. The exchange was enabled.

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
woudini: >Now when I put in an amount that's too large, an error shows but, it never dismisses

Right, but allowing the dismissal would result in the label flashing between error and conversion output each time a too high/low conversion error comes in from the socket (which also happens irregularly).
Changed to apply an `ExchangeCreateInterface.AnimatedUpdate`.

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
woudini: We need a way to display volume errors returned from the socket, in case the user limits don't fully encapsulate the volume limits. These errors don't always return a particular minimum to use, and if they do, they need to be parsed from the description.

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -349,12 +358,23 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
+        }, onError: { error in
+            Logger.shared.error("Error subscribing to quote with trading pair")
         })
 
         let errorDisposable = markets.errors.subscribe(onNext: { [weak self] socketError in
-            // TODO: Implement error handling from Socket.
+            guard let this = self else { return }
+            if socketError.code == .tooSmallVolume ||
+                socketError.code == .resultCurrencyRatioTooSmall {
+                this.output?.entryBelowMinimumValue(minimum: nil)
```
thisisalexmcgregor: I still don't think `nil` should be used. Could you pass in `LocalizationConstants.Exchange.belowTradingLimit` here?

thisisalexmcgregor:
`confirmConversion()` doesn't look like it gets called anymore in `ExchangeCreatePresenter`. Also I guess `cancelErrorDisappearanceTimer`, `setErrorDisappearanceTimer`, and `triggerErrorFeedback` can be removed. RIP.

Blockchain/Homebrew/Exchange/Presenters/ExchangeCreatePresenter.swift
```diff
@@ -304,6 +294,11 @@ extension ExchangeCreatePresenter: ExchangeCreateOutput {
         interface?.apply(presentationUpdates: [.loadingIndicator(visibility)])
     }
 
+    func exchangeButtonEnabled(_ enabled: Bool) {
+        hideError()
```
thisisalexmcgregor: If I call `exchangeButtonEnabled` I wouldn't think this would also hide the error label. I know why it should be hidden but, it shouldn't be in this function is what I mean.

Blockchain/Homebrew/Exchange/API/ExchangeCreateContracts.swift
```diff
@@ -57,8 +61,12 @@ protocol ExchangeCreateOutput: class {
     func updateTradingPairValues(left: String, right: String)
     func updateTradingPair(pair: TradingPair, fix: Fix)
     func insufficientFunds(balance: String)
+    func genericError(message: String)
```
bchrisarriola: Rename to `showError` to match `hideError`

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -242,15 +269,38 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
+            with: conversion,
+            from: model.marketPair.fromAccount,
+            to: model.marketPair.toAccount,
+            success: { [weak self] orderTransaction, conversion in
+                guard let this = self else { return }
+                this.output?.loadingVisibility(.hidden)
+                this.output?.showSummary(orderTransaction: orderTransaction, conversion: conversion)
+            }, error: { [weak self] errorMessage in
+                guard let this = self else { return }
+                AlertViewPresenter.shared.standardError(message: errorMessage)
```
bchrisarriola: This should call `this.output?.genericError(message: errorMessage)` (better to rename to `showError` though)

PR #582, opened 2018-10-09
fix(Homebrew-KYC): treat userIds and tokens with length 0 as invalid

woudini:
## Objective

Fix issue where an empty `userId` is sent to the server. The error "userId format is incorrect" is returned.

## Description

The `userId` and `lifetimeToken` were both strings of length 0 when checking the QA wallet that was having this issue. With the changes made in 45970d4, tapping on the Exchange showed the "User already exists" error which displays as "This email address has already been used to verify an existing wallet." in an alert view.

Unsure of whether this issue is fully resolved at this point and looking for feedback here.

## How to Test

Tap on Exchange, should not see an error.

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Models/Wallet.m
+++ b/Blockchain/Models/Wallet.m
@@ -2619,7 +2619,8 @@
     if ([self isInitialized]) {
         JSValue *userId = [self.context evaluateScript:@"MyWalletPhone.KYC.userId()"];
         if ([userId isNull] || [userId isUndefined]) return nil;
-        return [userId toString];
+        NSString *userIdString = [userId toString];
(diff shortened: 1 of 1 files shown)


bchrisarriola:
What do you think caused kycUserId and lifetime token to be set as the empty string in the first place? That shouldn't happen at all.

woudini:
I think what might be happening is that wallets created on web are setting `userId` and `lifetimeToken` to empty strings. The issue appears to be inconsistent for some reason but this fix seems to at least allow it to display the "email already in use" error.

PR #584, opened 2018-10-09
fix(Homebrew-Exchange): add and use transparent visibility case for e…

woudini:
…xchange button

## Objective

Indicate to the user whether the Exchange button is disabled.

## Description

- Added new `Visibility` case `transparent` (alpha = 0.5).
- Added getter `isExchangeButtonEnabled` to maintain the correct `Visibility` when dismissing the `conversionRatesView`

I had considered and tried another approach to this which involved subclassing UIButton to add mutable closures like `onDidSetIsEnabled ((UIButton, isEnabled) -> ())` but it felt messy managing states in two different places (the closure and the interactor/presenter/interface). The trade-off here is repeating `disableExchangeButton()` in a few places to avoid side effects.

## How to Test

Go to Exchange. When errors appear, the "Exchange X for Y" button should be disabled. It should also be hidden appropriately when expanding the rates view and remain disabled after dismissing it if the input is still invalid (and enabled if the input is valid).

`.transparent` (alpha = 0.5)
screen shot 2018-10-09 at 7 21 19 pm

`.visible` (alpha = 1)
screen shot 2018-10-09 at 7 21 39 pm

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Models/ViewModels/Visibility.swift
+++ b/Blockchain/Models/ViewModels/Visibility.swift
@@ -10,13 +10,18 @@
 /// specific views. It's easier to read
 enum Visibility: Int {
     case hidden
+    case transparent
     case visible
 
     var defaultAlpha: CGFloat {
-        return self == .visible ? 1 : 0
+        switch self {
+        case .hidden: return 0
+        case .transparent: return 0.5
+        case .visible: return 1
+        }
     }
 
     var isHidden: Bool {
-        return self == .visible ? false: true
+        return self == .hidden ? true : false
     }
 }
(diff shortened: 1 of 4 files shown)


Blockchain/Models/ViewModels/Visibility.swift
```diff
@@ -10,13 +10,18 @@
 /// specific views. It's easier to read
 enum Visibility: Int {
     case hidden
+    case transparent
```
bchrisarriola: I think "translucent" is more accurate

PR #592, opened 2018-10-10
feat(Lockbox): Resolving IOS-1251 - Now showing lockbox side menu item.

thisisalexmcgregor:
## Objective

The `Lockbox` menu item is now visible.

## Description

I added a `.lockbox` menu item. Also, `SideMenuCell` now has a corresponding xib. If the menu item is not new, the `newContainer` will be hidden. 

## How to Test

1. Build and run
2. Go to the menu
3. See that `Lockbox` is shown along with the `new` label.
4. See that the icons have also been updated. 

simulator screen shot - iphone 8 plus - 2018-10-10 at 15 01 00

Only other thing I noticed is the peek amount seems to be a bit larger in the comps but, if we want to update that I can do that in a separate PR. 

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Coordinators/AppCoordinator.swift
```diff
@@ -222,6 +222,9 @@ extension AppCoordinator: SideMenuViewControllerDelegate {
             handleBuyBitcoin()
         case .exchange:
             handleExchange()
+        case .lockbox:
+            // TODO: Present `Lockbox` screen.
```
bchrisarriola: Nit: TICKET: IOS-1397 cc @achtenhagen

Blockchain/LocalizationConstants.swift
```diff
@@ -337,6 +337,14 @@ struct LocalizationConstants {
             "Support",
             comment: "Item displayed on the side menu of the app for when the user wants to contact support."
         )
+        static let new = NSLocalizedString(
+            "New",
+            comment: "New tag shown for menu items that are new."
+        )
+        static let lockbox = NSLocalizedString(
+            "Lockbox",
```
bchrisarriola: Not sure if this needs to be localized 🤔 Let's leave it here though

Blockchain/LocalizationConstants.swift
```diff
@@ -337,6 +337,14 @@ struct LocalizationConstants {
...
+            comment: "Lockbox menu itme title."
```
bchrisarriola: s/itme/item

Blockchain/Coordinators/AppCoordinator.swift
thisisalexmcgregor: done 6ea7259

PR #603, opened 2018-10-12
fix(ExchangeCreateInteractor): IOS-1435 record last ether tx and prevent sendi…

woudini:
…ng ether if waiting for confirmation

## Objective

Maintain same ETH send behavior in Exchange as in the regular send screen.

## Description

- Added recording of last ETH tx after sending ETH from exchange
- Show error message to inform user that they must wait for the last ETH tx to confirm before sending again

## How to Test

- Send ETH from regular Send tab or via Exchange
- See error message in Exchange when attempting to send from ETH 

simulator screen shot - iphone 6 - 2018-10-11 at 21 37 34

## Merge Checklist

- [x] All unit tests pass.


Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -524,6 +537,16 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
             return inputs.canAddAssetCharacter(value)
         }
     }
+
+    /// If the from asset type can be used, this method returns a nil string.
+    /// If it cannot be used, this method returns the localized error message explaining
+    /// why the asset type cannot be used.
+    private func canUseFrom(assetType: AssetType) -> String? {
```
bchrisarriola: This method name implies that it'll return a boolean. What about:

```
private func errorMessage(for assetType: AssetType) -> String?
```

...and this is invoked from the callsite as:

```
guard this.canTrade(assetType: assetType) else {
    if let errorMessage = this.errorMessage(for: assetType) {
        output.showError(message: errorMessage)
    }
}
```

Blockchain/js/wallet-ios.js
```diff
@@ -2464,20 +2463,25 @@ MyWalletPhone.sendEtherPaymentWithNote = function(note) {
 MyWalletPhone.sendEtherPayment = function(payment, success, error, dismiss) {
     var eth = MyWallet.wallet.eth;
 
+    let recordTx = function(tx) {
+        MyWalletPhone.recordLastTransaction(tx.txHash);
+        success()
```
bchrisarriola: Having a side-effect like this is not really good because reading the method "recordTx" implies that only recording the TX is done, however, it's also invoking the success function as well. Instead, I think it's better to chain a `.then` call.

Blockchain/js/wallet-ios.js
```diff
@@ -2464,20 +2463,25 @@ MyWalletPhone.sendEtherPaymentWithNote = function(note) {
...
+    }
+
     if (MyWallet.wallet.isDoubleEncrypted) {
         MyWalletPhone.getSecondPassword(function (pw) {
             var privateKey = eth.getPrivateKeyForAccount(eth.defaultAccount, pw);
             payment.sign(privateKey);
             payment
             .publish()
-            .then(success).catch(error);
+            .then(recordTx).catch(error);
```
bchrisarriola: From my comment above, that would turn this into:

```
.publish()
    .then(recordTx)
    .then(success)
    .catch(error);
```

Blockchain/js/wallet-ios.js
```diff
@@ -2464,20 +2463,25 @@ MyWalletPhone.sendEtherPaymentWithNote = function(note) {
...
         }, dismiss);
     } else {
         var privateKey = eth.getPrivateKeyForAccount(eth.defaultAccount);
         payment.sign(privateKey);
         payment
         .publish()
-        .then(success).catch(error);
+        .then(recordTx).catch(error);
```
bchrisarriola: Same here

Blockchain/Homebrew/Exchange/Interactors/ExchangeCreateInteractor.swift
```diff
@@ -287,13 +287,26 @@ extension ExchangeCreateInteractor: ExchangeCreateInput {
...
     func validateInput() {
         guard let model = model else { return }
+        guard let output = output else { return }
+
+        guard tradeExecution.canTradeAssetType(model.pair.from) else {
+            if let errorMessage = errorMessage(for: model.pair.from) {
+                output.showError(message: errorMessage)
+            } else {
+                // This shouldn't happen because the only case (eth) should have an error message,
+                // but just in case show an error here
```
bchrisarriola: 💯

PR #606, opened 2018-10-12
fix(Websocket Auth): fixing failed ws auth (IOS-1405, IOS-1434).

woudini:
## Objective

Fix issue where token expires after 5 minutes and isn't refreshed.

## Description

- `pendingSocketMessages` is an array used to store socket messages in case an attempt to send was made while the socket wasn't connected. This was not being cleared/reset properly, so multiple authentication messages were being sent, including ones that were using old session tokens. For now we clear the `pendingSocketMessages` the after `.forEach` has completed, but this should be improved on later.
- Disconnected from socket in `applicationDidEnterBackground`.
- Fixed logger typo.

## How to Test

- Open Exchange, leave on for > 5 minutes. You should still get conversion and rate updates.
- Open Exchange, leave on for > 5 minutes, tap sleep/wake button, go back to exchange, you should still get conversion and rate updates.

N/A

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Network/SocketManager.swift
+++ b/Blockchain/Network/SocketManager.swift
@@ -35,7 +35,9 @@ class SocketManager {
     // MARK: - Public methods
     func setupSocket(socketType: SocketType, url: URL) {
         switch socketType {
-        case .exchange: self.exchangeSocket = WebSocket(url: url); self.exchangeSocket?.advancedDelegate = self
+        case .exchange:
+            self.exchangeSocket = WebSocket(url: url)
+            self.exchangeSocket?.advancedDelegate = self
         default: Logger.shared.error(errorUnsupportedSocketType)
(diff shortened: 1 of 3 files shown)


Blockchain/Network/SocketManager.swift
```diff
@@ -35,7 +35,9 @@ class SocketManager {
     // MARK: - Public methods
     func setupSocket(socketType: SocketType, url: URL) {
         switch socketType {
-        case .exchange: self.exchangeSocket = WebSocket(url: url); self.exchangeSocket?.advancedDelegate = self
+        case .exchange:
+            self.exchangeSocket = WebSocket(url: url)
```
thisisalexmcgregor: `self.` isn't needed here.

PR #609, opened 2018-10-15
fix(Homebrew-Exchange): display deposit/withdrawal amounts as 0 if va…

woudini:
…lues are nil

## Objective

Fix issue where trades are not loading because of a null `deposit`.

## Description

- Made `deposit` optional.
- Displayed zero amount if the required `withdrawal` or `deposit` value was nil.

## How to Test

Go to Exchange - trades should load, and failed trade should show zero received amount.

simulator screen shot - iphone 6 - 2018-10-15 at 13 39 39
simulator screen shot - iphone 6 - 2018-10-15 at 13 39 37

## Merge Checklist

- [x] All unit tests pass.

### DIFF
--- a/Blockchain/Homebrew/Exchange/Models/CellModels/ExchangeTradeCellModel.swift
+++ b/Blockchain/Homebrew/Exchange/Models/CellModels/ExchangeTradeCellModel.swift
@@ -126,7 +126,9 @@ extension ExchangeTradeModel {
         case .partner(let model):
             return model.amountDepositedCryptoValue
         case .homebrew(let model):
-            return model.deposit.value + " " + model.deposit.symbol
+            return model.deposit != nil ?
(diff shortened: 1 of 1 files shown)


Blockchain/Homebrew/Exchange/Models/CellModels/ExchangeTradeCellModel.swift
```diff
@@ -126,7 +126,9 @@ extension ExchangeTradeModel {
         case .partner(let model):
             return model.amountDepositedCryptoValue
         case .homebrew(let model):
-            return model.deposit.value + " " + model.deposit.symbol
+            return model.deposit != nil ?
```
thisisalexmcgregor: Could we do a guard or an `if let` so we don't have to force unwrap? Whenever I see a `!` it feels like a code smell, even though in this case it wouldn't fail.

